PR #2, opened 2022-10-11
Skeleton of the orchestrator service, with plugins loader

jbonofre:


services/orchestrator/src/main/java/org/eclipse/osc/orchestrator/OrchestratorPlugin.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+public interface OrchestratorPlugin {
+
+    /**
+     * Register a managed service using the provided OCL descriptor.
+     *
+     * @param ocl the OCL model describing the managed service.
+     * @return the service ID on the CSP infrastructure.
+     */
+    String registerManagedService(Ocl ocl);
```
niuzhenguo, line 16: does this include preparing infrastructure and deploying the manged service? Seems we didn't talk about deploying before.

services/orchestrator/src/main/java/org/eclipse/osc/orchestrator/OrchestratorPlugin.java
```diff
@@ -0,0 +1,39 @@
...
+
+    /**
+     * Start (exposing the managed service to the users) the managed service.
+     *
+     * @param sid the service ID to start.
+     */
+    void startManagedService(String sid);
```
niuzhenguo, line 23: would it be possible to start/stop the managed service? seems there's some security issues to allow us to login to users guest os to do such operations.

services/orchestrator/src/main/java/org/eclipse/osc/orchestrator/OrchestratorPlugin.java
```diff
@@ -0,0 +1,39 @@
...
+     * @param sid the service ID to stop.
+     */
+    void stopManagedService(String sid);
+
+    /**
+     * Unregister and destroy/clean managed service resources.
+     *
+     * @param sid the service ID to unregister and destroy.
+     */
+    void unregisterManagedService(String sid);
```
niuzhenguo, line 37: besides creating/destroying infrastructure, do we need to consider updating resources like expanding storage size, changing flavor type of vm with high performance cores for the managed service?

services/orchestrator/src/main/java/org/eclipse/osc/orchestrator/OrchestratorPlugin.java
niuzhenguo: ignore this, if it means just exposing/hiding to users from osc layer.

README.md
```diff
@@ -28,6 +28,10 @@ OCL is a json descriptor of a managed service, describing the expected final sta
   "name": "my-service",
   "category": "compute",
   "namespace": "my-namespace",
+  "artifacts": [
+    "mvn:groupId/artifact/1.6",
```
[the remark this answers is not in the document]
jbonofre: +1, it's good idea. I will do the change. Thanks !

services/orchestrator/pom.xml
```diff
@@ -0,0 +1,52 @@
...
+
+    <dependencies>
+        <dependency>
+            <groupId>org.projectlombok</groupId>
+            <artifactId>lombok</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.karaf</groupId>
+            <artifactId>karaf-boot</artifactId>
```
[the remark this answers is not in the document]
jbonofre: Karaf K5 donation is currently in progress at Apache.
For now, I propose to deploy the K5 artifacts on my own Maven repo and add the repo in OSC.

services/orchestrator/src/main/java/org/eclipse/osc/orchestrator/OrchestratorPlugin.java
jbonofre: The idea behind this method is to be "global": this method is responsible of:
1. creating the infra as described in OCL
2. download the artifacts from the locations in OCL `artifacts`
3. bootstrapping the resources needed to actually start the service

NB: the Orchestrator (and generally OSC runtime) is supposed to run "in the CloudServiceProvider" infra. End user doesn't know OSC.

services/orchestrator/src/main/java/org/eclipse/osc/orchestrator/OrchestratorPlugin.java
jbonofre: Yes, correct. As said in previous comment, OSC is not visible for end users. The OSC runtime (and so orchestrator) is on CSP infra. It's actually the CSP "admin" service that interact with OSC orchestrator.

services/orchestrator/src/main/java/org/eclipse/osc/orchestrator/OrchestratorPlugin.java
jbonofre: Good question. Do you mean that we could have an update with a new OCL descriptor or generally speaking, starting from a OCL descriptor the CSP can change the config (or both) ?
I think it makes sense for a CSP to slightly update/change the initial OCL setup.

niuzhenguo: the idea is coming from Terraform, it's possible to update the TF script to make it change on the Cloud side.

jbonofre: It sounds good. I will update the API then. :+1:

services/orchestrator/src/main/java/org/eclipse/osc/orchestrator/OrchestratorPlugin.java
niuzhenguo: Maybe we can introduce Packer here to help build image on the CSP with the artifact and boot VM with that image. As far as I know, Spinnaker Rosco use this way.
https://www.packer.io/
https://github.com/spinnaker/rosco

jbonofre: Good idea. I think it could make sense in orchestrator plugin level.

services/orchestrator/src/main/java/org/eclipse/osc/orchestrator/OrchestratorPlugin.java
jbonofre: Agree, good point.

services/orchestrator/src/main/java/org/eclipse/osc/orchestrator/OrchestratorPlugin.java
```diff
@@ -0,0 +1,39 @@
+package org.eclipse.osc.orchestrator;
+
+import org.eclipse.osc.services.ocl.loader.Ocl;
+
+/**
+ * This interface describes orchestrator plugin in charge of interacting with backend fundamental APIs.
```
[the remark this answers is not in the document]
jbonofre: For the "existing" API (like compute/network/storage), I agree. However, I think it won't be enough: how can we do billing integration, admin console integration, monitoring/logging integration for instance ?
Does it mean that OCL would be a mix of IaC and additional metadata ? I see more IaC translation in the orchestrator plugin but both are possible.
Actually, I was thinking why not defining all with `properties` and then translate in the Orchestrator plugin.

Do you have kind of example snippet in mind ?

services/orchestrator/pom.xml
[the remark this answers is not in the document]
jbonofre: FYI https://lists.apache.org/thread/wl1l5ovgbdwblr30llppcoyhpdrso066
I'm renaming K5 to Karaf Minho, I will publish SNAPSHOTs on Apache maven repository and then I will update OSC PoC.

services/orchestrator/src/main/java/org/eclipse/osc/orchestrator/OrchestratorPlugin.java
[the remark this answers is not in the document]
niuzhenguo: actually, I still not quite sure about how we can do billing/console integration for now, so just raise the idea of IaC translation. it would be better to do some further discussion about this after the demo show.
If biling/console integration process are different with cloud service integration, Iac translation could make sense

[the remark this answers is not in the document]
jbonofre: Yes, it makes sense.And I agree: let's move forward with the "know how it works" things ;)

services/orchestrator/pom.xml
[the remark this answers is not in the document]
jbonofre: @niuzhenguo I propose to merge this PR. I created #11 to move to the "official" Apache Karaf Minho (I'm working on bootstrapping resources at Apache), I will do a PR dedicated to this one.

services/orchestrator/src/main/java/org/eclipse/osc/orchestrator/OrchestratorPlugin.java
jbonofre: I propose to create later issues and PRs about first `OrchestratorPlugins` we can create.

services/orchestrator/pom.xml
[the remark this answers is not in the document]
niuzhenguo: sure

PR #13, opened 2022-10-21
Update OCL for vm/image/artifacts.

### LINKED ISSUE #10: Refine OCL definition for compute vm

iskey:

- Update OCL with vm/image/artifacts.
- Throw out a disscution about the block attachment (Map or List).

Which issue this PR fixes:
Resolves #10


website/docs/ocl.md
```diff
@@ -120,25 +120,43 @@ This is the list of security (groups) defined in the service network. Each secur
...
-      "name": "my-artifact",
-      "type": "jar",
-      "url": "mvn:groupId/artifact/1.6",
-      "properties": {
-        "additional": "property",
-        "another": "one"
+  "image": {
+    "provisioners": [{
+      "name": "my-kafka-release",
+      "type": "shell",
```
jbonofre, line 126: Does it mean that we can have different kind of provisioning (like `ansible`, `shell`, ...) ? It looks good to me.

jbonofre: Just a note (not necessary in this PR): it would be great to document the expected behavior/lifecycle in the orchestrator.

iskey: > Does it mean that we can have different kind of provisioning (like `ansible`, `shell`, ...) ? It looks good to me.

That need a futher discussion. It is possible, but for now,  it would be better to compatible with `packer`,  which we will buiding on.

iskey: > Just a note (not necessary in this PR): it would be great to document the expected behavior/lifecycle in the orchestrator.

Good idea.

jbonofre: :+1 

OK to start with `packer`, but we should do that in kind of abstract way. For instance, you could also `tekton` in some use cases.
Remember the purpose is really to integrate with CSP fundamentals APIs, we have to "limit" the number of prerequisites services for the long term.

iskey: That's right, the `provisioners.type` here must be abstract. we just keep it similar to `packer`. The translation should be limited in the CSP's plugin.

jbonofre: :+1: 

Yes, fully agree ! The CSP plugin will deal with the provisioners ;)

website/docs/ocl.md
```diff
@@ -120,25 +120,43 @@ This is the list of security (groups) defined in the service network. Each secur
...
+      "environment_vars": [
+        "WORK_HOME=/usr1/KAFKA/"
+      ],
+      "inline": [
+        "cd ${WORK_HOME} && wget http://xxxx/kafka/release.jar"
+      ]
+    }
+    ],
+    "base": {
```
niuzhenguo, line 135: suggest to keep `base` and `artifacts` to List as others for now.

jbonofre: Agree: a service can use several image/artifacts (as a composition). That's why I used `List<Artifact>`.

website/docs/ocl.md
```diff
@@ -152,16 +170,23 @@ This is the list of security (groups) defined in the service network. Each secur
...
-      "subnet": "$.network.subnet[0]",
-      "security": ["$.network.security[0]"],
-      "storage": ["$.storage[0]}"],
-      "publicly": true
-    }]
+    "vm": [
+      {
+        "name": "my-vm",
+        "type": "t2.large",
+        "platform": "linux-x64",
```
niuzhenguo, line 177: seems platform is unneeded here.

website/docs/ocl.md
```diff
@@ -152,16 +170,23 @@ This is the list of security (groups) defined in the service network. Each secur
...
+        "image": "$.image.artifacts.kafka_image",
+        "vpc": "$.network.vpc[0]",
+        "subnet": "$.network.subnet[0]",
+        "security": [
+          "$.network.security[0]"
+        ],
+        "storage": [
+          "$.storage[0]}"
+        ],
+        "publicly": true
```
niuzhenguo, line 187: does this means public ip associated? if so how about replacing with a bandwidth parameter and when it's set greater than 0 means we need a public ip.

website/docs/ocl.md
iskey: > What this PR does / why we need it:
> 
> * Update OCL with vm/image/artifacts.
> * Throw out a disscution about the block attachment (Map or List).
> 
> Which issue this PR fixes: Resolves #10

Yes, As I mentioned in the PR header, we had a offline disscution yet. 
Now we decided to use List instead of Map.

website/docs/ocl.md
iskey: Agree, as the architecture information for `vm` will be carryied by `artifact_image`, `platform` is not needed any more.

website/docs/ocl.md
```diff
@@ -152,16 +170,23 @@ This is the list of security (groups) defined in the service network. Each secur
...
-      "storage": ["$.storage[0]}"],
-      "publicly": true
-    }]
+    "vm": [
+      {
+        "name": "my-vm",
+        "type": "t2.large",
+        "platform": "linux-x64",
+        "image": "$.image.artifacts.kafka_image",
+        "vpc": "$.network.vpc[0]",
```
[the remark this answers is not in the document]
iskey: Agree, I will delete it.

website/docs/ocl.md
niuzhenguo: ok, we can add public ip related things later.

README.md
```diff
@@ -64,22 +76,29 @@ OCL is a json descriptor of a managed service, describing the expected final sta
...
+        "type": "t2.large",
+        "image": "$.image.artifacts[0]",
+        "subnet": [
+          "$.network.subnet[0]"
+        ],
+        "security": [
+          "$.network.security[0]"
+        ],
+        "storage": [
+          "$.storage[0]}"
```
niuzhenguo, line 91: get rid of `}`

website/docs/ocl.md
```diff
@@ -24,6 +24,54 @@ An artifact is described with:
...
+* `environment_vars` - (Required) The variables for provisioner.
+* `inline` - (Required) The inline attributes for provisioner.
+
+```json
+{
+    "name": "my-kafka-release",
+    "type": "shell",
+    "environment_vars": [
+        "WORK_HOME=/usr1/KAFKA/"
+        ],
```
niuzhenguo, line 44: bad indentation

website/docs/ocl.md
```diff
@@ -24,6 +24,54 @@ An artifact is described with:
...
+    "inline": [
+    "cd ${WORK_HOME} && wget http://xxxx/kafka/release.jar"
```
niuzhenguo, line 46: ditto.

website/docs/ocl.md
```diff
@@ -120,25 +166,45 @@ This is the list of security (groups) defined in the service network. Each secur
...
-      "name": "another-artifact",
-      "type": "docker",
-      "url": "https://path/to/artifact",
-      "properties": {
-        "one": "property"
+    ],
+    "base": [
+      {
+        "name": "ubuntu-x64",
+        "ssh_user": "root",
```
niuzhenguo, line 185: ditto.

website/docs/ocl.md
```diff
@@ -120,25 +166,45 @@ This is the list of security (groups) defined in the service network. Each secur
...
+        "type": "t2.large",
+        "filters": {
+          "name": "ubuntu-for-osc-*"
+        }
+      },
+      {
+        "name": "centos-x64",
+        "ssh_user": "root",
```
niuzhenguo, line 193: ditto.

README.md
iskey: Agree.

Issue #5, opened 2022-10-17
Create orchestrator storage API with first local file impl


jbonofre:

jbonofre:
This one will started once #2 is merged.

PR #15, opened 2022-10-23
[#5] Create OrchestratorStorage with service loading from runtime

jbonofre:


jbonofre:
@niuzhenguo @iskey FYI, I updated the GitHub action workflow to build/check on each pull request (see the check tab).

services/orchestrator/src/main/java/org/eclipse/osc/orchestrator/MemoryOrchestratorStorageImpl.java
```diff
@@ -0,0 +1,32 @@
+package org.eclipse.osc.orchestrator;
+
+import org.apache.karaf.minho.boot.spi.Service;
+
+import java.util.Set;
+import java.util.concurrent.ConcurrentSkipListSet;
+
+public class MemoryOrchestratorStorageImpl implements OrchestratorStorage, Service {
```
niuzhenguo, line 8: Should we keep this memory storage way? As discussed last time, we are supposed to have a local file storage impl at least.

jbonofre: Yes. Good point. Let me update the PR with local file system storage and provide another distributed one (powered by Hazelcast for instance) as example.

jbonofre:
I'm resuming my work on this one, implementing a local filesystem storage by default (instead of memory).

jbonofre:
By the way, just a note: the storage implementation could be part of specific orchestrator plugin. For instance, within HuaweiCloud plugin, we can provide a orchestrator storage implementation based on etcd, etc.

niuzhenguo:
> By the way, just a note: the storage implementation could be part of specific orchestrator plugin. For instance, within HuaweiCloud plugin, we can provide a orchestrator storage implementation based on etcd, etc.

Yes, maybe an OBS storage implementation on Huawei Cloud and Swift on OpenStack.

jbonofre:
@niuzhenguo I updated this PR with:
1. Use properties with local file storage by default instead of in-memry
2. Add Swift based orchestrator storage skeleton in the openstack plugin

jbonofre:
@niuzhenguo @iskey I added openstack authentication in this PR by mistake. If it's OK for you, I will merge this PR and create a new one with moving forward on openstack plugin. Thoughts ?

services/openstack/src/main/java/org/eclipse/osc/orchestrator/plugin/openstack/OpenstackOrchestratorPlugin.java
```diff
@@ -1,22 +1,90 @@
...
+     * @return the token
+     * @throws Exception if the authentcation fails
+     */
+    private String authenticate() throws Exception {
+        log.info("Authenticate on " + authenticationUrl);
+        // TODO support proxy
+        HttpURLConnection connection = (HttpURLConnection) new URL(authenticationUrl).openConnection();
+        connection.setRequestMethod("POST");
+        connection.setRequestProperty("Content-Type", "application/json");
+        connection.setRequestProperty("Accept", "application/jsom");
```
niuzhenguo, line 70:
```suggestion
        connection.setRequestProperty("Accept", "application/json");
```

niuzhenguo (commented):
LGTM, suggest to use OpenStack Java sdk to do authentication in future, but it's OK for now as a POC implementation.

iskey:
LGTM, thanks!

services/openstack/src/main/java/org/eclipse/osc/orchestrator/plugin/openstack/OpenstackOrchestratorPlugin.java
jbonofre: Oh good catch, let me fix that.

jbonofre:
@niuzhenguo yes I plan to use openstack java sdk,  it's just skeleton for now.

iskey (approved):

PR #19, opened 2022-10-26
[#17] Update OCL base classes and test suit.

### LINKED ISSUE #17: Fix the test suits.

iskey:

- Fix the test suit.

Which issue this PR fixes:

Resolves #17

### DIFF
--- a/services/ocl-loader/src/main/java/org/eclipse/osc/services/ocl/loader/Artifact.java
+++ b/services/ocl-loader/src/main/java/org/eclipse/osc/services/ocl/loader/Artifact.java
@@ -4,13 +4,13 @@ import lombok.Data;
 
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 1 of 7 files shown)


niuzhenguo (approved):

jbonofre (approved):
Just a note: I think it would be great to avoid mentioning `huaweicloud` in PR title and/or commit message (like `huaweicloud#17`). It could be an issue when we will push to Eclipse.
It would be better to use ` osc#17` or just `#17`.
Thanks !

iskey:
> Just a note: I think it would be great to avoid mentioning `huaweicloud` in PR title and/or commit message (like `huaweicloud#17`). It could be an issue when we will push to Eclipse. It would be better to use ` osc#17` or just `#17`. Thanks !

Fully agree. I will amend it right away.

PR #28, opened 2022-10-29
Add skeleton of Huawei Cloud plugin

niuzhenguo:
Related #21

### DIFF
--- a/services/huaweicloud/pom.xml
+++ b/services/huaweicloud/pom.xml
@@ -1,0 +1,36 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.eclipse.osc</groupId>
+        <artifactId>services</artifactId>
+        <version>1.0.0-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+
+    <groupId>org.eclipse.osc.services</groupId>
+    <artifactId>huaweicloud</artifactId>
+    <name>OSC :: Services :: HuaweiCloud</name>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.projectlombok</groupId>
+            <artifactId>lombok</artifactId>
+            <scope>provided</scope>
+        </dependency>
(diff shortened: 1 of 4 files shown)


iskey (commented):
Just a note (maybe it is not a good way to move OrchestratorPlugin.java this to ../plugins).

jbonofre:
@iskey not sure I got your point.

IMHO, `OrchestratorPlugin` class should stay in `orchestrator` service.

For me, each plugin in a new service; so, I would expect to have (in terms of Maven modules):

* `services/huaweicloud`
* `services/openstack`
* ...

services/orchestrator/src/main/java/org/eclipse/osc/orchestrator/plugins/huaweicloud/HuaweiCloudPlugin.java
```diff
@@ -0,0 +1,48 @@
+package org.eclipse.osc.orchestrator;
+
+import lombok.extern.java.Log;
+import org.apache.karaf.minho.boot.spi.Service;
+import org.eclipse.osc.services.ocl.loader.Ocl;
+
+@Log
+public class HuaweiCloudPlugin implements OrchestratorPlugin, Service {
```
jbonofre, line 8: `HuaweiCloudPlugin` should be in its own module, not directly in the `Orchestrator`.

I think it's better to create `services/huaweicloud`. This module will have `orchestrator` as dependency and implements `HuaweiCloudPlugin`.

I'm doing the same with `services/openstack`.

Like this, we have a complete extensible mechanism where users can provide their own plugins.

It's in the runtime that we define the plugins we want (in the assembly).
We can imagine to have several runtime (like `osc-huaweicloud`, `osc-opentstack`, `osc-aws`, ...).

jbonofre (changes requested):
`OrchestratorPlugin` should stay in `Orchestrator` module and we should have a dedicated service/module for HuaweiCloud (e.g. `services/huaweicloud`).

jbonofre:
@niuzhenguo @iskey you can see the `openstack` plugin service skeleton on PR #29 

It's what I meant by having a `huaweicloud` plugin service.

iskey:
> @iskey not sure I got your point.
> 
> IMHO, `OrchestratorPlugin` class should stay in `orchestrator` service.
> 
> For me, each plugin in a new service; so, I would expect to have (in terms of Maven modules):
> 
> * `services/huaweicloud`
> * `services/openstack`
> * ...

Yes， I think maven modules are better.

niuzhenguo:
@jbonofre @iskey I will follow the openstack plugin way, we can do more discussion there :P

iskey:
lgtm.

iskey (approved):

jbonofre (approved):

PR #47, opened 2022-11-14
Add Ocl2Hcl.

### LINKED ISSUE #23: Translate OCL to Plugin description for Huawei Cloud
## Task List

* [ ] Translate image block to Packer description.
* [ ] Translate vm block to Terraform description.
* [ ] Translate network block to Terraform description.
* [ ] Translate storage block to Terraform description.

iskey:

Add OCL2HCL.

Which issue this PR fixes:
fix #23

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,4 +8,4 @@ atlassian-ide-plugin.xml
 bin
 dependency-reduced-pom.xml
 .apt_generated_tests
-
+*.iml
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ OCL is a json descriptor of a managed service, describing the expected final sta
       {
         "name": "my-kafka-release",
         "type": "shell",
-        "environment_vars": [
+        "environments": [
           "WORK_HOME=/usr1/KAFKA/"
         ],
         "inline": [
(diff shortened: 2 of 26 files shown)


services/huaweicloud/src/main/java/org/eclipse/osc/orchestrator/plugin/huaweicloud/builders/terraform/TFExecutor.java
```diff
@@ -0,0 +1,121 @@
...
+    public boolean createTFScript() {
+        Ocl2Hcl hcl = new Ocl2Hcl(ocl);
+        String hclStr = hcl.getHcl();
+
+        try {
+            try (FileWriter tfFile = new FileWriter(tfPath)) {
+                tfFile.write(hclStr);
+            }
+        } catch (IOException ex) {
+            log.warning(ex.getMessage() + Arrays.toString(ex.getStackTrace()));
```
Jiajia-Wen, line 63: I think we should unify our log framework, instead of ex.getStackTrace(),  Just using log.warn("some message", ex)

PR #48, opened 2022-11-14
adding SLF4j as the logging framework

Jiajia-Wen:
Adding a SLF4j as the logging framework and move the huawei cloud plugins to use SLF4j

### DIFF
--- a/services/huaweicloud/src/main/java/org/eclipse/osc/orchestrator/plugin/huaweicloud/HuaweiCloudOrchestratorPlugin.java
+++ b/services/huaweicloud/src/main/java/org/eclipse/osc/orchestrator/plugin/huaweicloud/HuaweiCloudOrchestratorPlugin.java
@@ -3,14 +3,14 @@ package org.eclipse.osc.orchestrator.plugin.huaweicloud;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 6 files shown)


niuzhenguo (approved):

iskey (approved):
Look good to me.

Jiajia-Wen:
I already checked the output ,since the underlying log framework is still java.util.log,  so the output format is the same as before.

services/huaweicloud/src/main/java/org/eclipse/osc/orchestrator/plugin/huaweicloud/HuaweiCloudOrchestratorPlugin.java
```diff
@@ -39,16 +40,15 @@ public void updateManagedService(String managedServiceName, Ocl ocl) {
     @Override
     public void startManagedService(String managedServiceName) {
         if (!managedOcl.containsKey(managedServiceName)) {
-            log.log(Level.WARNING, "Service: " + managedServiceName + " not registered.");
+            log.warn("Service: " + managedServiceName + " not registered.");
```
jbonofre, line 43: Using slf4j, we can also use interpolation like this:

```
log.warn("Service {} not registered.", managedServiceName);
```

Issue #45, opened 2022-11-14
Create OSC distribution per CSP


jbonofre:
I propose to create an OSC runtime distribution per CSP/plugin:
- osc-hc (for huawei cloud)
- osc-openstack
- osc-k8s

PR #52, opened 2022-11-15
[#45] Create one runtime distribution per CSP plugin (HuaweiCloud & Openstack for now)

jbonofre:


runtime/src/main/resources/osc-huaweicloud.json
```diff
@@ -1,14 +1,15 @@
 {
-  "name": "open-services-cloud",
+  "name": "osc-hc",
```
niuzhenguo, line 2: how about using `osc-huaweicloud` instead of `osc-hc`, to make it more recognizable.

jbonofre: Good idea. I wanted something "short", but I agree. Let me update to use full qualified name ;)

niuzhenguo (approved):
Thanks!

runtime/src/main/resources/osc-huaweicloud.json
niuzhenguo: not sure if `hc` can be a short name for huaweicloud :P

iskey (approved):
/lgtm

PR #53, opened 2022-11-16
add deep Copy method for Ocl Object 

Jiajia-Wen:
add deep Copy method for Ocl Object ,clean the log by using SLF4j and suppressed warning due to reasonable type conversion

### DIFF
--- a/.github/workflows/kanban.yml
+++ b/.github/workflows/kanban.yml
@@ -4,7 +4,7 @@ on:
   issues:
     types: [opened]
 env:
-  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  MY_GITHUB_TOKEN: ${{ secrets.MY_GITHUB_TOKEN }}
 
 jobs:
   assign_one_project:


iskey (approved):
It looks good to me.

services/ocl-loader/pom.xml
```diff
@@ -40,6 +40,17 @@
...
         </dependency>
 
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+            <version>2.0.3</version>
+        </dependency>
+
+        <dependency>
+            <groupId>org.slf4j</groupId>
```
jbonofre, line 50: Slf4j impl should be only in test scope. The runtime will décides of the impl/binding to use.

services/ocl-loader/src/main/java/org/eclipse/osc/services/ocl/loader/Ocl.java
```diff
@@ -1,21 +1,25 @@
 package org.eclipse.osc.services.ocl.loader;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
 import lombok.Data;
-import lombok.extern.java.Log;
+import lombok.extern.slf4j.Slf4j;
```
jbonofre, line 7: Use of slf4j is already done on main. Maybe you can rebase ?

services/ocl-loader/src/main/java/org/eclipse/osc/services/ocl/loader/Ocl.java
```diff
@@ -31,18 +35,41 @@ public class Ocl {
...
+
+    /**
+     * an OCL object might be passed to different plugins for processing,
+     * in case any plugin want to change the property of Ocl, we should not change
+     * the original
+     * Object, we should change a deep copy. The method is for deep copy
+     * 
+     * @return
+     */
+    public Ocl deepCopy() {
```
jbonofre, line 48: I see the point. The first intention was to have ocl kind of immutable object once loaded. The plug-ins could use their own ocl internal object. No problem for me though to use deep copy.

services/ocl-loader/src/main/java/org/eclipse/osc/services/ocl/loader/Ocl.java
```diff
@@ -53,21 +80,21 @@ public <T> Optional<T> referTo(String jsonPath, Class<T> valueType) {
...
-                } catch (NumberFormatException e){
+                } catch (NumberFormatException e) {
                     Method getter = object.getClass().getDeclaredMethod("get" +
-                            matchStr.substring(0,1).toUpperCase(Locale.ROOT) + matchStr.substring(1));
+                            matchStr.substring(0, 1).toUpperCase(Locale.ROOT) + matchStr.substring(1));
                     object = getter.invoke(object);
                 }
             } catch (Exception ex) {
-                log.log(Level.WARNING, ex.getMessage() + "\nStack Info:\n" + Arrays.toString(ex.getStackTrace()));
+                log.error(ex.getMessage() + "\nStack Info:\n", ex);
```
jbonofre, line 89: It should by ex directly as done on main.

jbonofre (changes requested):
Lgtm. Maybe just rebase for slf4j use.

services/ocl-loader/src/main/java/org/eclipse/osc/services/ocl/loader/Ocl.java
iskey: I think the plug-ins not only use the immutable ocl object but also have to clone a new one every time we call the `startManagedService()`. Am I right?

services/ocl-loader/pom.xml
jbonofre: Not necessary. It depends of the use they will do: some plugins might want, some others don't. Having a convenient method in the ocl loader service is ok.

iskey: Ok, it makes sense.

PR #66, opened 2022-11-21
Bugfix: fix the equals and hashCode method of data object when they extended a base data object

Jiajia-Wen:
The previous Data Object's equals and hashcode code generation have bugs, it doesn't call super() so two object will be equals even the property inherited from super is not equals. 

For example:
<code>
@Data
public class Storage extends RuntimeBase {
    private String name;
    private String type;
    private String size;
}
</code>

will have the wrong behavior:

        Storage storage = new Storage();
        Storage anotherStorage = new Storage();
        storage.setState("some state");
        anotherStorage.setState("another state");
        storage.setName("sameName");
        anotherStorage.setName("sameName");
        Assertions.Equals(storage, anotherStorage); //this line return true and the test can pass, which is wrong.


iskey (approved):
lgtm.

PR #69, opened 2022-11-25
Bugfix: the input and output stream are not closed.  It will cause me…

Jiajia-Wen:
There are several bugs in existing code :  #1.  Inputstream and Outputstream are not closed   #2.  We should not use properties.save().  According to JDK API doc,   the method is deprecated. 
  The method does not throw an IOException if an I/O error occurs while saving the property list. The preferred way to save a properties list is via the store(OutputStream out, String comments) method or the storeToXML(OutputStream os, String comment) method.


jbonofre (commented):
I changed the logic in this PR: https://github.com/huaweicloud/osc/pull/68/files#diff-e289ffc68a4ca41787c41b5775db6e038068e1a3d994d9a65fb20e9a87b55372

jbonofre:
I will merge your change proposal in the PR.

jbonofre:
@Jiajia-Wen thanks for the catch !

jbonofre (approved):
Thanks !

PR #73, opened 2022-11-25
Add terraform executor for huaweicloud plugin.

### LINKED ISSUE #72: Add terraform executor to huaweicloud plugin.

iskey:
fix #72

### DIFF
--- a/services/huaweicloud/src/main/java/org/eclipse/osc/orchestrator/plugin/huaweicloud/builders/terraform/TFExecutor.java
+++ b/services/huaweicloud/src/main/java/org/eclipse/osc/orchestrator/plugin/huaweicloud/builders/terraform/TFExecutor.java
@@ -33,7 +33,7 @@ public class TFExecutor {
                     + ws.getAbsolutePath());
         }
 
-        tfPath = ws.getAbsolutePath() + FileSystems.getDefault().getSeparator() + "main.tf";
+        tfPath = ws.getAbsolutePath() + FileSystems.getDefault().getSeparator() + "resources.tf";
         String verPath =
             ws.getAbsolutePath() + FileSystems.getDefault().getSeparator() + "version.tf";
         workPath = ws.getAbsolutePath();
--- a/services/huaweicloud/src/main/java/org/eclipse/osc/orchestrator/plugin/huaweicloud/builders/terraform/TFExecutor.java
+++ b/services/huaweicloud/src/main/java/org/eclipse/osc/orchestrator/plugin/huaweicloud/builders/terraform/TFExecutor.java
@@ -1,0 +1,139 @@
+package org.eclipse.osc.orchestrator.plugin.huaweicloud.builders.terraform;
+
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.nio.file.FileSystems;
+import java.util.Arrays;
+import java.util.Map;
+import lombok.extern.slf4j.Slf4j;
+import org.eclipse.osc.orchestrator.plugin.huaweicloud.exceptions.TFExecutorException;
+import org.eclipse.osc.services.ocl.loader.Ocl;
+
+@Slf4j
+public class TFExecutor {
+
+    private final Ocl ocl;
+    private final Map<String, String> env;
+    private String tfPath;
+    private String workPath;
+
+    public TFExecutor(final Ocl ocl, Map<String, String> env) {
+        this.ocl = ocl;
+        this.env = env;
+    }
+
+    public void createWorkspace() {
+        File ws =
+            new File("terraform_ws" + FileSystems.getDefault().getSeparator() + ocl.getName());
+
+        if (!ws.exists() && !ws.mkdirs()) {
+            throw new TFExecutorException(
+                "Create workspace for TFExecutor failed, File path not created: "
+                    + ws.getAbsolutePath());
+        }
+
+        tfPath = ws.getAbsolutePath() + FileSystems.getDefault().getSeparator() + "main.tf";
+        String verPath =
+            ws.getAbsolutePath() + FileSystems.getDefault().getSeparator() + "/version.tf";
+        workPath = ws.getAbsolutePath();
+        log.info("Terraform working directory is " + workPath);
+
+        try {
+            try (FileWriter versionFile = new FileWriter(verPath)) {
+                versionFile.write(""
+                    + "terraform {\n"
+                    + "  required_providers {\n"
+                    + "    huaweicloud = {\n"
+                    + "      source = \"huaweicloud/huaweicloud\"\n"
+                    + "      version = \">= 1.20.0\"\n"
+                    + "    }\n"
+                    + "  }\n"
+                    + "}");
+            }
+        } catch (IOException ex) {
+            log.error("Create terraform version file failed.", ex);
+            throw new TFExecutorException("Create terraform version file failed.", ex);
+        }
+    }
+
+    public void createTFScript() {
+        Ocl2Hcl hcl = new Ocl2Hcl(ocl);
+        String hclStr = hcl.getHcl();
+
+        try {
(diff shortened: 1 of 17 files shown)


services/huaweicloud/src/main/java/org/eclipse/osc/orchestrator/plugin/huaweicloud/builders/terraform/TFExecutor.java
```diff
@@ -0,0 +1,139 @@
...
+        File ws =
+            new File("terraform_ws" + FileSystems.getDefault().getSeparator() + ocl.getName());
+
+        if (!ws.exists() && !ws.mkdirs()) {
+            throw new TFExecutorException(
+                "Create workspace for TFExecutor failed, File path not created: "
+                    + ws.getAbsolutePath());
+        }
+
+        tfPath = ws.getAbsolutePath() + FileSystems.getDefault().getSeparator() + "main.tf";
```
niuzhenguo, line 38: seems this empty file is unneeded.

services/huaweicloud/src/main/java/org/eclipse/osc/orchestrator/plugin/huaweicloud/builders/terraform/TFExecutor.java
```diff
@@ -0,0 +1,139 @@
...
+    private String workPath;
+
+    public TFExecutor(final Ocl ocl, Map<String, String> env) {
+        this.ocl = ocl;
+        this.env = env;
+    }
+
+    public void createWorkspace() {
+        File ws =
+            new File("terraform_ws" + FileSystems.getDefault().getSeparator() + ocl.getName());
```
niuzhenguo, line 30: when to delete this dir?

services/huaweicloud/src/main/java/org/eclipse/osc/orchestrator/plugin/huaweicloud/builders/terraform/TFExecutor.java
```diff
@@ -0,0 +1,139 @@
...
+        if (!ws.exists() && !ws.mkdirs()) {
+            throw new TFExecutorException(
+                "Create workspace for TFExecutor failed, File path not created: "
+                    + ws.getAbsolutePath());
+        }
+
+        tfPath = ws.getAbsolutePath() + FileSystems.getDefault().getSeparator() + "main.tf";
+        String verPath =
+            ws.getAbsolutePath() + FileSystems.getDefault().getSeparator() + "/version.tf";
+        workPath = ws.getAbsolutePath();
```
niuzhenguo, line 41: It's ok for now, but suggest to use obs as backend storage to store tf state file in future.
https://registry.terraform.io/providers/huaweicloud/huaweicloud/latest/docs/guides/remote-state-backend

services/huaweicloud/src/main/java/org/eclipse/osc/orchestrator/plugin/huaweicloud/builders/terraform/TFExecutor.java
```diff
@@ -0,0 +1,139 @@
+package org.eclipse.osc.orchestrator.plugin.huaweicloud.builders.terraform;
```
niuzhenguo, line 1: how about wrapping an executor class, there will be more excutors like TFExecutor、RFExecutor、CloudFormationExecutor、PackerExecutor、APIExecutor...

iskey: IMHO, these executors can be implemented as different packages in the `builders` directory. Thus an executor maybe is so complicated to be carried by a class object.

services/huaweicloud/src/main/java/org/eclipse/osc/orchestrator/plugin/huaweicloud/builders/terraform/TFExecutor.java
iskey: Agree.
This is a temporary working directory, the valuable information needs to be analyzed and pushed to persistent storage. 
We can pull them out to a new working directory again for resource updating/destruction.

services/huaweicloud/src/main/java/org/eclipse/osc/orchestrator/plugin/huaweicloud/builders/terraform/TFExecutor.java
iskey: This will be a temporary directory, valuable information will be extracted to persistent storage, then this dir can be cleared safelly.

services/huaweicloud/src/main/java/org/eclipse/osc/orchestrator/plugin/huaweicloud/builders/terraform/TFExecutor.java
iskey: This path will be needed by the OCL2HCL process.

niuzhenguo: main.tf is usually used to place the provider configuration, maybe better to name it resources.tf or something like this.

services/huaweicloud/src/main/java/org/eclipse/osc/orchestrator/plugin/huaweicloud/builders/terraform/TFExecutor.java
niuzhenguo: ok, agree to extract some common methods for these executors and implement like this in different pkg. just some suggestion.

services/huaweicloud/src/main/java/org/eclipse/osc/orchestrator/plugin/huaweicloud/builders/terraform/TFExecutor.java
niuzhenguo: This should be with the service lifecycle unless the managed serveice destroyed, maybe we can remove this dir with that operation after `terraform destroy`

services/huaweicloud/src/main/java/org/eclipse/osc/orchestrator/plugin/huaweicloud/builders/terraform/TFExecutor.java
iskey: It makes sense.

services/huaweicloud/src/main/java/org/eclipse/osc/orchestrator/plugin/huaweicloud/builders/terraform/TFExecutor.java
iskey: Compared to the long lifecycle of the managed service,
I think the directories are too weak to hold such important things.
Anyway, let us have a further discussion about this.

niuzhenguo: ok, agree. but it's ok for now as a PoC.

services/huaweicloud/src/main/java/org/eclipse/osc/orchestrator/plugin/huaweicloud/builders/terraform/TFExecutor.java
niuzhenguo: ok, we can do more further discussion about this later.

niuzhenguo (approved):

services/huaweicloud/src/main/java/org/eclipse/osc/orchestrator/plugin/huaweicloud/builders/terraform/TFExecutor.java
niuzhenguo: do a change here?

niuzhenguo (approved):
/lgtm

Issue #20, opened 2022-10-26
Create Kafka OCL descriptor


jbonofre:
The purpose is really to describe Kafka managed service using OCL and use it as development base for Orchestrator plugins.

PR #75, opened 2022-11-27
[#20] Add simple kafka OCL descriptor (to be used for demos)

jbonofre:


samples/kafka.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "apache-kafka",
+  "category": "streaming",
+  "namespace": "default",
+  "image": {
+    "provisioners": [
+      {
+        "name": "apache-kafka-image",
+        "type": "docker",
```
niuzhenguo, line 9: seems this should be `shell` type as what the inline commands not that specific for docker, and we don't support docker type in the backends for now :(

samples/kafka.json
```diff
@@ -0,0 +1,19 @@
...
+        "environments": [],
+        "inline": [
+          "docker network create app-tier --driver bridge",
+          "docker run -d --name zookeeper-server --network app-tier -e ALLOW_ANONYMOUS_LOGIN=yes binami/zookeeper:latest",
+          "docker run -d --name kafka-server --network app-tier -e ALLOW_PLAINTEXT_LISTENER=yes -e KAFKA_CFG_ZOOKEEPER_CONNECT=zookeeper-server:2181 binami/kafka:latest"
+        ]
+      }
+    ]
+  }
```
niuzhenguo, line 18: suggest to add a basic image like centos or ubuntu for this and add corresponding cmd to install docker in above inline block.

jbonofre: +1, I  will update accordingly. Thanks!

jbonofre: @niuzhenguo what do you mean exactly ? I'm using the binami docker images.
Do you mean:
1. install docker daemon via OSC ?
2. or construct the zookeeper/kafka docker image by OSC ?

niuzhenguo: Hi JB, I mean, we need to deploy this on a virtual machine, so we need to build a packer image then run VM with that, and then within the VM, we run docker related cmds and using binami kafka docker image.

samples/kafka.json
```diff
@@ -0,0 +1,19 @@
...
+  "namespace": "default",
+  "image": {
+    "provisioners": [
+      {
+        "name": "apache-kafka-image",
+        "type": "shell",
+        "environments": [],
+        "inline": [
+          "docker network create app-tier --driver bridge",
+          "docker run -d --name zookeeper-server --network app-tier -e ALLOW_ANONYMOUS_LOGIN=yes binami/zookeeper:latest",
```
iskey, line 13:
```suggestion
          "docker run -d --name zookeeper-server --network app-tier -e ALLOW_ANONYMOUS_LOGIN=yes bitnami/zookeeper:latest",
```

samples/kafka.json
```diff
@@ -0,0 +1,19 @@
...
+          "docker run -d --name kafka-server --network app-tier -e ALLOW_PLAINTEXT_LISTENER=yes -e KAFKA_CFG_ZOOKEEPER_CONNECT=zookeeper-server:2181 binami/kafka:latest"
```
iskey, line 14:
```suggestion
          "docker run -d --name kafka-server --network app-tier -e ALLOW_PLAINTEXT_LISTENER=yes -e KAFKA_CFG_ZOOKEEPER_CONNECT=zookeeper-server:2181 bitnami/kafka:latest"
```

iskey (commented):
IMHO,  the command `docker run ...` should be moved to a startup script.
or just add some lines like:
```
"echo \"docker restart zookeeper-server\" >> /etc/profile",
"echo \"docker restart kafka-server\" >> /etc/profile"
```

samples/kafka.json
niuzhenguo: Regarding the VM image, it's in another block within the `image`, we can define the artifact thing for demo then, this for provisioners seems OK for me.

jbonofre:
@iskey @niuzhenguo as you guys have a clear view about OCL content (according to Huawei Cloud plugin), do you mind to push a Kafka OCL matching your needs ? I will adapt k8s and openstack plugins.

niuzhenguo:
> @iskey @niuzhenguo as you guys have a clear view about OCL content (according to Huawei Cloud plugin), do you mind to push a Kafka OCL matching your needs ? I will adapt k8s and openstack plugins.

I think it's ok to just merge this one with the provisioners defination, and we can add additional content for image artifact part.

iskey:
ok, thanks, let me explore the Ocl demo with Kafka in the HuaweiCloud plugin.

jbonofre:
@iskey @niuzhenguo I updated the PR with fix on docker image name. I suggest to merge this one and you guys can push your change. I will update in K8S and openstack plugins accordingly. Thanks !

niuzhenguo (approved):

PR #83, opened 2022-12-20
fix some bugs.  1. Java strings should use equlas but not = or !=.  2…

Jiajia-Wen:
some bugfix 

#1. Java strings should use equals but not == or !=
https://stackoverflow.com/questions/767372/string-equals-versus

#2. We should use %n instead of \n to make string.format portable across platform. 
https://stackoverflow.com/questions/50366242/using-n-or-n-in-java-to-format-string


iskey (approved):
LGTM.
By the way, packer/terraform scripts seem compatible with Unix-like line breaks and are platform-independent.

niuzhenguo (approved):

PR #87, opened 2022-12-30
update oscserverstarter

kylin321370:


jbonofre:
This PR doesn't make sense as we already have a starter (in osc-runtime). Let me update the documentation (I'm on it) and you will see.

jbonofre:
By the way, see #65 for tracking.

jbonofre:
I updated `README.md` to show how to launch the runtime (using minho.json as runtime conf).

PR #97, opened 2023-01-06
Add logging configuration module, set maven plugin versions, add new OscMain class

swaroopar:
1. Added new logging module
2. Added versions to maven plugins
3. added new Main class


services/api/src/main/java/org/eclipse/osc/services/api/OrchestratorApi.java
```diff
@@ -33,21 +36,16 @@ public Response fetch(@HeaderParam("ocl") String oclLocation) throws Exception {
     @Path("/health")
     @GET
     @Produces(MediaType.TEXT_PLAIN)
-    public String health() throws Exception {
-        if (getOrchestrator() == null) {
-            throw new IllegalStateException("Orchestrator service is not ready");
-        }
+    public String health() {
```
jbonofre, line 39: Why removing the check on the orchestrator service presence here ?

runtime/pom.xml
```diff
@@ -123,7 +130,7 @@
                             </tags>
                         </to>
                         <container>
-                            <mainClass>org.apache.karaf.minho.boot.Main</mainClass>
+                            <mainClass>org.eclipse.osc.runtime.OscMain</mainClass>
```
jbonofre, line 133: It means the OscMain will be used only in Docker container. Maybe worth to add in the runtime MANIFEST (Main-Class) in maven-jar-plugin to be able to do `java -jar runtime-1.0-SNAPSHOT.jar`.

jbonofre (changes requested):
Generally speaking, I think it's better:
1. To have dedicated PR per change (avoid to mix multiple changed in one PR)
2. Can you please squash your commits in one (easier to merge) ?

Thanks !

services/api/src/main/java/org/eclipse/osc/services/api/OrchestratorApi.java
swaroopar: This check is already there in the getOrchestrator() method. That's why removed it.

jbonofre: Ah yes, good point. I agree then.

swaroopar:
Will create new separate MRs.

jbonofre:
It makes sense. Thanks !

PR #100, opened 2023-01-08
Feature/logging

swaroopar:


modules/api/src/main/java/org/eclipse/osc/modules/api/OrchestratorApi.java
```diff
@@ -31,21 +36,16 @@ public Response fetch(@HeaderParam("ocl") String oclLocation) throws Exception {
     @Path("/health")
     @GET
     @Produces(MediaType.TEXT_PLAIN)
-    public String health() throws Exception {
-        if (getOrchestrator() == null) {
-            throw new IllegalStateException("Orchestrator service is not ready");
-        }
+    public String health() {
```
jbonofre, line 39: These changes are not related to logging right ?

plugins/huaweicloud/src/main/java/org/eclipse/osc/orchestrator/plugin/huaweicloud/HuaweiCloudOrchestratorPlugin.java
```diff
@@ -61,7 +61,7 @@ public void startManagedService(String managedServiceName) {
             throw new IllegalArgumentException("Service:" + managedServiceName + "not registered.");
         }
 
-        Optional<AtomBuilder> optionalAtomBuilder = createBuiler(managedServiceName);
+        Optional<AtomBuilder> optionalAtomBuilder = createBuilder(managedServiceName);
```
jbonofre, line 64: This typo fix is not related to logging. I have no problem to include in this PR, but the PR name could be "confusing".

pom.xml
```diff
@@ -145,6 +145,7 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
+                <version>3.10.1</version>
```
jbonofre, line 148: Same here: settings the plugin versions is not related to logging.

jbonofre (changes requested):

jbonofre:
I will create separated PRs for typo fix and maven plugin versions. I would focus on logging module in this PR (in a single commit), just to keep it simple.

modules/api/src/main/java/org/eclipse/osc/modules/api/OrchestratorApi.java
swaroopar: removed

swaroopar:
Hi @jbonofre Thanks for your review. i have removed the commits not related to logging. But i'm not able to squash all into one commit since the parents have diverged for one commit. If allowed, we can try squashing them while merging to main. I will take care of this next time.

jbonofre (changes requested):
Can you please squash all commits in one ?
To squash, you have to do:
```
git rebase -i origin/main
```
you keep the first commit as `pick` and you change all other commits as `squash` (instead of `pick`). Then, you will edit the commit message to something like `Adding OSC logging module`.
As you change the commit history, you have to force push on your branch:
```
git push --force
```
Then, you will have one commit instead of x ;)

swaroopar:
@jbonofre Thanks for the hint. Commits are squashed now. Please review.

jbonofre (approved):

PR #103, opened 2023-01-09
add main class and build fat jar

swaroopar:


swaroopar:
PR to be discussed with JB today.

runtime/pom.xml
```diff
@@ -51,10 +51,20 @@
             <artifactId>api</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.projectlombok</groupId>
```
jbonofre, line 55: Runtime should not have lombok dep.

runtime/pom.xml
```diff
@@ -51,10 +51,20 @@
...
+            <artifactId>lombok</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
```
jbonofre, line 60: Runtime already includes slf4j impl, so the api is implicite there.

runtime/pom.xml
```diff
@@ -140,10 +154,34 @@
                         </execution>
                     </executions>
                 </plugin>
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-assembly-plugin</artifactId>
```
jbonofre, line 159: assembly plugin is an option, we can also be simpler by using shade plugin.

jbonofre (changes requested):
The PR should be rebase to fix a conflict.
Generally speaking, if possible, I would keep it simpler (probably just shading instead of using an assembly).
By the way, minho-maven-plugin exists too (just FYI).
I propose to contribute directly on this PR to show another possible approach.

runtime/pom.xml
swaroopar: I added it since there is a log line in the OscMain class. Should we remove this log line?

runtime/pom.xml
swaroopar: removed.

swaroopar:
Hi JB, regarding shading - i gave it a try before assembly plugin. It is unpacking all the dependencies in the uber jar. It is somehow after that not clear which version of jars are actually pulled. 

So the jar looks like this now. with the  lib folder containing all jars. Please let me know if it is not really a best practice. Yes, please feel free to update the same PR with other approaches. 

image

runtime/pom.xml
jbonofre: I think so, I would avoid to added extra deps for just this.

runtime/pom.xml
```diff
@@ -86,7 +92,7 @@
                             <goal>copy-resources</goal>
                         </goals>
                         <configuration>
-                            <outputDirectory>${project.build.directory}/runtime</outputDirectory>
+                            <outputDirectory>${project.build.directory}</outputDirectory>
```
jbonofre, line 95: By the way, I think it's better to keep the resources in `runtime` folder, else we have all jar files directly in `target` folder.

runtime/pom.xml
swaroopar: switched to shade plugin.

runtime/pom.xml
swaroopar: reverted this change.

swaroopar:
@jbonofre Thanks again for the detailed review. I have now added jar-plugin for the osc-runtime jar itself. and shade-plugin for the uber jar.

PR #104, opened 2023-01-09
[22] Openstack plugin

jbonofre:


jbonofre:
That's rebase and use of openstack. We still need work on this one:
* check OCL resources binding
* check provisioning
* check storage

modules/ocl-loader/src/main/java/org/eclipse/osc/modules/ocl/loader/Artifact.java
```diff
@@ -11,6 +11,7 @@ public class Artifact extends RuntimeBase {
 
     private String name;
     private String base;
+    private String type;
```
swaroopar, line 14: If we know the list of Artifact types, we can make it an enum. Need not be a complete list. We can start with whatever values we know also.

plugins/openstack/src/main/java/org/eclipse/osc/orchestrator/plugin/openstack/OpenstackOrchestratorPlugin.java
```diff
@@ -6,90 +6,145 @@
...
+                ContainerBuilder builder = new MagnumContainer.ContainerConcreteBuilder();
+                builder.image(artifact.getBase());
+                builder.name(artifact.getName());
+                Container container = osClient.magnum().createContainer(builder.build());
+                osClient.magnum().startContainer(artifact.getName());
+                log.info("Docker container " + container.getStatus());
+            }
+            if (artifact.getType().equalsIgnoreCase("image")) {
+                log.info("Starting bare image via Glance ...");
+                Image image = osClient.imagesV2().create(Builders.imageV2()
```
swaroopar, line 131: From OSC, should we consider even uploading image or the assumption would be that the images already exist in openstack and we create only a VM out of it?

swaroopar (changes requested):

modules/ocl-loader/src/main/java/org/eclipse/osc/modules/ocl/loader/Artifact.java
jbonofre: The idea here is to have "open" type. So it's not necessary "closed" values, it could be specific to some plugins.

swaroopar: agreed.. IMHO, if we have an enum, then the user knows what values can be provided. The  enum can always be extended.

plugins/openstack/src/main/java/org/eclipse/osc/orchestrator/plugin/openstack/OpenstackOrchestratorPlugin.java
iskey: Agree to refactor, actually we have three types of images for now, `image`, `container`, and `artifact` (image+provisioners). 
For the `artifact`, how about we embed the base image into the `artifact` block directly?

iskey (approved):

plugins/openstack/src/main/java/org/eclipse/osc/orchestrator/plugin/openstack/OpenstackOrchestratorPlugin.java
swaroopar: @iskey Thanks. For sure we need some rework on this area of OCL data model. Let's discuss this as part of next sprint tasks.

swaroopar (approved):

iskey:
LGTM.

PR #112, opened 2023-01-17
Turn to the DefaultOrchestratorStorageBean in huawei plugin.

### LINKED ISSUE #111: Add a generic `file-storage` to modules.
As #89 mentioned, We should add a generic `file-storage` service.

iskey:
fix #111


iskey:
As we turned to spring boot, the default `file-storage` works fine. This module is not needed.

swaroopar (approved):
/LGTM

PR #127, opened 2023-02-02
Rename maven modules and API

iskey:
1. Rename maven modules.
2. Rename the API URI prefix from osc to xpanse.


swaroopar (approved):
/LGTM

Issue #122, opened 2023-01-31
Remove images baking from the huawei plugin


iskey:
1. Sperate the image baking to a new repository.

Issue #116, opened 2023-01-30
Update API documentation


swaroopar:
1. Rework data model structure -> needs discussions between developers. 
	2. Add types
	3. Add full documentation that will be also exposed via OpenAPI specs. - bean validation

PR #131, opened 2023-02-06
Add UserData for VM.

iskey:
Add UserData to `data.models` for VM.
related to #122 , #116

### DIFF
--- a/modules/ocl-loader/src/main/java/org/eclipse/xpanse/modules/ocl/loader/data/models/enums/UserDataType.java
+++ b/modules/ocl-loader/src/main/java/org/eclipse/xpanse/modules/ocl/loader/data/models/enums/UserDataType.java
@@ -1,0 +1,45 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * SPDX-FileCopyrightText: Huawei Inc.
+ *
+ */
+
+package org.eclipse.xpanse.modules.ocl.loader.data.models.enums;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonValue;
+
+/**
+ * user data types for UserData.
+ */
+public enum UserDataType {
+    SHELL("shell"),
+    POWERSHELL("powershell");
+
+    private final String scriptType;
+
+    UserDataType(String scriptType) {
+        this.scriptType = scriptType;
+    }
+
+    /**
+     * For UserDataType serialize.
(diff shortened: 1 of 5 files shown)


modules/ocl-loader/src/main/java/org/eclipse/xpanse/modules/ocl/loader/data/models/enums/UserDataType.java
```diff
@@ -0,0 +1,52 @@
...
+ * user data types for UserData.
+ */
+public enum UserDataType {
+    SHELL,
+    POWERSHELL;
+
+    private static Map<String, UserDataType> enumsMap = new HashMap<String, UserDataType>(
+            2);
+
+    static {
```
swaroopar, line 26: Its better if we can avoid these maps and define enums as below.
SHELL("shell"),
POWERSHELL("powershell");

the full class will look something like this (have not tested it). This gives a better readability of enum values.
```
public enum UserDataType {

    SHELL("shell"),
    POWEpublic enum UserDataType {

    SHELL("shell"),
    POWERSHELL("powershell");

    private final String scriptType;
RSHELL("powershell");

    private final String scriptType;

    UserDataType(String scriptType) {
        this.scriptType = scriptType;
    }

    /**
     * For UserDataType serialize.
     */
    @JsonCreator
    public UserDataType getByValue(String scriptType) {
        for (UserDataType userDataType : values()) {
            if (userDataType.scriptType.equals(scriptType)) {
                return userDataType;
            }
        }
        return null;
    }

    /**
     * For UserDataType deserialize.
     */
    @JsonValue
    public String toValue() {
        return this.scriptType;
    }
}

```

swaroopar (commented):
Rest looks good. Only enum class needs to changed.

plugins/huaweicloud/src/test/resources/huawei_test.json
```diff
@@ -71,7 +71,20 @@
         "storage": [
           "$.storage[0]"
         ],
-        "publicly": true
+        "publicly": true,
```
niuzhenguo, line 74: If changing to this userdata way, public ip will be a must :(

iskey: If something must be installed from the internet, yes. If only configure something, no.

swaroopar (approved):

niuzhenguo (approved):
/lgtm

PR #132, opened 2023-02-07
Remove image-baking

### LINKED ISSUE #122: Remove images baking from the huawei plugin

iskey:
fix #122 

- [x] 1. Remove Image-baking codes.
- [x] 2. Revert some missing codes (caused by Spring-Boot refactoring) for the Huawei plugin.


plugins/huaweicloud/src/main/java/org/eclipse/xpanse/orchestrator/plugin/huaweicloud/builders/utils/IpTool.java
```diff
@@ -0,0 +1,59 @@
...
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.net.URL;
+import org.apache.commons.validator.routines.InetAddressValidator;
+
+/**
+ * A Tool to probe external IP.
+ */
+public class IpTool {
```
swaroopar, line 18: The purpose of this class is not very clear to me. 
if it is to get external IP of the VM, then we will need some other better way of handling it. (Eg., getting it from the EIP resource)
Also we can make such methods as static methods. .

swaroopar (changes requested):
Please check the IpTool implementation again. Rest looks good.

plugins/huaweicloud/src/main/java/org/eclipse/xpanse/orchestrator/plugin/huaweicloud/builders/utils/IpTool.java
iskey: `IpTool` is a tool class to probe the external IP of the host of the `xpanse` which will be used by the security-group-rules.
As cloud-init in the HuaweiCloud has some issues(still checking). I choose to use the `provisions` ability of terraform to implement the software installing.

swaroopar (approved):
/LGTM

PR #143, opened 2023-02-14
use database instead of file storage

### LINKED ISSUE #142: Update response model for "state" and "services" API methods
Currently the state and services API method just returns a string. 
We must instead return a well defined complex type. 
1. Name
2. State -> enum - registered, started, deploying, deployed. 
3. id -> will be a new field once we have asynchronous processing implemented.

service -> to return the list of the "state" of all deployed services.

swaroopar:
Fix #142
Switch to database instead of file storage. 
Add proper types to API.


modules/ocl-loader/src/main/java/org/eclipse/xpanse/modules/ocl/loader/data/models/enums/HealthStatus.java
```diff
@@ -0,0 +1,46 @@
...
+        this.state = state;
+    }
+
+    /**
+     * For RuntimeState serialize.
+     */
+    @JsonCreator
+    public HealthStatus getByValue(String state) {
+        for (HealthStatus healthStatus : values()) {
+            if (healthStatus.state.equals(StringUtils.lowerCase(state))) {
```
baixinsui, line 32: StringUtils.lowerCase() should be replaced by StringUtils.upperCase()

swaroopar: fixed.

iskey (approved):
looks good.

PR #146, opened 2023-02-17
Asynchronous start and stop service with @Async

baixinsui:


modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/OrchestratorService.java
```diff
@@ -33,15 +37,23 @@
 @Component
 public class OrchestratorService implements ApplicationListener<ApplicationEvent> {
 
+    private static final int STATUS_MSG_MAX_LENGTH = 255;
+
+    private static final int CPU_COUNT = Runtime.getRuntime().availableProcessors();
+
+    public static ThreadPoolExecutor executorService = new ThreadPoolExecutor(CPU_COUNT * 2,
```
swaroopar, line 44: Would it be possible to use the Spring Async framework instead of directly creating threads?

modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/OrchestratorService.java
```diff
@@ -147,23 +161,41 @@ public void updateManagedService(String managedServiceName, Ocl ocl) {
...
+                }
+            }
             try {
                 serviceStatusEntity.setServiceState(ServiceState.STARTING);
                 this.databaseOrchestratorStorage.store(serviceStatusEntity);
-                orchestratorPlugin.startManagedService(managedServiceName);
-                serviceStatusEntity.setServiceState(ServiceState.STARTED);
-                this.databaseOrchestratorStorage.store(serviceStatusEntity);
-
+                executorService.execute(
```
swaroopar, line 190: We need async to be on all services.. register, start, stop, unregister. 
With our current plugins, we have only "start" doing something. But in our end goal, all these 4 methods will have actions which will take longer.

swaroopar (changes requested):

modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/OrchestratorService.java
baixinsui: Ok. I add the API of unregister and async register, start, stop, unregister services

modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/OrchestratorService.java
baixinsui: Ok.

modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/OrchestratorService.java
```diff
@@ -78,19 +84,25 @@ public void registerManagedService(String oclLocation) throws Exception {
...
+            throw new RuntimeException(
+                "Managed service " + ocl.getName() + " already registered.");
+        }
         if (plugins.isEmpty()) {
             log.warn("No plugins available. Request ignored.");
-            return;
+            throw new RuntimeException("No plugins available.");
         }
-
+        Thread.currentThread().setName(Thread.currentThread().getName() +
```
swaroopar, line 97: Do we really need to add the thread names to be updated? 
I prefer that we dont touch the thread name and instead add a MDC  for plugin name in the log line.

modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/OrchestratorService.java
```diff
@@ -68,7 +72,9 @@ public void onApplicationEvent(ApplicationEvent event) {
      * @param oclLocation the location of the OCL descriptor.
      * @throws Exception if registration fails.
      */
+    @Async("taskExecutor")
```
swaroopar, line 75: Can you also please check if the API now returns HTTP 202 immediatly and swagger UI also shows response code 202?

swaroopar (changes requested):
@baixinsui Thank you for the changes. Just some more minor queries added.

modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/OrchestratorService.java
baixinsui: It returns HTTP 200 immediatly when call the API

modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/OrchestratorService.java
baixinsui: I add plugin name and service name into the thread name now.

modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/thread/ServiceThreadPoolTaskExecutor.java
```diff
@@ -0,0 +1,42 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * SPDX-FileCopyrightText: Huawei Inc.
+ *
+ */
+
+package org.eclipse.xpanse.orchestrator.thread;
```
swaroopar, line 7: please name the packge as 'async' instead of thread.

modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/thread/ThreadMdcUtil.java
```diff
@@ -0,0 +1,86 @@
...
+ */
+public final class ThreadMdcUtil {
+
+    private static final String TRACE_ID = "TRACE_ID";
+
+    /**
+     * @return uuid
+     */
+    public static String generateTraceId() {
+        return UUID.randomUUID().toString();
```
swaroopar, line 25: we are creating a UUID for each register request while creating the DB entry. We can reuse the same to be consistent here.

modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/thread/ServiceThreadPoolTaskExecutor.java
baixinsui: ok

modules/api/src/main/java/org/eclipse/xpanse/api/interceptor/LogInterceptor.java
```diff
@@ -0,0 +1,57 @@
...
+     *
+     * @param request  http request
+     * @param response http response
+     * @param handler  http handler
+     * @return true
+     */
+    @Override
+    public boolean preHandle(HttpServletRequest request, HttpServletResponse response,
+            Object handler) {
+        String tid = UUID.randomUUID().toString().replace("-", "");
```
swaroopar, line 34: we are still creating a new UUID here?

modules/ocl-loader/src/main/java/org/eclipse/xpanse/modules/ocl/loader/data/models/ServiceStatus.java
```diff
@@ -19,6 +19,11 @@
 @Data
 public class ServiceStatus {
 
+    @NotNull
+    @NotBlank
+    @NotEmpty
+    @Schema(description = "The uuid of the service")
+    String serviceId;
```
swaroopar, line 26: type can be UUID instead of String?

swaroopar (changes requested):
Please also confirm if the async methods now return HTTP 202 instead of 200.

modules/api/src/main/java/org/eclipse/xpanse/api/interceptor/LogInterceptor.java
baixinsui: This UUID can be used to link the http request，if not to output in log，I will remove related code

modules/ocl-loader/src/main/java/org/eclipse/xpanse/modules/ocl/loader/data/models/ServiceStatus.java
baixinsui: Yes. I will change this type

baixinsui:
start and stop use @Async and http status is 202

baixinsui:
> Please also confirm if the async methods now return HTTP 202 instead of 200.
Ok. I I modify the HTTP STATUS of  async methods to ACCEPTED now return HTTP 202.

modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/OrchestratorService.java
baixinsui: The start and stop API using @Async now return HTTP 202 immediatly

swaroopar (approved):
/LGTM

PR #135, opened 2023-02-09
Refactor APIs.

iskey:
related to #116 , #124 

- [x] 1. Refactor data models.
- [x] 2. Remove some unused data models.
- [x] 3. Make some objects to Enums.

PR #153, opened 2023-02-24
Add objects fo OCL2.0

### LINKED ISSUE #124: Refactor the OCL
- [x] 1. Refactor the image part.(fixed by #135)
- [x] 2. Use existing infra?
- [x] 3. Replace `json` with `yaml`? (need more discussion)
- [x] 4. Do we need the controller?
- [x] 5. Refactor the OCL with more service-level structures, like flavor, versions, lifecycle, and ...?
- [x] 6. Try to replace the infra deploy with `terraform` directly. (need more discussion)

As discussed by #144

iskey:
fix #124, as discussed by #144 

- [x] 1. Refactor the image part.(fixed by #135)
- [x] 2. Use existing infra?（we can use existing infra from deployment）
- [x] 3. Replace `json` with `yaml`? (make a yaml version)
- [x] 4. Do we need the controller? (seems we no longer need a controller for Oclv2,)
- [x] 5. Refactor the OCL with more service-level structures, like flavor, versions, lifecycle, and ...?(done)
- [x] 6. Try to replace the infra deploy with `terraform` directly. (we think terraform is better.)

Also, did some API refactors:

```java
APIs for ISV
---
POST: /register 
    ==>  POST:   /register  + "application/x-yaml"
POST: /register/fetch   				  
    ==>  POST:   /register + "application/x-www-form-urlencoded"
PUT:  /update/{managedServiceName}        
    ==>  PUT:    /register/{managedServiceName} + "application/x-yaml"
PUT:  /update/fetch/{managedServiceName}   
    ==>  PUT:    /register/{managedServiceName} + "application/x-www-form-urlencoded"
POST: /unregister/{managedServiceName}     
    ==>  DELETE: /register/{managedServiceName} + "application/json"

APIs for End User
---
GET:  /services/state/{managedServiceName} 
    ==>  GET:    /service/{managedServiceName}
POST: /start/{managedServiceName} 
    ==>  POST:   /service/{managedServiceName}  + "application/json"
POST: /stop/{managedServiceName} 
    ==>  DELETE: /service/{managedServiceName}  + "application/json"

TODO : (will be added by demand)
    ==>  GET:    /register
    ==>  GET:    /register/{managedServiceName} 

```

image


modules/api/src/main/java/org/eclipse/xpanse/api/OrchestratorApi.java
```diff
@@ -61,6 +63,24 @@ public Response register(@Valid @RequestBody Ocl ocl) {
         return Response.successResponse(successMsg);
     }
 
+    /**
+     * Register new managed service.
+     *
+     * @param ocl object of managed service.
+     * @return response
+     */
+    @PostMapping(value = "/v2/register",
```
swaroopar, line 72: @iskey As discussed, we can remove the V2 for register and just update the existing register method to use the OclV2

swaroopar (changes requested):
Rest looks good. Just the register method can be updated instead of adding new one.

modules/api/src/main/java/org/eclipse/xpanse/api/OrchestratorApi.java
iskey: > @iskey As discussed, we can remove the V2 for register and just update the existing register method to use the OclV2

Hi @swaroopar , I fix this with some API refactors. just like [this](https://github.com/huaweicloud/xpanse/pull/153#:~:text=POST%3A%20/register%20%0A%20%20%20%20%3D%3D%3E%20%20POST%3A%20%20%20/,GET%3A%20%20%20%20/register/%7BmanagedServiceName%7D)

iskey:
> Rest looks good. Just the register method can be updated instead of adding new one.

Yes, fixed with some API refactors. pls, take a look again at your convenience.

modules/ocl-loader/src/main/java/org/eclipse/xpanse/modules/ocl/loader/OclLoader.java
```diff
@@ -17,10 +19,15 @@
 @Component
 public class OclLoader {
 
-    private final ObjectMapper mapper = new ObjectMapper();
+    private final ObjectMapper mapper = new ObjectMapper(new YAMLFactory());
```
swaroopar, line 22: do we really need two ObjectMappers here?

modules/ocl-loader/src/main/java/org/eclipse/xpanse/modules/ocl/loader/data/models/enums/DeployerKind.java
```diff
@@ -0,0 +1,45 @@
...
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonValue;
+
+/**
+ * Deployer kinds.
+ */
+public enum DeployerKind {
+    TERRAFORM("terraform"),
+    PULUMI("pulumi");
```
swaroopar, line 17: i think we can remove this 2nd value for now as we do not support it for now. we examples in the field description if needed.

modules/ocl-loader/src/main/java/org/eclipse/xpanse/modules/ocl/loader/data/models/v2/Billing.java
```diff
@@ -0,0 +1,39 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * SPDX-FileCopyrightText: Huawei Inc.
+ *
+ */
+
+package org.eclipse.xpanse.modules.ocl.loader.data.models.v2;
```
swaroopar, line 7: do we really need a V2 package? I would avoid that if it is possible.

modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/OrchestratorService.java
```diff
@@ -84,17 +85,17 @@ public void onApplicationEvent(ApplicationEvent event) {
      * @param oclLocation the location of the OCL descriptor.
      * @throws Exception if registration fails.
      */
-    public void registerManagedService(String oclLocation) throws Exception {
+    public void registerManagedServiceDeprecated(String oclLocation) throws Exception {
```
swaroopar, line 88: instead of renaming the method.. can we simply delete it? since it is anyways not used by any real users.

modules/api/src/main/java/org/eclipse/xpanse/api/OrchestratorApi.java
```diff
@@ -122,7 +144,7 @@ public List<ServiceStatus> services() {
      * @param managedServiceName name of managed service
      * @return response
      */
-    @PostMapping("/start/{managedServiceName}")
+    @PostMapping("/service/{managedServiceName}")
```
swaroopar, line 147: now since this API has methods for ISV and end user both, we can consider add tags to the services

@Api(tags = "Service Provider or Service Vendor)
@Api(tags = "End User or Service Consumer")

With this the services should be ideally grouped in the open api json and also the swagger UI

swaroopar (changes requested):
Hi @iskey thanks for the changes. Just some minor comments written.

modules/ocl-loader/src/main/java/org/eclipse/xpanse/modules/ocl/loader/OclLoader.java
iskey: yes, This PR is only for the API refactor. I will get rid of the old codes soon.

swaroopar (approved):
/LGTM

PR #164, opened 2023-03-02
add register manage APIs

baixinsui:


modules/api/src/main/java/org/eclipse/xpanse/api/OrchestratorApi.java
```diff
@@ -44,78 +51,205 @@ public class OrchestratorApi {
...
+    }
+
+
+    /**
+     * Get registered service using id.
+     *
+     * @param id id of registered service.
+     * @return response
+     */
+    @Tag(name = "Service Provider or Service Vendor",
```
swaroopar, line 230: Tag name should be just "Service Vendor" for all vendor related services

modules/api/src/main/java/org/eclipse/xpanse/api/OrchestratorApi.java
```diff
@@ -44,78 +51,205 @@ public class OrchestratorApi {
...
+            description = "APIs to get detail the registered services.")
+    @Operation(description = "Get registered service using id.")
+    @GetMapping(value = "/register/service/{id}",
+            produces = MediaType.APPLICATION_JSON_VALUE)
+    @ResponseStatus(HttpStatus.OK)
+    public Response detail(@PathVariable("id") String id) {
```
swaroopar, line 236: we must return OCL object of the service.

modules/api/src/main/java/org/eclipse/xpanse/api/OrchestratorApi.java
```diff
@@ -44,78 +51,205 @@ public class OrchestratorApi {
...
-    @Tag(name = "Service Vendor",
-            description = "APIs to publish the managed services.")
+    @Tag(name = "Service Provider or Service Vendor",
+            description = "APIs to manage register services.")
+    @Operation(description = "List cloud service providers.")
+    @GetMapping(value = "/register/providers",
+            produces = MediaType.APPLICATION_JSON_VALUE)
+    @ResponseStatus(HttpStatus.OK)
+    @Transactional
+    public Response providers() {
```
swaroopar, line 75: we already have the providers as an enum and any client can/must simply use  that directly. I think we dont need this extra service.

runtime/src/main/java/org/eclipse/xpanse/runtime/XpanseApplication.java
```diff
@@ -21,6 +22,7 @@
 @EnableJpaRepositories(basePackages = "org.eclipse.xpanse")
 @EntityScan(basePackages = "org.eclipse.xpanse")
 @ComponentScan(basePackages = "org.eclipse.xpanse")
+@EnableJpaAuditing
```
swaroopar, line 25: do we need this?

modules/service/src/main/java/org/eclipse/xpanse/service/impl/DatabaseRegisterServiceStorage.java
```diff
@@ -0,0 +1,181 @@
...
+     * @param registerServiceEntity register service entity
+     */
+    @Override
+    public void remove(RegisterServiceEntity registerServiceEntity) {
+        registerServiceRepository.delete(registerServiceEntity);
+    }
+
+
+    private String getRegisterServicePrimaryKey(RegisterServiceEntity registerService) {
+        StringBuilder stringBuilder = new StringBuilder();
```
swaroopar, line 168: Primary key is the UUID in the entity. Do we change it to a different one now?

swaroopar (changes requested):
have written some comments after initial review.

runtime/src/main/java/org/eclipse/xpanse/runtime/XpanseApplication.java
baixinsui: It is used  to automatically assign values to database column attributes, such as some creation time and modification time

modules/api/src/main/java/org/eclipse/xpanse/api/OrchestratorApi.java
baixinsui: By calling this service. When we change the list of providers,  we only need to modify the enum on the server side, not the client side.

modules/service/src/main/java/org/eclipse/xpanse/service/impl/DatabaseRegisterServiceStorage.java
baixinsui: This method is used to find the registered service with the same service name, version, service version and service provider in the database when a new service is registering.

modules/api/src/main/java/org/eclipse/xpanse/api/OrchestratorApi.java
baixinsui: Ok，I will set the OCL object into the Response directly.

modules/api/src/main/java/org/eclipse/xpanse/api/OrchestratorApi.java
baixinsui: Ok.

modules/api/src/main/java/org/eclipse/xpanse/api/OrchestratorApi.java
```diff
@@ -44,78 +51,205 @@ public class OrchestratorApi {
...
+    /**
+     * Unregister registered service using id.
+     *
+     * @param id id of registered service.
+     * @return response
+     */
+    @Tag(name = "Service Provider or Service Vendor",
+            description = "APIs unregister registered service.")
+    @Operation(description = "Unregister registered service using id.")
+    @DeleteMapping("/register/service/{id}")
```
iskey, line 191:
```suggestion
    @DeleteMapping("/register/{id}")
```
I think /register/{id} is enought.

modules/api/src/main/java/org/eclipse/xpanse/api/OrchestratorApi.java
```diff
@@ -44,78 +51,205 @@ public class OrchestratorApi {
...
+     * List registered service with query params.
+     *
+     * @param cspName     name of cloud service provider.
+     * @param serviceName name of registered service.
+     * @return response
+     */
+    @Tag(name = "Service Provider or Service Vendor",
+            description = "APIs to list the managed services.")
+    @Operation(description = "List registered service with query params.")
+    @GetMapping(value = "/register/services",
```
iskey, line 216:
```suggestion
    @GetMapping(value = "/register",
```
Did `/register` works here?

modules/api/src/main/java/org/eclipse/xpanse/api/OrchestratorApi.java
```diff
@@ -44,78 +51,205 @@ public class OrchestratorApi {
...
+    /**
+     * Get registered service using id.
+     *
+     * @param id id of registered service.
+     * @return response
+     */
+    @Tag(name = "Service Provider or Service Vendor",
+            description = "APIs to get detail the registered services.")
+    @Operation(description = "Get registered service using id.")
+    @GetMapping(value = "/register/service/{id}",
```
iskey, line 245:
```suggestion
    @GetMapping(value = "/register/{id}",
```

modules/service/src/main/java/org/eclipse/xpanse/service/impl/DatabaseRegisterServiceStorage.java
baixinsui: No. This method is just used to find the registered service with the same service name, version, service version and service provider in the database when a new service is registering.  If the same registered service is found, throw exception to show the service already registered.

modules/api/src/main/java/org/eclipse/xpanse/api/config/YamlJackson2HttpMessageConverterConfiguration.java
```diff
@@ -1,35 +1,35 @@
...
-                    MediaType.parseMediaType("application/yaml"),
-                    MediaType.parseMediaType("application/yml"));
-        }
-    }
-
-    @Bean
-    public AbstractJackson2HttpMessageConverter fastJsonHttpMessageConverters() {
-        return new YamlJackson2HttpMessageConverter();
-    }
+/*
```
iskey, line 1: Do we have anything changed here?

modules/api/src/main/java/org/eclipse/xpanse/api/exceptions/CommonExceptionHandler.java
```diff
@@ -60,20 +63,38 @@ public Response handleConstraintViolationException(ConstraintViolationException
...
+        return Response.errorResponse(ResultCode.BAD_PARAMETERS, failMessage);
     }
 
+    /**
+     * Exception handler for Exception.
+     */
     @ExceptionHandler({Exception.class})
     @ResponseStatus(HttpStatus.NOT_FOUND)
     public Response handleException(Exception ex) {
+        ex.printStackTrace();
```
iskey, line 96: I think we shouldn't keep this, just use a log instead.

iskey (changes requested):

modules/api/src/main/java/org/eclipse/xpanse/api/config/YamlJackson2HttpMessageConverterConfiguration.java
baixinsui: No，I just move it into package config

modules/api/src/main/java/org/eclipse/xpanse/api/exceptions/CommonExceptionHandler.java
baixinsui: Sure. I will delete it.

modules/api/src/main/java/org/eclipse/xpanse/api/OrchestratorApi.java
baixinsui: I will delete service in the URL path.

modules/api/src/main/java/org/eclipse/xpanse/api/OrchestratorApi.java
swaroopar: Ok. in this case. the issue is applicable to all enums. Not just this one. In my opinion, the dependent clients must do an update if needed.

modules/service/src/main/java/org/eclipse/xpanse/service/impl/DatabaseRegisterServiceStorage.java
swaroopar: ok understood. then we can rename the method to something else which does not involve 'primary key'.

modules/api/pom.xml
```diff
@@ -62,5 +62,11 @@
             <artifactId>junit-jupiter-engine</artifactId>
             <scope>test</scope>
         </dependency>
+      <dependency>
+        <groupId>org.eclipse.xpanse.modules</groupId>
+        <artifactId>service</artifactId>
+        <version>1.0.0-SNAPSHOT</version>
```
swaroopar, line 68: we can use the project.version instead of fixed version

swaroopar (changes requested):
comments added.

modules/api/pom.xml
baixinsui: Yes. I will replace the fixed version.

modules/service/src/main/java/org/eclipse/xpanse/service/impl/DatabaseRegisterServiceStorage.java
baixinsui: Yes. I will change the method name.

modules/api/src/main/java/org/eclipse/xpanse/api/OrchestratorApi.java
baixinsui: Now the UI client shows baseed on return data. I remove the API

swaroopar (approved):
Thanks. /LGTM

iskey (approved):
looks good.

Issue #155, opened 2023-02-25
Design and Implement XpanseDeployEngine in the skeleton


iskey:
- [x] XpanseDeployEngine:  Implement the core engine to execute terraform scripts, including preparing the environment、collecting the result、redirect the logs
- [x] XpanseTask: Implement the task of the TFDeployEngine
- [x] XpanseDeployHandler:  Implement a handler to convert the result of the @XpanseDeployEngine to @XpanseResource
- [x] XpanseResource: Design and Implement the XpanseResource

A simple draft design for this:

![image](https://user-images.githubusercontent.com/1907997/223636729-83d469ff-3340-4e3b-b576-c7b360606d50.png)

PR #166, opened 2023-03-06
Add deployment && remove unused codes.

iskey:
related to #155 

- [ ]  XpanseDeployEngine: Implement the core engine to execute terraform scripts, including preparing the environment、collecting the result、redirect the logs [The logs need to refine later.]
- [x]  XpanseTask: Implement the task of the TFDeployEngine.
- [ ]  XpanseDeployHandler: Implement a handler to convert the result of the @XpanseDeployEngine to @XpanseResource
- [x]  XpanseResource: Design and Implement the XpanseResource.[]


modules/api/src/main/java/org/eclipse/xpanse/api/OrchestratorApi.java
```diff
@@ -261,10 +263,10 @@ public SystemStatus health() {
      * @return Status of the managed service.
      */
     @Tag(name = "Service", description = "APIs to manage the service instances")
-    @GetMapping("/service/{managedServiceName}")
+    @GetMapping("/service")
     @ResponseStatus(HttpStatus.OK)
-    public ServiceStatus state(@PathVariable("managedServiceName") String managedServiceName) {
-        return this.orchestratorService.getManagedServiceState(managedServiceName);
+    public ServiceStatus state(@RequestParam("id") String id) {
```
baixinsui, line 268: Is it better to use service/{id} with @PathVariable("id") to keep a unified style？

modules/database/src/main/java/org/eclipse/xpanse/modules/database/service/DeployResourceEntity.java
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * DeployResourceEntity for persistence.
+ */
+@Data
+@Table(name = "deployResource")
+@Entity
+public class DeployResourceEntity {
+
+    @Id
+    private String id;
```
baixinsui, line 31: It is better to replace String with UUID for the "id"

modules/orchestrator/pom.xml
```diff
@@ -52,5 +52,23 @@
             <artifactId>database</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.eclipse.xpanse.modules</groupId>
+            <artifactId>deployment</artifactId>
+            <version>1.0.0-SNAPSHOT</version>
```
baixinsui, line 58: It is better to replace the fixed version with ${project.version}

modules/terraform-deployment/pom.xml
```diff
@@ -0,0 +1,72 @@
...
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.huaweicloud.sdk</groupId>
+            <artifactId>huaweicloud-sdk-ces</artifactId>
+            <version>3.1.26</version>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.xpanse.modules</groupId>
+            <artifactId>deployment</artifactId>
```
baixinsui, line 55: It is better to replace the fixed version with ${project.version}

modules/terraform-deployment/pom.xml
```diff
@@ -0,0 +1,72 @@
...
+            <version>1.0.0-SNAPSHOT</version>
+            <scope>compile</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.xpanse.modules</groupId>
+            <artifactId>database</artifactId>
+            <version>1.0.0-SNAPSHOT</version>
```
swaroopar, line 62: version to be corrected here as well.

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/Deployment.java
```diff
@@ -0,0 +1,19 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * SPDX-FileCopyrightText: Huawei Inc.
+ *
+ */
+
+package org.eclipse.xpanse.modules.deployment;
```
swaroopar, line 7: i feel it is a little confusing that we have two modules - one called deployment and other one called terraform-deployment. Would it be possible to move terraform-deployment under 'deployment' package itself?
for example -  deployment -> deployers->  terraform, pulumi, etc (all deployer types will get a new folder here).

modules/deployment/pom.xml
```diff
@@ -0,0 +1,66 @@
...
+            <artifactId>junit-jupiter-engine</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.huaweicloud.sdk</groupId>
```
swaroopar, line 49: Do we really need this dependency?

modules/database/src/main/java/org/eclipse/xpanse/modules/database/service/DeployResourceEntity.java
```diff
@@ -0,0 +1,47 @@
...
+import jakarta.persistence.Table;
+import java.util.Map;
+import lombok.Data;
+import org.eclipse.xpanse.modules.database.common.DeployResourceKind;
+
+/**
+ * DeployResourceEntity for persistence.
+ */
+@Data
+@Table(name = "deployResource")
```
swaroopar, line 26: just to stick to naming conventions, please make the table name full uppercase. 
DEPLOY_RESOURCE

modules/database/src/main/java/org/eclipse/xpanse/modules/database/service/DeployServiceEntity.java
```diff
@@ -0,0 +1,93 @@
...
+import org.eclipse.xpanse.modules.database.common.CreateModifiedTime;
+import org.eclipse.xpanse.modules.database.common.OclConverter;
+import org.eclipse.xpanse.modules.ocl.loader.data.models.Ocl;
+import org.eclipse.xpanse.modules.ocl.loader.data.models.enums.Csp;
+import org.hibernate.annotations.Type;
+
+/**
+ * DeployServiceEntity for persistence.
+ */
+@Table(name = "deployService")
```
swaroopar, line 37: same here. Table name DEPLOY_SERVICE

plugins/huaweicloud/src/main/java/org/eclipse/xpanse/orchestrator/plugin/huaweicloud/HuaweiTerraformResourceHandler.java
```diff
@@ -0,0 +1,93 @@
...
+            if (!stateFile.getClass().equals(String.class)) {
+                throw new RuntimeException("stateFile is unsupported.");
+            }
+            tfState = objectMapper.readValue(stateFile.toString(), TfState.class);
+        } catch (IOException ex) {
+            log.error("Parse terraform state content failed.");
+            throw new TerraformExecutorException("Parse terraform state content failed.", ex);
+        }
+        for (TfStateResource tfStateResource : tfState.getResources()) {
+            DeployResourceEntity deployResourceEntity = new DeployResourceEntity();
```
swaroopar, line 54: If i understood it right, we are reading the result of the TF execution and then writing data entries. Should it be the other way round? that we write DB entities first when the request is received and later only update the status of it based on the result?

plugins/huaweicloud/src/main/java/org/eclipse/xpanse/orchestrator/plugin/huaweicloud/HuaweiTerraformResourceHandler.java
```diff
@@ -0,0 +1,93 @@
...
+                    DeployResourceEntity xpResource = new DeployResourceEntity();
+                    xpResource.setKind(DeployResourceKind.Vpc);
+                    xpResource.setId((String) instance.getAttributes().get("id"));
+                    xpResource.setName((String) instance.getAttributes().get("name"));
+                }
+            }
+            xpResourceList.add(deployResourceEntity);
+        }
+
+        deployResult.setResources(xpResourceList);
```
swaroopar, line 91: the entities which we created are not saved to the DB here?

modules/ocl-loader/src/main/java/org/eclipse/xpanse/modules/service/CreateRequest.java
```diff
@@ -0,0 +1,70 @@
...
+     * The flavor of the Service.
+     */
+    @NotNull
+    @NotBlank
+    String flavor;
+
+    /**
+     * The property of the Service.
+     */
+    @ElementCollection
```
swaroopar, line 64: is this bean also a DB entity?

modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/OrchestratorService.java
```diff
@@ -78,278 +77,76 @@ public void onApplicationEvent(ApplicationEvent event) {
...
+        DeployServiceEntity deployServiceEntity = new DeployServiceEntity();
+        deployServiceEntity.setVersion(deployResult.getTask().getCreateRequest().getVersion());
+        deployServiceEntity.setName(deployResult.getTask().getCreateRequest().getName());
+        deployServiceEntity.setId(deployResult.getTask().getCreateRequest().getId());
+        deployServiceEntity.setCsp(deployResult.getTask().getCreateRequest().getCsp());
+        deployServiceEntity.setFlavor(deployResult.getTask().getCreateRequest().getFlavor());
+        deployServiceEntity.setProperty(deployResult.getTask().getCreateRequest().getProperty());
+        deployServiceEntity.setOcl(deployResult.getTask().getOcl());
+        deployServiceEntity.setDeployResourceEntity(deployResult.getResources());
+        this.entityManager.persist(deployServiceEntity);
```
swaroopar, line 94: we already have the "store" methods in the JPA interface class. We can use that instead of the entity manager directly.

modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/OrchestratorService.java
```diff
@@ -78,278 +77,76 @@ public void onApplicationEvent(ApplicationEvent event) {
...
-                this.databaseOrchestratorStorage.store(serviceStatusEntity);
-            } catch (RuntimeException exception) {
-                updateFailedServiceStatusEntity(serviceStatusEntity, exception);
-                this.databaseOrchestratorStorage.store(serviceStatusEntity);
-                throw exception;
-            }
-        }
+    public void storeDeployService(DeployResult deployResult) {
+        DeployServiceEntity deployServiceEntity = new DeployServiceEntity();
+        deployServiceEntity.setVersion(deployResult.getTask().getCreateRequest().getVersion());
```
swaroopar, line 86: same here. Should we write the DB entities before execution and later only update the state based on the result?

plugins/huaweicloud/src/main/java/org/eclipse/xpanse/orchestrator/plugin/huaweicloud/HuaweiTerraformResourceHandler.java
iskey: yes, not here. the entities should be saved in the skeleton.

modules/database/src/main/java/org/eclipse/xpanse/modules/database/service/DeployResourceEntity.java
iskey: okie, I have some database trouble here, let's fix this in the next PR.

modules/ocl-loader/src/main/java/org/eclipse/xpanse/modules/service/CreateRequest.java
iskey: no, let me delete these database annotations.

modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/OrchestratorService.java
iskey: The "store" methods will not create the related tables automatically. I must store the different tables separately using the "store" method.

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/Deployment.java
iskey: agree.

modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/OrchestratorService.java
swaroopar: oh ok. Ideally the DB tables should be created with different standalone scripts (which we don't have and also not needed at the moment since we are just using H2 DB)

swaroopar (approved):
can be merged. Other open points to be  implemented with further PRs.

Issue #167, opened 2023-03-06
Implement the OCLv2 Service database model


iskey:
- [x] implement the OCLv2 service database models, such as service/resources/...
- [x] integrate with the deployment part.
- [x] implement the service database interface.

iskey:
fixed by #170

PR #169, opened 2023-03-06
Add category to OCL && do some small fix.

iskey:
related to #167


modules/database/src/main/java/org/eclipse/xpanse/modules/database/service/DeployResourceEntity.java
```diff
@@ -30,14 +30,26 @@ public class DeployResourceEntity {
     @Id
```
swaroopar, line 30: can we remove this JPA annotation here?

swaroopar (changes requested):
one minor comment added.

modules/database/src/main/java/org/eclipse/xpanse/modules/database/service/DeployResourceEntity.java
iskey: yes, we can remove this.
I shouldn't follow the IDE's suggestion with "xxx should have primary key ".

swaroopar (approved):
/LGTM

baixinsui:
look good.

Issue #157, opened 2023-02-25
Refine and implement the Database model for the OCLv2


iskey:
- [ ] Need ORM？ import new middleware plugin? (need more discussion..)
- [ ] Serialize/deserialize for the OCLv2 object.
- [ ] Serialize/deserialize for the  DeployEngines.

PR #170, opened 2023-03-07
Async deploy service and destroy service && remove module named service

baixinsui:
related to #157 , #167


modules/ocl-loader/src/main/java/org/eclipse/xpanse/modules/ocl/loader/data/models/query/RegisteredServiceQuery.java
```diff
@@ -16,14 +16,17 @@
  */
 @Data
 @Valid
-public class RegisterServiceQuery {
+public class RegisteredServiceQuery {
 
     @Schema(description = "Name of the cloud service provider.")
-    private String cspName;
+    private String csp;
```
swaroopar, line 22: can this be the `Csp` Enum instead of string?

swaroopar (changes requested):
Just one minor comment. rest looks good.

modules/ocl-loader/src/main/java/org/eclipse/xpanse/modules/ocl/loader/data/models/query/RegisteredServiceQuery.java
baixinsui: Ok

iskey (approved):
looks good.

swaroopar (approved):
/LGTM

PR #171, opened 2023-03-07
Update manage service APIs and Add query APIs

### LINKED ISSUE #157: Refine and implement the Database model for the OCLv2

baixinsui:
fix #157


modules/ocl-loader/src/main/java/org/eclipse/xpanse/modules/ocl/loader/data/models/vo/CategoryOclVo.java
```diff
@@ -0,0 +1,21 @@
...
+import java.util.List;
+import lombok.Data;
+
+/**
+ * Define view object for UI Client query registered service by category.
+ */
+@Data
+public class CategoryOclVo {
+
+    private String name;
```
swaroopar, line 18: please also add descriptions to fields here.

modules/ocl-loader/src/main/java/org/eclipse/xpanse/modules/ocl/loader/data/models/vo/CategoryOclVo.java
```diff
@@ -0,0 +1,21 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * SPDX-FileCopyrightText: Huawei Inc.
+ *
+ */
+
+package org.eclipse.xpanse.modules.ocl.loader.data.models.vo;
```
swaroopar, line 7: i think it is better to call the package with full name instead of short form `vo`. It is then clear for everyone.

modules/ocl-loader/src/main/java/org/eclipse/xpanse/modules/ocl/loader/data/models/vo/ProviderOclVo.java
```diff
@@ -0,0 +1,30 @@
...
+import org.eclipse.xpanse.modules.ocl.loader.data.models.enums.Csp;
+
+
+/**
+ * Define view object for UI Client query registered service by csp.
+ */
+@Data
+public class ProviderOclVo {
+
+    @Schema(description = "The Cloud Service Provider. valid values: aws, azure, alibaba, huawei")
```
swaroopar, line 21: we can avoid providing enum values in description again. There are few other places also where we are doing it. we can remove it everywhere.

baixinsui: OK. I removed the enum values from description.

modules/api/src/main/java/org/eclipse/xpanse/api/OrchestratorApi.java
```diff
@@ -65,6 +67,21 @@ public OrchestratorApi(OrchestratorService orchestratorService,
...
+     *
+     * @return response
+     */
+    @Tag(name = "Service Vendor",
+            description = "APIs to manage register services.")
+    @Operation(description = "Get category list.")
+    @GetMapping(value = "/register/categories",
+            produces = MediaType.APPLICATION_JSON_VALUE)
+    @ResponseStatus(HttpStatus.OK)
+    public Response<List<Category>> listCategories() {
```
swaroopar, line 81: hi @baixinsui we must discuss why we need to wrap actual response object additionally in `Response` object. It makes sense to return only the actual response object.

swaroopar (changes requested):
comment added

modules/api/src/main/java/org/eclipse/xpanse/api/OrchestratorApi.java
iskey: I think the original idea is to make the sider menubar items be determined by the backend categories.

iskey: But In my view, fixed menu items are also ok for us.

modules/ocl-loader/src/main/java/org/eclipse/xpanse/modules/ocl/loader/data/models/view/VersionOclVo.java
```diff
@@ -0,0 +1,25 @@
+/*
```
iskey, line 1: Can we move the view models to the orchestrator module? seems only the orchestrator will consume these view objects.

modules/api/src/main/java/org/eclipse/xpanse/api/OrchestratorApi.java
baixinsui: All interface return the Response object, which makes the client  operation of data more consistent.

modules/ocl-loader/src/main/java/org/eclipse/xpanse/modules/ocl/loader/data/models/view/VersionOclVo.java
baixinsui: Sure. It's a good idea.

modules/api/src/main/java/org/eclipse/xpanse/api/OrchestratorApi.java
baixinsui: If the enums of Category change, now we only do some change in backend, not both backend and front.

swaroopar: we dont need all interfaces to return response object. Thats the benefit of micro services. Each operation can return its own response object. At the moment i dont see any benefit in wrapping all actual return objects in Response object.

modules/ocl-loader/src/main/java/org/eclipse/xpanse/modules/ocl/loader/data/models/view/VersionOclVo.java
swaroopar: but these are also exposed in the API?
I thought they are only consumed by the GUI. Is that not true?

baixinsui: Yes. they are only consumed by the GUI.

swaroopar: a little tricky case then. it does not belong to OCL nor to orchestrator. Can we put it in the 'api' module?

baixinsui: We can't move these into  the 'api' module, which will case  Circular dependency

modules/api/src/main/java/org/eclipse/xpanse/api/OrchestratorApi.java
baixinsui: Is there a unified moudle in front client for calling API and processing return data not yet?  and we will not create a moudle with this function?

swaroopar: the client knows which API method returns what type if we define it correctly in the open api

modules/ocl-loader/src/main/java/org/eclipse/xpanse/modules/ocl/loader/data/models/view/VersionOclVo.java
baixinsui: It better to rename the moule  'ocl-loader', which contains all enums, querymodel, view model. Do you have some good idea?

swaroopar: may be  `types` ?

baixinsui: How about 'models' ?

swaroopar: yes that is also fine.

baixinsui: Ok. I will rename the moule 'ocl-loader' to 'models' to put all kind data object.

swaroopar: sure. we can have a sub package called `ocl` under it and we can put all types related to OCL in it

modules/api/src/main/java/org/eclipse/xpanse/api/OrchestratorApi.java
```diff
@@ -263,10 +313,8 @@ public Response detail(
     @Tag(name = "Admin", description = "APIs for administrating Xpanse")
     @GetMapping("/health")
     @ResponseStatus(HttpStatus.OK)
-    public SystemStatus health() {
```
swaroopar, line 266: we used SystemStatus so that we  can add additional fields to it. For now we just have health status.

swaroopar (changes requested):
to revert change to SystemStatus

modules/api/src/main/java/org/eclipse/xpanse/api/OrchestratorApi.java
baixinsui: Ok.I will recover it.

iskey:
@baixinsui , @swaroopar , How about we separate these changes into some small PRs?Maybe the `ocl-loader` refactor can be one of them.

baixinsui:
[#Refine and implement the Database model for the OCLv2](https://github.com/huaweicloud/xpanse/issues/157)

swaroopar:
@iskey it is fine I think for now. it will be extra work now and we need the new API method for the other UI user story as soon as possible.

modules/api/src/main/java/org/eclipse/xpanse/api/response/Response.java
```diff
@@ -26,7 +29,7 @@ public class Response {
     @Schema(description = "The success boolean of response.")
     private Boolean success;
     @Schema(description = "The result data of response.")
-    private Object data;
+    private T data;
```
swaroopar, line 32: Looks like we lost the important changes.

swaroopar (changes requested):
earlier changes missing.

modules/api/src/main/java/org/eclipse/xpanse/api/response/Response.java
baixinsui: I reset the commit and push force. Now, it is ok.

swaroopar (approved):
/LGTM

PR #194, opened 2023-03-14
Add validation for models && check validation of deploy property

baixinsui:


modules/models/src/main/java/org/eclipse/xpanse/modules/models/service/DeployResource.java
```diff
@@ -19,18 +22,26 @@ public class DeployResource {
     /**
      * The ID of the deployed resource.
      */
+    @NotBlank
```
swaroopar, line 25: for all strings, please also add `@notNull`

modules/models/src/main/java/org/eclipse/xpanse/modules/models/resource/DeployVariable.java
```diff
@@ -19,28 +18,25 @@
...
-
     @NotNull
     @Schema(description = "The kind of the deploy variable")
     private DeployVariableKind kind;
 
-    @NotNull
     @NotBlank
-    @NotEmpty
     @Schema(description = "The type of the deploy variable")
     private String type;
```
swaroopar, line 31: can we change the type from String to `DeployVariableType`

modules/models/src/main/java/org/eclipse/xpanse/modules/models/utils/DeployVariableValidator.java
```diff
@@ -0,0 +1,317 @@
...
+
+
+    /**
+     * Check validation of deploy property map by list of deployVariables in registered service.
+     *
+     * @param deployVariables list of deployVariables in registered service
+     * @param deployProperty  deploy property map
+     * @return isValid
+     */
+    public boolean checkVariableValidation(List<DeployVariable> deployVariables,
```
swaroopar, line 193: we can rename the method to `isVariableValid`

modules/models/src/main/java/org/eclipse/xpanse/modules/models/utils/DeployVariableValidator.java
```diff
@@ -0,0 +1,317 @@
...
+            Map<String, String> deployProperty) {
+
+        if (!CollectionUtils.isEmpty(deployVariables) && Objects.nonNull(deployProperty)) {
+            // check required keys
+            Set<String> userPutKeys = deployProperty.keySet();
+            log.error("userPutKeys:{}", userPutKeys);
```
swaroopar, line 199: is log level error correct here? May be DEBUG is better?

modules/models/src/main/java/org/eclipse/xpanse/modules/models/service/DeployResource.java
baixinsui: Ok.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/resource/DeployVariable.java
baixinsui: Ok.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/utils/DeployVariableValidator.java
baixinsui: Ok.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/utils/DeployVariableValidator.java
baixinsui: Yes.

modules/models/src/test/resources/test.yaml
```diff
@@ -50,48 +52,72 @@ deployment:
   # - variable: The value of the variable parameters are defined by the user on the console.
   # The parameters will be used to generate the API of the managed service.
   context:
```
swaroopar, line 54: i realised that the field name"context" is not clear. Can we think of any other name?

swaroopar (changes requested):
comments added.

modules/models/src/test/resources/test.yaml
baixinsui: varibale?

iskey: yes, it is not clear enough; 
I think we can create an issue for the new name, leaving it as it is in this PR.

iskey: I add #195 for this.

iskey (approved):
looks good.

modules/models/src/test/resources/test.yaml
```diff
@@ -50,48 +52,72 @@ deployment:
...
+      type: string
+      mandatory: true
+      validator:
+    - name: secgroup_id
+      description: The secgroup id.
       kind: variable
       type: string
       mandatory: false
-      validator: length(1-256)
+      validator:
```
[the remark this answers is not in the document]
iskey: How about write some descriptions in the readme.txt for the validator?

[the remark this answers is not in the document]
baixinsui: I added description for every validator in the readme.md

PR #206, opened 2023-03-22
Format terraform codes in README

niuzhenguo:
1. Use HCL code format indentation
2. Use vps, subnets, and secgroups datasource instead
3. Get rid of dns param in subnet as it will be automatically added


swaroopar (approved):
looks good.

PR #198, opened 2023-03-15
Add a area level for regions

WangLiNaruto:
[#181 ](url)

PR #210, opened 2023-03-24
update area

WangLiNaruto:


iskey (changes requested):
Please also fix the examples in the Readme.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/resource/Region.java
```diff
@@ -0,0 +1,20 @@
...
+import lombok.Data;
+
+/**
+ * The regions of the Cloud Service Provider.
+ */
+@Data
+public class Region {
+
+    private String name;
+    private String area;
```
swaroopar, line 18: can you also please add open API description and also validations for both fields

swaroopar (changes requested):
OpenAPI details to be added.

swaroopar:
Also what is the difference between this PR and #198

iskey:
> Also what is the difference between this PR and #198

#198 is another way to implement the `area` thoughts.
It should be closed now for we decide to add the area as a field of `Region`.

swaroopar (approved):
looks good to me.

iskey (approved):
looks good.

PR #223, opened 2023-03-29
Add FlexibleEngine plugin and service samples

baixinsui:


baixinsui:
@niuzhenguo Request review

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/deployers/terraform/TerraformProviders.java
```diff
@@ -8,6 +8,7 @@
 
 import java.util.Objects;
 import org.eclipse.xpanse.modules.deployment.deployers.terraform.providers.Aws;
+import org.eclipse.xpanse.modules.deployment.deployers.terraform.providers.Flexible;
```
niuzhenguo, line 11: better to rename this to FlexibleEngine to be more readable.

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/deployers/terraform/TerraformProviders.java
```diff
@@ -16,7 +17,8 @@
  */
 public enum TerraformProviders {
     AWS(Csp.AWS, new Aws()),
-    HUAWEI(Csp.HUAWEI, new Huawei());
+    HUAWEI(Csp.HUAWEI, new Huawei()),
+    FLEXIBLE(Csp.FLEXIBLE, new Flexible());
```
niuzhenguo, line 21: ditto.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/enums/Csp.java
```diff
@@ -18,7 +18,8 @@ public enum Csp {
     AZURE("azure"),
     ALI("alibaba"),
     HUAWEI("huawei"),
-    OPENSTACK("openstack");
+    OPENSTACK("openstack"),
+    FLEXIBLE("flexible");
```
iskey, line 22: FLEXIBLEENGINE("flexibleEngine") I am going to do the same change on the xpanse-ui.

swaroopar: Better as FLEXIBLE_ENGINE.. with underscore separated.

iskey: yes, I believe `FLEXIBLE_ENGINE` is better.

baixinsui: Ok. I renamed it to FLEXIBLE_ENGINE

iskey (approved):
looks good to me.

PR #226, opened 2023-03-29
use temp folder for terraform workspace

swaroopar:
1. Create workspace in temp directory.
2. Move workspace directory name to properties file.

### DIFF
--- a/modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/deployers/terraform/TerraformDeployment.java
+++ b/modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/deployers/terraform/TerraformDeployment.java
@@ -22,6 +22,7 @@ import org.eclipse.xpanse.modules.models.enums.TerraformExecState;
 import org.eclipse.xpanse.modules.models.resource.DeployVariable;
 import org.eclipse.xpanse.modules.models.resource.Flavor;
 import org.eclipse.xpanse.modules.models.service.DeployResult;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Component;
 
 /**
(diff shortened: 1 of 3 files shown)


modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/deployers/terraform/TerraformDeployment.java
```diff
@@ -34,7 +35,13 @@ public class TerraformDeployment implements Deployment {
     public static final String VERSION_FILE_NAME = "version.tf";
     public static final String SCRIPT_FILE_NAME = "resources.tf";
 
-    private static final String WORKSPACE = "xpanse_deploy_ws";
+
+    private final String workspaceDirectory;
+
+    public TerraformDeployment(
+            @Value("${terraform.workspace.directory}") String workspaceDirectory) {
```
baixinsui, line 42: We can use @Value("${terraform.workspace.directory:xpanse_deploy_ws}") set a default value

swaroopar: @baixinsui  the default is already there in application.properties. 
My idea was to keep all default in application.properties and override it with any specific value via environment variables or via plugin specific property files. 
Do you see any issue with this approach?

baixinsui: It’s OK. Set the default value just in case of losing the configuration item in application.properties.

swaroopar: @baixinsui ok got it. done. Have added it now.

baixinsui (approved):

iskey (approved):
looks good.

PR #231, opened 2023-03-31
Add check parameters of the update service.

### LINKED ISSUE #228: service update - validate and allow only fields that can be updated.
Currently the update simply overrides the complete OCL of an already registered service. 
We need to define attributes that can be updated and simply reject if the update is changing even basic things in the OCL.
Eg., version, name cannot be updated. This will break other things. 
1. Discuss what properties can be updated. 
2. Add validation to the update method.

WangLiNaruto:
fixed #228
image

### DIFF
--- a/modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/register/impl/RegisterServiceImpl.java
+++ b/modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/register/impl/RegisterServiceImpl.java
@@ -9,10 +9,12 @@ package org.eclipse.xpanse.orchestrator.register.impl;
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.UUID;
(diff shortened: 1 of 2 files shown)


modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/register/impl/RegisterServiceImpl.java
```diff
@@ -68,14 +70,45 @@ public void updateRegisteredServiceByUrl(String id, String oclLocation) throws E
     public void updateRegisteredService(String id, Ocl ocl) {
         RegisterServiceEntity existedService = storage.getRegisterServiceById(UUID.fromString(id));
         if (Objects.isNull(existedService)) {
+            log.error("Registered service with id {} not existed.", id);
```
swaroopar, line 73: `not existing`

swaroopar (approved):
implementation looks good. just the error message and code must be updated.

swaroopar (changes requested):
comments added.

swaroopar (approved):

PR #235, opened 2023-03-31
Replace DB entities with View objects in REST API

### LINKED ISSUE #233: Replace DB entities with View objects in REST API
Currently DB entities are directly exposed on the REST API. 
We must replace this with view objects and add all OpenAPI annotations  - Eg., schema, bean validations, etc

baixinsui:
fix #233


modules/models/src/main/java/org/eclipse/xpanse/modules/models/enums/ResourceKindInstance.java
```diff
@@ -0,0 +1,49 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * SPDX-FileCopyrightText: Huawei Inc.
+ *
+ */
+
+package org.eclipse.xpanse.modules.models.enums;
```
iskey, line 7: I suggest create a new PR for this.

swaroopar (changes requested):
looks good. but as Yichao mentioned, better to keep not related changes outside this PR.

swaroopar (approved):
looks good to me.

PR #238, opened 2023-04-03
Rename alicloud to be more specific

niuzhenguo:


modules/models/src/main/java/org/eclipse/xpanse/modules/models/resource/CloudServiceProvider.java
```diff
@@ -21,7 +21,7 @@
 public class CloudServiceProvider {
 
     @NotNull
-    @Schema(description = "The Cloud Service Provider. valid values: aws, azure, alibaba, huawei")
+    @Schema(description = "The Cloud Service Provider. valid values: aws, azure, alicloud, huawei, flexibleEngine")
```
swaroopar, line 24: we had actually discussed to remove the name from the description. Looks like it is still there in this one place. So we can remove here too.
The list of valid CSPs is anyways in the enum.

niuzhenguo: done

iskey (approved):
looks good.

PR #241, opened 2023-04-03
Add some deploy variable kinds.

### LINKED ISSUE #236: Refine the DeployVariableKind for deployment.
I think the `fix_env,fix_variable,env,variable` is not enough for the deployment, I want to refine the variable kinds like this：
```
fix_env: This kind of variables must be settled in the OCL template, and will be injected to the deployment as environment variables.
fix_variable: This kind of variables must be settled in the OCL template, and will be injected to the deployment as variables.

env: This kind of variables can be settled by the API or the xpanse runtime environment(The API takes higher priority), and will be injected to the deployment as environment variables.
variable: This kind of variables can be settled by the API or the xpanse runtime environment(The API takes higher priority), and will be injected to the deployment as variables.

env_env: This kind of variables must be settled by the xpanse runtime environment, and will be injected to the deployment as environment variables.
env_variable: This kind of variables must be settled by the xpanse runtime environment, and will be injected to the deployment as variables.
```
@swaroopar  thoughts?

iskey:
fix #236


README.md
```diff
@@ -92,9 +92,13 @@ flavors:
 deployment:
   # kind, Supported values are terraform, pulumi, crossplane.
   kind: terraform
-  # Context for deployment: the context including some kind of parameters for the deployment, such as fix,variable.
-  # - env: The value of the fix parameters are defined by the ISV with the @value at the initial time.
-  # - variable: The value of the variable parameters are defined by the user on the console.
+  # Context for deployment: the context including some kind of parameters for the deployment, such as fix_env, fix_variable, env, variable, env_env, env_variable.
```
swaroopar, line 95: Based on our discussion, please see if we can replace the description as below.

- FIX_ENV:  Values for variable of this type are defined by the managed service provider in the OCL template. Runtime will inject it to deployer as enviroment variables. This varible is not visible to the end user. 

- FIX_VAR: Values for variable of this type are defined by the managed service provider in the OCL template. Runtime will inject it to deployer as usual variables. This varible is not visible to the end user. 

- ENV: Value for a variable of this type can be provided by end user. If marked as mandatory then end user must provide value to this variable. If marked as optional and if end user does not provided it,  then the fallback value to this variable is read by runtime (it can read from other sources, e.g., OS env variables). This variable is injected as a  environment variable to the deployer. 

- VAR: Value for a variable of this type can be provided by end user. . If marked as mandatory then end user must provide value to this variable. If marked as optional and if end user does not provided it, then the fallback value to this variable is read by runtime (it can read from other sources, e.g., OS env variables). This variable is injected as a  regular variable to the deployer.

- ENV_ENV:  Value to this variable is read by runtime (it can read from other sources, e.g., OS env variables) and injected as a  environment variable to the deployer. End user cannot see or change this variable.

- ENV_VARIABLE: Value to this variable is read by runtime (it can read from other sources, e.g., OS env variables) and injected as a  regular variable to the deployer. End user cannot see or change this variable.

swaroopar (changes requested):
@iskey Thank you for thinking through all possible combinations. I have just  updated the enum description based on my understanding. Please see if it makes sense to you.

swaroopar (approved):
looks good to me.

PR #249, opened 2023-04-04
Transform TFResource to DeployResource in all plugins

### LINKED ISSUE #243: Redefine TFResource transform into DeployResource.

baixinsui:
fix #243


swaroopar (approved):
looks good to me.

PR #252, opened 2023-04-05
Feature/customerservicename

swaroopar:
- [x]  Add customer service name to API and DB


modules/api/src/main/java/org/eclipse/xpanse/api/OrchestratorApi.java
```diff
@@ -417,4 +422,9 @@ public String openApi(@PathVariable("id") String id) {
         return apiUrl;
     }
 
+    private String generateCustomerServiceName(CreateRequest createRequest) {
+        return createRequest.getServiceName().substring(0, 4) + "-"
```
iskey, line 426: what will happen if the service name is shorter than 4?

swaroopar: @iskey good point. Fixed this.

PR #259, opened 2023-04-06
Generate openApi for registered services

### LINKED ISSUE #158: Refine the API Gateway part
- [ ] which will be included in the API Gateway (need more discussion).
- [ ] Implement the Query part of the new services, which will be used by the end-user.
- [ ] Dynamically generates the APIs documents.

baixinsui:
fix #158


modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/utils/OpenApiUtil.java
```diff
@@ -0,0 +1,346 @@
...
+                        "description": "Generated server url"
+                      }
+                    ],
+                    "paths": {
+                      "/xpanse/service": {
+                        "post": {
+                          "tags": [
+                            "Service"
+                          ],
+                          "operationId": "start",
```
swaroopar, line 150: should the operation be "deploy" instead?

modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/utils/OpenApiUtil.java
```diff
@@ -0,0 +1,346 @@
...
+                    deployVariableValidator.getRequiredKeySet(deployVariables));
+        } catch (JsonProcessingException e) {
+            log.error("Failed to write value as string.", e);
+        }
+        String apiJson = String.format("""
+                {
+                    "openapi": "3.0.1",
+                    "info": {
+                      "title": "OpenAPI definition",
+                      "version": "v0"
```
swaroopar, line 136: can this be the service version?

modules/api/src/main/java/org/eclipse/xpanse/api/OrchestratorApi.java
```diff
@@ -412,12 +411,12 @@ public Response destroy(@PathVariable("id") String id) {
      */
     @Tag(name = "Service Vendor",
             description = "APIs to manage register services.")
-    @GetMapping(value = "/service/openapi/{id}", produces = MediaType.APPLICATION_JSON_VALUE)
+    @GetMapping(value = "/register/openapi/{id}", produces = MediaType.APPLICATION_JSON_VALUE)
     @ResponseStatus(HttpStatus.OK)
     @Operation(description = "API to get openapi of service deploy context")
     public String openApi(@PathVariable("id") String id) {
```
swaroopar, line 417: Instead of generating the swagger html file via this REST API method which is always an extra step, would it make sense to do the following
1. Always generate the API as part of the service register request.
2. Return the URL as part of the "register" response object.
3. Also add the URL to the other VO objects where the registered services are displayed. 

Additionally we must also generate these files files during the service startup if the files are missing. 
@iskey whats your opinion on this?

modules/api/src/main/java/org/eclipse/xpanse/api/OrchestratorApi.java
```diff
@@ -412,12 +411,12 @@ public Response destroy(@PathVariable("id") String id) {
      */
     @Tag(name = "Service Vendor",
             description = "APIs to manage register services.")
-    @GetMapping(value = "/service/openapi/{id}", produces = MediaType.APPLICATION_JSON_VALUE)
+    @GetMapping(value = "/register/openapi/{id}", produces = MediaType.APPLICATION_JSON_VALUE)
```
swaroopar, line 414: produces JSON -> but return type is string.

swaroopar (changes requested):
@baixinsui very neatly done. Please have a look at the comments and fix if they are fine.

modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/utils/OpenApiUtil.java
```diff
@@ -0,0 +1,346 @@
...
+                            "name",
+                            "region",
+                            "version"
+                          ],
+                          "type": "object",
+                          "properties": {
+                            "category": {
+                              "type": "string",
+                              "description": "The category of the service",
+                              "enum": [
```
swaroopar, line 250: can this list be also read dynamically?

modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/utils/OpenApiUtil.java
```diff
@@ -0,0 +1,346 @@
...
+                            }
+                          }
+                        },
+                        "property": {
+                          "required": %s,
+                          "type": "object",
+                          "properties": %s
+                        },
+                        "CreateRequest": {
+                          "required": [
```
swaroopar, line 237: can this list be also read dynamically?

modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/utils/OpenApiUtil.java
```diff
@@ -0,0 +1,346 @@
...
+                              "description": "The version of service"
+                            },
+                            "region": {
+                              "type": "string",
+                              "description": "The region of the provider."
+                            },
+                            "csp": {
+                              "type": "string",
+                              "description": "The csp of the Service.",
+                              "enum": [
```
swaroopar, line 278: can this list be also read dynamically?

swaroopar (changes requested):
comments added.

modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/utils/OpenApiUtil.java
baixinsui: Yes.I will replace the 'start' with 'deploy'

modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/utils/OpenApiUtil.java
baixinsui: Yes. I will fill the field 'version' with service version.

modules/api/src/main/java/org/eclipse/xpanse/api/OrchestratorApi.java
baixinsui: I will remove the produces config.

modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/utils/OpenApiUtil.java
baixinsui: Yes. I will add a method to get the list of fields in the Class 'CreateRequest' which has @NotNull.

modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/utils/OpenApiUtil.java
baixinsui: Yes. I will get the list enum of Csp and fill the list here.

modules/api/src/main/java/org/eclipse/xpanse/api/OrchestratorApi.java
iskey: Is it possible to redirect the static pages of the API document?
My initial idea is to make this api as the endpoint of the APIs.

modules/api/src/main/java/org/eclipse/xpanse/api/OrchestratorApi.java
iskey: Hi, Swaroop, my initial idea is to make the API part as an independence with others. These swagger pages are not part of the OCL. They are derivations of the OCL. so, from my point of view, they should like this:
  1. `register` will do not care about these swagger pages. 
  2. the REST API will redirect to the swagger pages when the swagger files existing.
  3. if the swagger files not existed, the REST API will trigger to generate the swagger files, and then redirect to them. (lazy loading)
 
If we do it this way, we will restrict the swagger files known to this API only, and it will be graceful and reliable. (perhaps with a little late responding for the first time)

iskey (commented):
comments added.

modules/api/src/main/java/org/eclipse/xpanse/api/OrchestratorApi.java
baixinsui: I will redirect the url.

modules/api/src/main/java/org/eclipse/xpanse/api/OrchestratorApi.java
swaroopar: Hi @iskey 
Ok agreed.  In this case, i would still prefer to return the REST API link in the register response and the GET registered services response body. Then the user knows that such a service exists and can use it. What do you think_

baixinsui: Referring to your two opinions，I made the following modifications:
1.  In the end of  register service request I add @Async method to create API file，add @Async method  to delete API  file in the end of unregister service request, add @Async method  to delete API  file in the end of unregister service request. 
2. The  service /xpanse/opanApi/{id} will redirect to the existed html file or the new created html file(when the swagger html file not existed)
3. I add a method to get openApiUrl by serviceId. If  adding the field 'openApiUrl' into VOs  is  needed, I can call this method to get value and fill the the field 'openApiUrl' in all VOs.

swaroopar: @baixinsui great. this is good. 
Yes, I feel it is good to have the URL returned in register response body (we can display then this as a link in the register UI also) and in the GET response bodies. 
Need not be in the same PR. If needed, we can do it in a different one.

baixinsui: Ok. Please create a new issue then assign it to me. I will do it next week by a new PR

iskey: @swaroopar, The scene in my mind is that, the swagger static files will be a well-known URL with the register service uuid, so it is no need to make the URL part of the API response, and the real position will be volatile.
someday, if we have a web proxy such as Nginx, they will provide these static pages for us, then the API to generate the swagger files will be triggered by the access missing of the Nginx.

swaroopar: @iskey yes the host of the URL will change if we plan to access the API via the load balancer. 
We must check about using the HATEOS concept. Spring also has some support for this. If HATEOS cannot handle moving host names, then just the REST URI is enough. 
I will add this in the new issue.

swaroopar (approved):
looks good to me.

modules/api/src/main/java/org/eclipse/xpanse/api/OrchestratorApi.java
iskey: ok, thank you @swaroopar, so kind of you. we can merge this PR for now.

PR #266, opened 2023-04-10
Add Link OpenApiUrl in Registered Service View Object with HATEOS

### LINKED ISSUE #265: Return OpenAPI URL of the service in response objects with HATEOS
Return the OpenAPI URL of the services in register response and also in all other relevant view objects.
1. These fields must be links according to HATEOS concept. 
2. Check if spring has support of handling links in HATEOS based fields when the REST API is hosted behind load balancers and add required configuration for it..

baixinsui:
fix #265

### DIFF
--- a/modules/models/src/main/java/org/eclipse/xpanse/modules/models/view/RegisteredServiceVo.java
+++ b/modules/models/src/main/java/org/eclipse/xpanse/modules/models/view/RegisteredServiceVo.java
@@ -14,16 +14,19 @@ import jakarta.validation.constraints.NotNull;
 import java.util.Date;
 import java.util.UUID;
 import lombok.Data;
+import lombok.EqualsAndHashCode;
 import org.eclipse.xpanse.modules.models.enums.Category;
 import org.eclipse.xpanse.modules.models.enums.Csp;
 import org.eclipse.xpanse.modules.models.enums.ServiceState;
 import org.eclipse.xpanse.modules.models.resource.Ocl;
+import org.springframework.hateoas.RepresentationModel;
 
 /**
- * Define view object for UI Client query registered service.
+ * Defines view object for query registered service.
  */
 @Data
-public class RegisteredServiceVo {
+@EqualsAndHashCode(callSuper = true)
+public class RegisteredServiceVo extends RepresentationModel<RegisteredServiceVo> {
 
     @NotNull
     @Schema(description = "ID of the registered service.")
@@ -44,25 +47,25 @@ public class RegisteredServiceVo {
     private Csp csp;
 
     @NotNull
-    @Schema(description = "ID of the registered service.")
+    @Schema(description = "Category of the registered service.")
     private Category category;
+
     @NotNull
-    @Schema(description = "ID of the registered service.")
+    @Schema(description = "Ocl model of the registered service.")
     private Ocl ocl;
 
     @NotNull
-    @Schema(description = "Time of register service.")
+    @Schema(description = "createTime of the registered service.")
     @JsonFormat(pattern = "yyyy-MM-dd HH:mm:ss")
     private Date createTime;
 
     @NotNull
     @JsonFormat(pattern = "yyyy-MM-dd HH:mm:ss")
-    @Schema(description = "Time of update service.")
+    @Schema(description = "Last updateTime of the registered service.")
     private Date lastModifiedTime;
 
     @NotNull
     @Schema(description = "State of service.")
     private ServiceState serviceState;
-
 
 }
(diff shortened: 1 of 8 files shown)


modules/models/src/main/java/org/eclipse/xpanse/modules/models/view/RegisteredServiceVo.java
```diff
@@ -14,16 +14,19 @@
...
 import org.eclipse.xpanse.modules.models.resource.Ocl;
+import org.springframework.hateoas.RepresentationModel;
 
 /**
  * Define view object for UI Client query registered service.
  */
+@EqualsAndHashCode(callSuper = true)
 @Data
-public class RegisteredServiceVo {
+public class RegisteredServiceVo extends RepresentationModel<RegisteredServiceVo> {
```
swaroopar, line 29: Instead of having such a generic list of links, please see if we can have a  specific field called "openApiUrl" and return the URL there?

modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/register/impl/RegisterServiceImpl.java
```diff
@@ -283,6 +290,8 @@ private RegisteredServiceVo convertToRegisteredServiceVo(RegisterServiceEntity s
         }
         RegisteredServiceVo registeredServiceVo = new RegisteredServiceVo();
         BeanUtils.copyProperties(serviceEntity, registeredServiceVo);
+        registeredServiceVo.add(
+                Link.of(openApiUtil.getOpenApiUrl(serviceEntity.getId().toString()), "openAPI"));
```
swaroopar, line 294: something is wrong with the current configuration.. The field is shown as "_links" but when we do an actual GET.. only "links" is returned.

image

swaroopar (changes requested):
looks good. But would like to have it as a separate href field instead of a generic list.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/view/RegisteredServiceVo.java
baixinsui: Ok. I will fill the url into field 'openApiUrl'.

modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/register/impl/RegisterServiceImpl.java
baixinsui: This error no longer exists  using field field 'openApiUrl' to return url.

swaroopar (changes requested):
@baixinsui as discussed, 
1. HATEOS is important here. So we use what is supported by HATEOS the best. 
2. Data model and response matches - "links" or "_links" but it should be same everywhere.
3. Let the framework decide the URL. and enable "Forwarded Header" handling.

baixinsui:
> @baixinsui as discussed,
> 
> 1. HATEOS is important here. So we use what is supported by HATEOS the best.
> 2. Data model and response matches - "links" or "_links" but it should be same everywhere.
> 3. Let the framework decide the URL. and enable "Forwarded Header" handling.

I fixed all above.

modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/utils/OpenApiUtil.java
```diff
@@ -107,23 +115,59 @@ public String getOpenApi(RegisterServiceEntity registerServiceEntity) {
      *
      * @return serviceUrl
      */
-    private String getServiceUrl() {
-        String ip = "localhost";
+    public String getServiceUrl() {
```
swaroopar, line 118: @baixinsui I'm a little confused about this method. 
If I understood the framework right, it automatically generates the correct URLs and we need not do anything extra. Are the URLs not generated correctly out of the box?

baixinsui: Is it ok now?

baixinsui: I had used WebMvcLinkBuilder.linkTo(
                    WebMvcLinkBuilder.methodOn(OrchestratorApi.class).openApi(id)).withRel("openApi"); happens
org.springframework.aop.framework.AopConfigException: Could not generate CGLIB subclass of class java.lang.String: Common causes of this problem include using a final class or a non-visible class.                                                                               
I had tried to fix it but not succeed.  And this mothed needs the openApi mothed returns String and only used in OrchestratorApi.class. So that, I use Link.of() and get the baseUrl

swaroopar (changes requested):
comment added.

modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/utils/OpenApiUtil.java
baixinsui: I use WebMvcLinkBuilder.linkTo(WebMvcLinkBuilder.methodOn(OrchestratorApi.class).openApi(serviceEntity.getId().toString())).withRel("openApi") to fill openApiUrl. This method is used to fill the serviceUrl in html file.

swaroopar (approved):
Looks great.

PR #267, opened 2023-04-11
Save TfState file and destroying build workspace

### LINKED ISSUE #268: Restructure the deployment environment when destory
- [ ] Save the informations of the deployment.
- [ ] Restructure the deployment environment when destroy.

WangLiNaruto:
fix #268

### DIFF
--- a/modules/database/src/main/java/org/eclipse/xpanse/modules/database/service/DeployServiceEntity.java
+++ b/modules/database/src/main/java/org/eclipse/xpanse/modules/database/service/DeployServiceEntity.java
@@ -104,6 +104,16 @@ public class DeployServiceEntity extends CreateModifiedTime {
             joinColumns = @JoinColumn(name = "DEPLOY_SERVICE_ID", nullable = false))
     @MapKeyColumn(name = "P_KEY")
     @Column(name = "P_VALUE")
-    private Map<String, String> deployedServiceProperties;
+    private Map<String, String> properties;
+
+    /**
+     * The properties of the deployed service.
+     */
+    @ElementCollection
+    @CollectionTable(name = "DEPLOY_RESULT_PROPERTY",
+            joinColumns = @JoinColumn(name = "DEPLOY_SERVICE_ID", nullable = false))
+    @MapKeyColumn(name = "P_KEY")
+    @Column(name = "P_VALUE", length = Integer.MAX_VALUE)
+    private Map<String, String> privateProperties;
 
 }
--- a/modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/Deployment.java
+++ b/modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/Deployment.java
@@ -17,7 +17,7 @@ public interface Deployment {
 
     DeployResult deploy(DeployTask task);
 
-    DeployResult destroy(DeployTask task);
+    DeployResult destroy(DeployTask task, String tfState);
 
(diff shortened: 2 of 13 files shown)


modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/Deployment.java
```diff
@@ -17,7 +17,7 @@ public interface Deployment {
 
     DeployResult deploy(DeployTask task);
 
-    DeployResult destroy(DeployTask task);
+    DeployResult destroy(DeployTask task, String tfState);
```
iskey, line 20: Shoud not use tfState here, for tfState is just for terraformDeployment. should not be exposed in the skeleton.

modules/database/src/main/java/org/eclipse/xpanse/modules/database/service/DeployServiceEntity.java
```diff
@@ -106,4 +106,14 @@ public class DeployServiceEntity extends CreateModifiedTime {
...
 
+    /**
+     * The properties of the deployed service.
+     */
+    @ElementCollection
+    @CollectionTable(name = "DEPLOY_RESULT_PROPERTY",
+            joinColumns = @JoinColumn(name = "DEPLOY_SERVICE_ID", nullable = false))
+    @MapKeyColumn(name = "P_KEY")
+    @Column(name = "P_VALUE", length = Integer.MAX_VALUE)
+    private Map<String, String> deployedResultProperties;
```
iskey, line 117: how about renaming this to deploymentProperties?

modules/models/src/main/java/org/eclipse/xpanse/modules/models/service/DeployResult.java
```diff
@@ -45,7 +45,11 @@ public class DeployResult {
      * The result property of the service task.
      */
     @NotNull
-    @Schema(description = "The result property of the service task.")
-    private Map<String, String> property = new HashMap<>();
+    @Schema(description = "The properties of the deployed service.")
+    private Map<String, String> deployedServiceProperties = new HashMap<>();
```
iskey, line 49: we can keep the field property as it is and renaming the `deployResultProperties` to `rawProperties`, may be properties is better?

plugins/flexibleengine/src/main/java/org/eclipse/xpanse/orchestrator/plugin/flexibleengine/FlexibleEngineTerraformResourceHandler.java
```diff
@@ -48,9 +48,8 @@ public void handler(DeployResult deployResult) {
         List<DeployResource> deployResourceList = new ArrayList<>();
         TfState tfState;
         try {
-            var stateFile = deployResult.getProperty().get("stateFile");
+            var stateFile = deployResult.getDeployedResultProperties().get("stateFile");
```
iskey, line 51: I suggest that we move these codes to the terraform deployment, not in the plugins.

iskey (changes requested):
comments added.

modules/database/src/main/java/org/eclipse/xpanse/modules/database/service/DeployServiceEntity.java
iskey: how about endpointProperties and privateProperties?

iskey (approved):
looks good.

PR #278, opened 2023-04-12
load flavor properties as regular input variables

### LINKED ISSUE #276: Use flavor properties as regular input variables in terraform
Currently the flavor variables are inserted as environment variables. To enhance simplicity, the following changes are needed.

1) remove the flavor name from the env variables.
2) delete the prefix of the flavor priorities.
3) turn the flavor priorities to input variables.

swaroopar:
fixes #276

### DIFF
--- a/modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/utils/DeployEnvironments.java
+++ b/modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/utils/DeployEnvironments.java
@@ -58,12 +58,10 @@ public class DeployEnvironments {
      */
     public static Map<String, String> getFlavorVariables(DeployTask task) {
         Map<String, String> variables = new HashMap<>();
-
-        variables.put("flavor", task.getCreateRequest().getFlavor());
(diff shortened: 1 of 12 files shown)


modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/utils/DeployEnvironments.java
```diff
@@ -99,7 +97,7 @@ public static Map<String, String> getVariables(DeployTask task) {
                 variables.put(variable.getName(), variable.getValue());
             }
         }
-
+        variables.putAll(getFlavorVariables(task));
```
iskey, line 100: I suggest we leave the kinds of variables to be assembled in the terraform deployment. 
I am wondering if all deployment implementations will do it the same way.

swaroopar: @iskey ok understood. I will move this to Terraform specific classes.

iskey (approved):
looks good.

PR #286, opened 2023-04-15
Return enumerated error code and separate field for actual error message

swaroopar:
these changes allow clients of the API to have an enumerated list of result types and also the details of the error in a separate field.


iskey (approved):
looks good.

PR #296, opened 2023-04-17
validate terraform script

### LINKED ISSUE #261: Validate Terraform syntax in register request body
Check if there is a possibility to validate the terraform syntax in the register request. 
Currently the failure is known only when the service is ordered for the first time.

swaroopar:
fixes #261 

image


iskey (approved):
looks good.

PR #298, opened 2023-04-17
use enum directly in request params

swaroopar:
With this change, category in the requests parameters are no more a string. 

image


iskey (approved):
seems good.
but we should wait for the GUI updated at the same time.

iskey (changes requested):
please wait for the GUI part.

iskey (approved):
related to /eclipse-xpanse/xpanse-ui#49

PR #301, opened 2023-04-18
validate response beans with schema definition

### LINKED ISSUE #300: Validate API responses before returning
Currently we are not validating the responses returned and we are breaking our own API specs in the returned responses sometimes. 
Validate the response beans and throw errors if data does not match to our spec.
This is will act as a safety net during our development to avoid breaking changes.

swaroopar:
Fixes #300 

With this change, the API will not return anything that does not match to our own Schema. If response is bad we return the bean validation errors back.


iskey (approved):
looks good.

PR #302, opened 2023-04-18
set service version in service API doc

swaroopar:


modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/utils/OpenApiUtil.java
```diff
@@ -253,8 +253,8 @@ private String getApiDocsJson(RegisterServiceEntity registerService) {
         if (Objects.isNull(registerService)) {
             return StringUtils.EMPTY;
         }
-        // version of OCL.
-        String oclVersion = registerService.getOcl().getVersion();
+        // version of the service.
+        String oclVersion = registerService.getVersion();
```
iskey, line 257: shall we rename the `oclVersion` to `serviceVersion` at the same time?

iskey (approved):
looks good. one minor comment was added.

modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/utils/OpenApiUtil.java
swaroopar: yes makes sense. I oversaw that. Fixed now.

PR #305, opened 2023-04-19
remove wrong validator on enum field

swaroopar:
This is a bug. When we add Notblank on enums, the response validation fails. Notblank is only for strings.


iskey (approved):
looks good.

PR #310, opened 2023-04-19
set the value of the regions.

### LINKED ISSUE #309: Set the value of the regions.
Set the value of the regions.

Alice1319:
fix #309 

image


swaroopar (approved):
looks good.

PR #312, opened 2023-04-20
Add a sensitive flag for hidden the input on the GUI

wanglin100:


iskey (approved):
looks good, we should merge the /eclipse-xpanse/xpanse-ui#59 and this PR in sync.

PR #314, opened 2023-04-20
return a proper object to avoid client breaking

swaroopar:
Current return type is not clear for clients. We must return a proper type.


iskey (approved):
looks good.

PR #315, opened 2023-04-21
Sensitive field scope defaults

wanglin100:


modules/models/src/main/java/org/eclipse/xpanse/modules/models/resource/DeployVariable.java
```diff
@@ -51,7 +52,6 @@ public class DeployVariable {
     @Schema(description = "Validator of the variable")
     private String validator;
 
-    @NotNull
     @Schema(description = "Sensitive scope of variable storage")
-    private SensitiveScope scope;
+    private SensitiveScope scope = SensitiveScope.NONE;
```
swaroopar, line 56: please also call the field name `sensitiveScope` instead of just `scope`.

Also, please see if this default is correctly displayed in the swagger UI / open API doc, if not, please add it to the description.

swaroopar (changes requested):
comment added.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/resource/DeployVariable.java
wanglin100: ok

wanglin100: Default values ​​are displayed correctly

swaroopar (approved):
looks good.

PR #319, opened 2023-04-23
add option to run Terraform cli with debug

### LINKED ISSUE #318: allow option to enable debug for Terraform CLI
1. Add option to enable debug for Terraform cli commands.
2. Split stderr and stdout capturing from the OS command because Terraform writes all debug logs only to stderr.

swaroopar:
Fixes #318


iskey (approved):
looks great.

PR #321, opened 2023-04-24
Add the skeleton of the monitor and credentialCenter

### LINKED ISSUE #317: Design for the Monitor of xpanse

iskey:
fix #317 
related to discussion #320


modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/Credential.java
```diff
@@ -0,0 +1,30 @@
...
+
+    /**
+     * The name of the credential, this field is provided by the plugins.
+     */
+    String name;
+
+    /**
+     * The userId of the credential.
+     */
+    String userId;
```
swaroopar, line 24: should this be the description?

modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/CredentialCenter.java
```diff
@@ -0,0 +1,56 @@
...
+
+package org.eclipse.xpanse.modules.credential;
+
+import java.util.List;
+import org.eclipse.xpanse.modules.models.enums.Csp;
+
+/**
+ * The credential center.
+ */
+public class CredentialCenter {
```
swaroopar, line 15: we can make it a spring bean to make this class a singleton.

modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/CredentialCenter.java
```diff
@@ -0,0 +1,56 @@
...
+
+    /**
+     * Get the credential abilities of Csp.
+     *
+     * @param csp The csp to get.
+     */
+    List<Credential> getCredentialAbility(Csp csp) {
```
swaroopar, line 22: should the return type be List<CredentialAbility>?

modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/Credential.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+package org.eclipse.xpanse.modules.credential;
+
+import org.eclipse.xpanse.modules.credential.enums.CredentialType;
+
+/**
+ * The credential object.
+ */
+public class Credential {
```
swaroopar, line 14: Would it make sense to name it as 'CredentialVarible'?

modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/Credential.java
```diff
@@ -0,0 +1,30 @@
...
+ * SPDX-FileCopyrightText: Huawei Inc.
+ *
+ */
+
+package org.eclipse.xpanse.modules.credential;
+
+import org.eclipse.xpanse.modules.credential.enums.CredentialType;
+
+/**
+ * The credential object.
```
swaroopar, line 12: Object which defines the credential variables required by a cloud provider.

modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/CredentialAbility.java
```diff
@@ -0,0 +1,26 @@
...
+
+package org.eclipse.xpanse.modules.credential;
+
+import java.util.List;
+import org.eclipse.xpanse.modules.credential.enums.CredentialType;
+
+/**
+ * Credential Ability of the special plugin for the CSP.
+ */
+public interface CredentialAbility {
```
swaroopar, line 15: How about calling the class just 'Authentication' ?

modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/CredentialAbility.java
```diff
@@ -0,0 +1,26 @@
...
+ *
+ */
+
+package org.eclipse.xpanse.modules.credential;
+
+import java.util.List;
+import org.eclipse.xpanse.modules.credential.enums.CredentialType;
+
+/**
+ * Credential Ability of the special plugin for the CSP.
```
swaroopar, line 13: Interface to be implemented by CSP plugins to define its authentication mechanisms.

modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/CredentialCenter.java
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    /**
+     * Create credential for the @Csp with @userName.
+     *
+     * @param csp        The csp  to create for.
+     * @param userName   The userName to create for.
+     * @param credential The credential to create.
+     */
+    boolean createCredential(Csp csp, String userName, Credential credential) {
```
swaroopar, line 33: Should the plugin call this method multiple times to store different Credential variables?

modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/InputVariable.java
```diff
@@ -0,0 +1,29 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * SPDX-FileCopyrightText: Huawei Inc.
+ *
+ */
+
+package org.eclipse.xpanse.modules.credential;
+
+/**
+ * The class object for the CredentialInput.
```
swaroopar, line 10: doc to be corrected.

modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/enums/CredentialType.java
```diff
@@ -0,0 +1,46 @@
...
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonValue;
+import org.apache.commons.lang3.StringUtils;
+
+/**
+ * The type of the credential.
+ */
+public enum CredentialType {
+    INPUT("input"),
```
swaroopar, line 17: would it make sense to have HTTP_AUTHENTICATION, APKI_KEY and OAUTH2 as types?

Just to avoid the generic 'INPUT' which is abstract.

modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/CredentialInput.java
```diff
@@ -0,0 +1,20 @@
...
+ * SPDX-FileCopyrightText: Huawei Inc.
+ *
+ */
+
+package org.eclipse.xpanse.modules.credential;
+
+import java.util.List;
+
+/**
+ * The credentialInput is kind of credential needs get values from the input.
```
swaroopar, line 12: Credential data that can be updated from end user.

modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/CredentialCenter.java
```diff
@@ -0,0 +1,56 @@
...
+        return false;
+    }
+
+    /**
+     * Get credential for the @Csp with @userName.
+     *
+     * @param csp      The csp  to get.
+     * @param userName The userName to get.
+     */
+    Credential getCredential(Csp csp, String userName) {
```
swaroopar, line 53: If i understood it right, a credential can be more than one 'Credential' object? for example one for usernme, one for password.

Can the method be List<Credential> getCredentialVariables(Csp csp, CredentialType credentialType) ?

modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/InputVariable.java
```diff
@@ -0,0 +1,29 @@
...
+
+    /**
+     * The name of the InputVariable.
+     */
+    String name;
+
+    /**
+     * The description of the InputVariable.
+     */
+    String description;
```
swaroopar, line 22: both  here and in Credential class, we need another name instead of description. 
Probably even an Enum is better here.

modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/Credential.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+public class Credential {
+
+    /**
+     * The name of the credential, this field is provided by the plugins.
+     */
+    String name;
```
swaroopar, line 19: can we make this class to pack all details together?
1. name -> may be we can even ignore this name field too? 
2. type
3. List of variable names.

swaroopar (changes requested):
@iskey thank you for thinking through this topic. Not an easy one.  Have reviewed based on my understanding of the topic after our discussions.

modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/Credential.java
iskey: No, the `userId` is to identify the users if needed. I will add the `description` field for the missing.

modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/CredentialCenter.java
iskey: No, the return type is not a problem, this method designed to get the detailed Credential abilities (e.g., with variables for `CredentialInput`)

modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/Credential.java
iskey: I think it is not suitable to name it `CredentialVariable`, 
because we are not sure about all kinds of credentials are Variables. variables should only be one of kind of them.

modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/CredentialCenter.java
iskey: I don't think so, we only need one kind of credential available for per-csp (maybe per-csp-user).

modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/CredentialInput.java
```diff
@@ -0,0 +1,20 @@
...
+ */
+public class CredentialInput extends Credential {
```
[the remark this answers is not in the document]
iskey: I think it is better to rename to `CredentialVariables` to make it more generic (from its kind, instead of from the business view).

modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/CredentialCenter.java
iskey: It makes sense to add CredentialType as parameters, but we must keep the userName here, for we are building a mult-user project.
for the return type, I think one available Credential is enough to drive the plugin to work,(also no need to create more than one available credential.)

modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/InputVariable.java
iskey: I feel `description` is good here. It is only some helpful messages for the document and the end-user, letting them know what is it defined for and how to use it.
`Enum` not make sense, or you misunderstood my ideas.

modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/Credential.java
iskey: I meant to make the Credential class to cover all kinds of credentials. Variables should be only one kind of them.
so we can use the Credential at the interface, and safely extend more kinds of Credentials.

swaroopar (approved):
looks good.

PR #326, opened 2023-04-27
store user details for deployment

### LINKED ISSUE #287: Store the user who ordered the service in DB
When a service is ordered 
1. Make user a mandatory field.
2. Store it in DB
3. In regular GET deployed services -> Make user the mandatory field in the request query. 

Additionally changes to be added to UI to use the logged in user. We can have a separate issue for this.

Note -> since we currently we do not have authentication and authorization concept, we can just assume a string field for username.

swaroopar:
Fixes #287


iskey (approved):
looks good.

PR #327, opened 2023-04-28
Define the metric name and monitor resource type

WangLiNaruto:
fixed

### DIFF
--- a/modules/monitor/src/main/java/org/eclipse/xpanse/modules/monitor/enums/MetricNameEnum.java
+++ b/modules/monitor/src/main/java/org/eclipse/xpanse/modules/monitor/enums/MetricNameEnum.java
@@ -1,0 +1,48 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * SPDX-FileCopyrightText: Huawei Inc.
+ *
+ */
+
+package org.eclipse.xpanse.modules.monitor.enums;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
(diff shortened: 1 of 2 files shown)


modules/monitor/src/main/java/org/eclipse/xpanse/modules/monitor/enums/MetricNameEnum.java
```diff
@@ -0,0 +1,48 @@
...
+    /**
+     * For MetricName serialize.
+     */
+    @JsonValue
+    public String toValue() {
+        return this.name;
+    }
+
+    /**
+     * For MetricName serialize.
```
swaroopar, line 37: comment should be `For MetricName deserialize.`

modules/monitor/src/main/java/org/eclipse/xpanse/modules/monitor/enums/MonitorResourceType.java
```diff
@@ -0,0 +1,48 @@
...
+    /**
+     * For MonitorResourceType serialize.
+     */
+    @JsonValue
+    public String toValue() {
+        return this.value;
+    }
+
+    /**
+     * For MonitorResourceType serialize.
```
swaroopar, line 37: comment should be `For MonitorResourceType deserialize.`

swaroopar (changes requested):
comment to be updatd.

swaroopar (approved):
looks good.

PR #330, opened 2023-04-28
add unit test for the register.

### LINKED ISSUE #189: Add test suit for the register/deploy services.
- [ ] Add test suits for APIs.
- [ ] Add tests for deployment.
- [ ] Add unit tests for the register.
- [ ] Add unit tests for the deployment.

jinyangyang222:
Fixes #189

### DIFF
--- a/modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/deployers/terraform/TerraformExecutor.java
+++ b/modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/deployers/terraform/TerraformExecutor.java
@@ -91,11 +91,11 @@ public class TerraformExecutor {
     private boolean executeWithVariables(StringBuilder command) {
         for (Map.Entry<String, String> entry : this.variables.entrySet()) {
             if (Objects.nonNull(entry.getKey()) && Objects.nonNull(entry.getValue())) {
-                command.append("-var=\"")
+                command.append("-var=")
                         .append(entry.getKey())
                         .append("=")
                         .append(entry.getValue())
-                        .append("\" ");
+                        .append(" ");
             }
         }
         SystemCmdResult systemCmdResult = execute(command.toString());


modules/orchestrator/src/test/java/org/eclipse/xpanse/orchestrator/XpanseTestApplication.java
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * Main entry class to xpanse runtime. This class can be directly executed to start the server.
+ */
+@SpringBootApplication
+@EnableAsync(proxyTargetClass = true)
+@EnableJpaRepositories(basePackages = "org.eclipse.xpanse")
+@EntityScan(basePackages = "org.eclipse.xpanse")
+@ComponentScan(basePackages = "org.eclipse.xpanse")
+@EnableJpaAuditing
+public class XpanseTestApplication {
```
swaroopar, line 26: is it not possible to simply start the actual main class here with test profile?

jinyangyang222: The actual main class cannot be started here, possible cyclic dependency.

swaroopar: oh ok. yes you are right. then can we move this test directly under runtime package?
If not, i would avoid using this and instead only use other ways to instantiate classes which are needed.

jinyangyang222: yes ok,move to the runtime package is good, i will do this.

swaroopar (approved):
looks good.

jinyangyang222:
> @jinyangyang222 is this PR now ready for merge? I see the status has been changed to Draft.

Don't merge now,In the runtime package, I think the testing of API modules is more reasonable,how do you like it?

PR #350, opened 2023-05-11
Redefining packages and interfaces of service,credential,storage,monitor

baixinsui:
@iskey Please review it.


modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/storage/DeployResourceStorage.java
```diff
@@ -0,0 +1,79 @@
...
+ * Interface for persist of DeployResource.
+ */
+public interface DeployResourceStorage {
+
+    void deleteByDeployServiceId(UUID id);
+
+    /**
+     * Interface to be implemented by register service database.
+     */
+    interface RegisterServiceStorage {
```
swaroopar, line 24: Do we really need such a nested interface?

swaroopar (changes requested):
package separation looks good. 
Only question about nested interfaces

swaroopar (approved):
@baixinsui looks fine for me. If Yichao has no further comments, we can merge it.

iskey (approved):
looks good

PR #363, opened 2023-05-18
Implement CredentialCenter methods of credential management APIs

baixinsui:
@iskey @swaroopar Please review this PR.

### DIFF
--- a/modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/CredentialCacheManager.java
+++ b/modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/CredentialCacheManager.java
@@ -1,0 +1,137 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * SPDX-FileCopyrightText: Huawei Inc.
+ *
+ */
+
+package org.eclipse.xpanse.modules.credential;
+
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Objects;
+import java.util.Set;
+import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
+import lombok.extern.slf4j.Slf4j;
+import org.eclipse.xpanse.modules.credential.enums.CredentialType;
+import org.slf4j.MDC;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.scheduling.annotation.Async;
+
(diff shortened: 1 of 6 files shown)


modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/credential/CredentialCenter.java
```diff
@@ -201,11 +250,89 @@ public CredentialDefinition getCredential(Csp csp, String userName) {
...
+
+    private CredentialDefinition fillCredentialWithDbEntity(CreateCredential createCredential,
+                                                            DeployServiceEntity deployService) {
+        createCredential.setCsp(deployService.getCsp());
+        createCredential.setUserName(deployService.getUserName());
+        return transCredential(createCredential);
+
+    }
+
+    private CredentialDefinition transCredential(CreateCredential createCredential) {
```
swaroopar, line 324: createCredentialDefinitionObject is a better name here

modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/CredentialCacheManager.java
```diff
@@ -0,0 +1,136 @@
...
+     * @return true or false.
+     */
+    public boolean isEmpty() {
+        return CACHES.isEmpty();
+    }
+
+    /**
+     * Put cache.
+     *
+     * @param key   key
```
swaroopar, line 46: in this entire class.. we must replace "key" with a more concrete value.
One idea would be.. to have a new object which contains - Csp and username. and use that in all method definitions. 
and move the getCacheKey method inside this class.

swaroopar (changes requested):
comments added

modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/credential/CredentialCenter.java
baixinsui: Ok.

modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/CredentialCacheManager.java
baixinsui: Done.

swaroopar (approved):
looks good.

iskey (approved):
looks good.

PR #371, opened 2023-05-24
add unit tests for the monitor api and credential manage api

wanglin100:


runtime/src/test/java/org/eclipse/xpanse/runtime/CredentialManageApiTest.java
```diff
@@ -0,0 +1,398 @@
...
+                        List.of(new CredentialVariable("name", "description"))));
+        when(mockCredentialCenter.getCredentialDefinitionsByServiceId("id",
+                CredentialType.VARIABLES)).thenReturn(credentialDefinitions);
+
+        // Run the test
+        final List<CredentialDefinition> result =
+                credentialManageApiUnderTest.getCredentialDefinitionsByServiceId("id",
+                        CredentialType.VARIABLES);
+
+        // Verify the results
```
swaroopar, line 296: assertions missing?

swaroopar (changes requested):
comment added

runtime/src/test/java/org/eclipse/xpanse/runtime/CredentialManageApiTest.java
wanglin100: I have modified

swaroopar (approved):
looks good.

PR #372, opened 2023-05-25
Implement the monitor plugin for the FlexibleEngine

### LINKED ISSUE #353: Implement the monitor plugin for the Flexibleengine(with cpu/mem for VM).

baixinsui:
fix #353


plugins/flexibleengine/src/main/java/org/eclipse/xpanse/orchestrator/plugin/flexibleengine/monitor/utils/FlexibleEngineMonitorConverter.java
```diff
@@ -0,0 +1,204 @@
...
+        }
+        return projectId;
+    }
+
+    private static StringBuilder getQueryMetricBasicUrl(String region, String projectId) {
+        return new StringBuilder(FlexibleEngineMonitorConstants.PROTOCOL_HTTPS)
+                .append(FlexibleEngineMonitorConstants.CES_ENDPOINT_PREFIX)
+                .append(region)
+                .append(FlexibleEngineMonitorConstants.ENDPOINT_SUFFIX).append("/")
+                .append("V1.0").append("/")
```
swaroopar, line 159: the version can also be moved to constants file.

plugins/flexibleengine/src/main/java/org/eclipse/xpanse/orchestrator/plugin/flexibleengine/monitor/utils/FlexibleEngineMonitorConverter.java
```diff
@@ -0,0 +1,204 @@
...
+import org.eclipse.xpanse.orchestrator.plugin.flexibleengine.monitor.constant.FlexibleEngineMonitorConstants;
+import org.eclipse.xpanse.orchestrator.plugin.flexibleengine.monitor.models.FlexibleEngineMonitorMetrics;
+import org.eclipse.xpanse.orchestrator.plugin.flexibleengine.monitor.models.FlexibleEngineNameSpaceKind;
+import org.springframework.util.CollectionUtils;
+
+/**
+ * Resource conversion.
+ */
+@Slf4j
+public class FlexibleEngineMonitorConverter {
```
swaroopar, line 36: plese see if we can make this a spring component. It will then be easy if we want to mock them later for our tests.

plugins/flexibleengine/src/main/java/org/eclipse/xpanse/orchestrator/plugin/flexibleengine/monitor/utils/FlexibleEngineMonitorConverter.java
```diff
@@ -0,0 +1,204 @@
...
+     *
+     * @param region The region of resource.
+     * @return Returns query url.
+     */
+    public static StringBuilder buildProjectQueryUrl(String region) {
+        return new StringBuilder(FlexibleEngineMonitorConstants.PROTOCOL_HTTPS)
+                .append(FlexibleEngineMonitorConstants.IAM_ENDPOINT_PREFIX)
+                .append(region)
+                .append(FlexibleEngineMonitorConstants.ENDPOINT_SUFFIX).append("/")
+                .append("v3").append("/")
```
swaroopar, line 176: version here can be moved to constants.

plugins/flexibleengine/src/main/java/org/eclipse/xpanse/orchestrator/plugin/flexibleengine/FlexibleEngineOrchestratorPlugin.java
```diff
@@ -67,7 +72,71 @@ public List<AbstractCredentialInfo> getCredentialDefinitions() {
...
+        if (StringUtils.isBlank(region)) {
+            log.error("GetMetrics param region is blank. resource id:{}",
+                    deployResource.getResourceId());
+            return metrics;
+        }
+        String queryProjectIdUrl =
+                FlexibleEngineMonitorConverter.buildProjectQueryUrl(region).toString();
+        FlexibleEngineMonitorClient monitorClient =
+                getFlexibleEngineMonitorClient((CredentialDefinition) credential);
+        String projectId = queryProjectId(monitorClient, queryProjectIdUrl);
```
swaroopar, line 86: Even getting the projectID does not work via Huaweicloud SDKs? Does the APIs do not match or do we have some other issue?

swaroopar (changes requested):
looks good except the fact that we are generating requests and processing responses in a very native way. 
I was expecting that Flexible engine will work with its SDK or worst case with HuaweiCloud SDK.

plugins/flexibleengine/src/main/java/org/eclipse/xpanse/orchestrator/plugin/flexibleengine/monitor/utils/FlexibleEngineMonitorConverter.java
baixinsui: OK.

plugins/flexibleengine/src/main/java/org/eclipse/xpanse/orchestrator/plugin/flexibleengine/monitor/utils/FlexibleEngineMonitorConverter.java
```diff
@@ -0,0 +1,204 @@
...
+    /**
+     * Get id of project from response body.
+     *
+     * @param result result.
+     * @return Returns id of project.
+     */
+    public static String getProjectIdFromResponse(String result) {
+        String projectId = "";
+        try {
+            JsonNode resultJson = objectMapper.readTree(result);
```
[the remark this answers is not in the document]
baixinsui: Ok. I will define the models ourselves and then directly deserialize the complete object and try to use Spring Resttemplate to do the http request.

plugins/flexibleengine/src/main/java/org/eclipse/xpanse/orchestrator/plugin/flexibleengine/monitor/utils/FlexibleEngineMonitorConverter.java
baixinsui: Ok.

plugins/flexibleengine/src/main/java/org/eclipse/xpanse/orchestrator/plugin/flexibleengine/monitor/utils/FlexibleEngineMonitorConverter.java
baixinsui: Ok.

plugins/flexibleengine/src/main/java/org/eclipse/xpanse/orchestrator/plugin/flexibleengine/FlexibleEngineOrchestratorPlugin.java
baixinsui: There is no API matched in the  Flexible engine SDK and  in the Huaweicloud SDK the endponit of service is fixed. In addition to using API queries, we can also add "PROJECT_ID" like 'AK' in the credential provided by user. But I think calling API is better. I will try to use Spring Resttemplate to call this API . Is it ok?

baixinsui:
> looks good except the fact that we are generating requests and processing responses in a very native way.
> I was expecting that Flexible engine will work with its SDK or worst case with HuaweiCloud SDK.

The  Flexible Engine SDK is too simple to process the API of the CES. The HuaweiCloud SDK is more perfect but the Endpoint value is fixed.

swaroopar (approved):
looks good.

PR #374, opened 2023-05-25
CPU and memory monitoring for openstack

### LINKED ISSUE #291: Implement the monitor plugin for the OpenStack (with `cpu/mem` for VM).
- [ ] Implement the monitor plugin for the OpenStack with cpu/mem for VM (with cpu/mem for VM).

swaroopar:
Fixes #291


iskey (approved):
looks good.

plugins/openstack/src/main/java/org/eclipse/xpanse/orchestrator/plugin/openstack/monitor/keystone/CustomEndPointResolver.java
```diff
@@ -0,0 +1,250 @@
...
+                return p.access.getEndpoint();
+            }
+        }
+        return null;
+    }
+
+    private String resolveV3(URLResolverParams urlResolverParams) {
+        Token token = urlResolverParams.token;
+
+        //in v3 api, if user has no default project, and token is unscoped,
```
iskey, line 135:
```suggestion
        // in v3 api, if user has no default project, and token is unscoped,
```
or, can we do it with `/**/` here?

PR #385, opened 2023-05-31
Add metrics query APIs

### LINKED ISSUE #376: Add period metrics querying API
Add period metrics querying API

baixinsui:
fix #376

### DIFF
--- a/modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/monitor/Monitor.java
+++ b/modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/monitor/Monitor.java
@@ -11,67 +11,205 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
 import java.util.UUID;
-import org.eclipse.xpanse.modules.credential.AbstractCredentialInfo;
+import java.util.stream.Collectors;
+import lombok.extern.slf4j.Slf4j;
+import org.eclipse.xpanse.modules.credential.CredentialDefinition;
 import org.eclipse.xpanse.modules.database.resource.DeployResourceEntity;
+import org.eclipse.xpanse.modules.database.resource.DeployResourceStorage;
 import org.eclipse.xpanse.modules.database.service.DeployServiceEntity;
 import org.eclipse.xpanse.modules.database.service.DeployServiceStorage;
+import org.eclipse.xpanse.modules.database.utils.EntityTransUtils;
+import org.eclipse.xpanse.modules.models.enums.Csp;
 import org.eclipse.xpanse.modules.models.enums.DeployResourceKind;
 import org.eclipse.xpanse.modules.models.service.DeployResource;
 import org.eclipse.xpanse.modules.monitor.Metric;
+import org.eclipse.xpanse.modules.monitor.ResourceMetricRequest;
+import org.eclipse.xpanse.modules.monitor.ServiceMetricRequest;
 import org.eclipse.xpanse.modules.monitor.enums.MonitorResourceType;
 import org.eclipse.xpanse.orchestrator.OrchestratorPlugin;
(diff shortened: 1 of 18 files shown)


modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/monitor/Monitor.java
```diff
@@ -11,67 +11,205 @@
...
+        if (Objects.isNull(from)) {
+            from = System.currentTimeMillis() - 60 * 1000;
+        }
+        if (Objects.isNull(to)) {
+            to = System.currentTimeMillis();
+        }
+        return new ServiceMetricRequest(deployResources, credential, monitorType, from, to, period);
+    }
+
+    private void checkParamsValid(Long from, Long to) {
```
swaroopar, line 205: method name can be -> validateToAndFromValues.

modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/monitor/Monitor.java
```diff
@@ -11,67 +11,205 @@
...
+        if (Objects.nonNull(from) && Objects.nonNull(to)) {
```
swaroopar, line 206: i think it is also better to validate if the `to` and from `values` are also valid time stamps.

modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/monitor/Monitor.java
```diff
@@ -11,67 +11,205 @@
...
+        }
+        return new ResourceMetricRequest(deployResource, credential, monitorType, from, to, period);
+    }
+
+    private ServiceMetricRequest getServiceMetricRequest(List<DeployResource> deployResources,
+                                                         CredentialDefinition credential,
+                                                         MonitorResourceType monitorType,
+                                                         Long from, Long to, Integer period) {
+        if (Objects.isNull(from)) {
+            from = System.currentTimeMillis() - 60 * 1000;
```
swaroopar, line 197: we must document this default value in the API also.

swaroopar (changes requested):
comments added. Also there are merge conflicts. Please check that too.

swaroopar (changes requested):
comments added. Also there are merge conflicts. Please check that too.

modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/monitor/Monitor.java
baixinsui: Ok.

modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/monitor/Monitor.java
baixinsui: Any number is a valid unix timestamps. I think it is not necessary. Or we can define a minimum start time '1970-01-01 00:00:00.000' with number 0?

modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/monitor/Monitor.java
baixinsui: I will add it.

modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/monitor/Monitor.java
```diff
@@ -11,67 +11,205 @@
...
+        }
+        if (Objects.isNull(to)) {
+            to = System.currentTimeMillis();
```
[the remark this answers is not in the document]
baixinsui: I will add it.

PR #386, opened 2023-05-31
Add monitor query APIs

### LINKED ISSUE #376: Add period metrics querying API

baixinsui:
fix #376


modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/monitor/Monitor.java
```diff
@@ -11,67 +11,210 @@
...
+        if (Objects.isNull(from)) {
+            from = System.currentTimeMillis() - FIVE_MINUTES_MILLISECONDS;
+        }
+        if (Objects.isNull(to)) {
+            to = System.currentTimeMillis();
+        }
+        return new ServiceMetricRequest(deployResources, credential, monitorType, from, to, period);
+    }
+
+    private void validateToAndFromValues(Long from, Long to) {
```
swaroopar, line 206: we still a validation to check if the provided Long value is really a valid time stamp. can you add that please?

swaroopar (changes requested):
comments added.

modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/monitor/Monitor.java
baixinsui: Any Long value is a valid timestamp. How to check it? Greater the value 1672531200000 of the time '2023-01-01 00:00:00.000'? Less the Long value of nowtime?

swaroopar: As discussed offline, we can have min 0 and max as LONG.MAX_VALUE

baixinsui: I add the Min(0), the Max(Long.MAX_VALUE) is unnecessary. When the input value greater than Long.MAX_VALUE，the following exception will occurs firstly:  "Failed to convert value of type 'java.lang.String' to required type 'java.lang.Long'; For input string: \"3442265300000000000000000000000000\""

swaroopar (approved):
looks good.

PR #387, opened 2023-06-01
add monitor metrics(network for vm)

jinyangyang222:


modules/monitor/src/main/java/org/eclipse/xpanse/modules/monitor/enums/MonitorResourceType.java
```diff
@@ -15,7 +15,8 @@
  */
 public enum MonitorResourceType {
     CPU("cpu"),
-    MEM("mem");
+    MEM("mem"),
+    VM_BANDWIDTH("vm_bandwidth");
```
swaroopar, line 19: Not sure if VM_BANDWIDTH is completely correct. 
Would just NETWORK be fine here or may be NETWORK_USAGE?  (even in the previous values.. there is no VM_ appended)
Also does it mean inbound network usage or outbound usage or both?

jinyangyang222: @swaroopar  It means both inbound and outbound networks.

swaroopar: @jinyangyang222 I think it is better to keep inbound and outbound separate. I see in Huaweicloud and Openstack, we have two separate metrics for it.
would it make sense to name it to VM_NETWORK_INCOMING and VM_NETWORK_OUTGOING ?

jinyangyang222: Thank you, that would be even better.

swaroopar (approved):
looks good.

PR #392, opened 2023-06-02
update query metrics implement of FlexibleEngine

baixinsui:


plugins/flexibleengine/src/main/java/org/eclipse/xpanse/orchestrator/plugin/flexibleengine/monitor/utils/FlexibleEngineMonitorConverter.java
```diff
@@ -186,11 +170,10 @@ public StringBuilder buildProjectQueryUrl(String region) {
...
+            MetricInfoList metricInfo) {
         Map<String, String> params = new HashMap<>();
-        params.put("namespace", FlexibleEngineNameSpaceKind.ECS_SYS.toValue());
-        params.put("metric_name", getMetricName(resourceMetricRequest.getMonitorResourceType(),
-                flexibleEngineNameSpaceKind));
+        params.put("namespace", metricInfo.getNamespace());
+        params.put("metric_name", metricInfo.getMetricName());
         params.put("from", String.valueOf(resourceMetricRequest.getFrom()));
         params.put("to", String.valueOf(resourceMetricRequest.getTo()));
         params.put("period", String.valueOf(resourceMetricRequest.getPeriod()));
```
swaroopar, line 179: should we add null checks for adding query parameters? because not all parameters are manadatory.

plugins/flexibleengine/src/main/java/org/eclipse/xpanse/orchestrator/plugin/flexibleengine/monitor/utils/FlexibleEngineMonitorConverter.java
```diff
@@ -83,79 +81,65 @@ private StringBuilder getQueryListMetricsBasicUrl(String region, String projectI
...
         Metric metric = new Metric();
-        metric.setName(getMetricName(type, flexibleEngineNameSpaceKind));
+        metric.setName(metricInfo.getMetricName());
         Map<String, String> labels = new HashMap<>();
         labels.put("id", deployResource.getResourceId());
         labels.put("name", deployResource.getName());
         metric.setLabels(labels);
         metric.setType(MetricType.GAUGE);
+        if (metricInfo.getUnit().equals("%")) {
+            metric.setUnit(MetricUnit.PERCENTAGE);
```
swaroopar, line 124: metricUnit is a mandatory field. We must set this value for all cases.

swaroopar (changes requested):
comments added.

plugins/flexibleengine/src/main/java/org/eclipse/xpanse/orchestrator/plugin/flexibleengine/monitor/utils/FlexibleEngineMonitorConverter.java
baixinsui: Ok.

plugins/flexibleengine/src/main/java/org/eclipse/xpanse/orchestrator/plugin/flexibleengine/monitor/utils/FlexibleEngineMonitorConverter.java
baixinsui: Ok.

swaroopar (approved):
looks good.

PR #393, opened 2023-06-02
fix query filter parameters

swaroopar:


baixinsui (approved):
Looks good to me.

iskey (approved):
looks good.

PR #394, opened 2023-06-02
add filters to openstack metrics

### LINKED ISSUE #389: Implement metrics filter in openstack plugin

swaroopar:
fixes #389

### DIFF
--- a/plugins/openstack/src/main/java/org/eclipse/xpanse/orchestrator/plugin/openstack/monitor/keystone/KeystoneManager.java
+++ b/plugins/openstack/src/main/java/org/eclipse/xpanse/orchestrator/plugin/openstack/monitor/keystone/KeystoneManager.java
@@ -77,7 +77,6 @@ public class KeystoneManager {
                 .builderV3()
                 .withConfig(buildClientConfig(url))
                 .credentials(userName, password, Identifier.byName("default"))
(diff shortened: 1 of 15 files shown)


plugins/openstack/src/main/java/org/eclipse/xpanse/orchestrator/plugin/openstack/monitor/keystone/KeystoneManager.java
```diff
@@ -93,6 +93,7 @@ private static String getIpAddressFromUrl(String url) {
     private Config buildClientConfig(String url) {
         return Config.newConfig()
                 .withEndpointNATResolution(getIpAddressFromUrl(url))
-                .withEndpointURLResolver(new CustomEndPointResolver());
+                .withEndpointURLResolver(new CustomEndPointResolver())
+                .withProxy(new ProxyHost("http://proxyde.huawei.com", 8080));
```
iskey, line 97: seems we need some configurations for Monitor too.

swaroopar: @iskey good catch. I must remove this. This is needed only for me to work from office network.

iskey (approved):
looks good.

PR #403, opened 2023-06-05
fix service tag

swaroopar:


modules/api/src/main/java/org/eclipse/xpanse/api/CredentialManageApi.java
```diff
@@ -125,7 +125,7 @@ public List<CredentialDefinition> getCredentialDefinitionsByCsp(
      * @param type The type of credential.
      * @return Link of credential openApi url.
      */
-    @Tag(name = "Services Available",
+    @Tag(name = "Credentials Management",
             description = "APIs to query the available services.")
```
baixinsui, line 129: Please update the description.

modules/api/src/main/java/org/eclipse/xpanse/api/CredentialManageApi.java
```diff
@@ -327,7 +327,7 @@ public Boolean deleteCredentialByServiceId(
...
```
baixinsui, line 331: Please update the description.

modules/api/src/main/java/org/eclipse/xpanse/api/CredentialManageApi.java
swaroopar: fixed

modules/api/src/main/java/org/eclipse/xpanse/api/CredentialManageApi.java
swaroopar: fixed

swaroopar:
@baixinsui comments are fixed.

baixinsui (approved):
LGTM.

iskey (approved):
looks good.

PR #404, opened 2023-06-06
fix credential userName not used consistently

### LINKED ISSUE #401: CredentialsApi - UserName value not used consistently
In the current credentials concept, the userName field is not used consistently. We seem to mix up between xpanse usernames and usernames to be used for cloud providers

https://github.com/eclipse-xpanse/xpanse/blob/0a50e2356071a7d2a7233e8411d67d75343bcb6f/modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/CreateCredential.java#L32

https://github.com/eclipse-xpanse/xpanse/blob/0a50e2356071a7d2a7233e8411d67d75343bcb6f/modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/AbstractCredentialInfo.java#L38

https://github.com/eclipse-xpanse/xpanse/blob/0a50e2356071a7d2a7233e8411d67d75343bcb6f/modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/credential/CredentialCenter.java#L373  -> this is the xpanse user name which we get from the deploy service entity.

We need to define with more clarity which userName is this. 
If this is supposed to be the Xpanse user name.. then we can call it xpanseUserName.

baixinsui:
fix #401

### DIFF
--- a/modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/AbstractCredentialInfo.java
+++ b/modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/AbstractCredentialInfo.java
@@ -29,13 +29,13 @@ public abstract class AbstractCredentialInfo {
     Csp csp;
 
     /**
-     * The userId of the credential.
+     * The user who create the credential.
      */
     @Getter
     @Setter
     @NotNull
-    @Schema(description = "The userId of the credential.", defaultValue = "default")
-    String userName = "default";
+    @Schema(description = "The user who create the credential.")
+    String xpanseUser;
 
     /**
      * The name of the credential, this field is provided by the plugins.
--- a/modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/CreateCredential.java
+++ b/modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/CreateCredential.java
@@ -28,8 +28,8 @@ public class CreateCredential {
 
     @NotNull
     @NotBlank
-    @Schema(description = "User who create the credential.")
-    private String userName;
+    @Schema(description = "The user who create the credential.")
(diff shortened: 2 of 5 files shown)


swaroopar:
@baixinsui since in the credentials context there are multiple usernames involved, would it make sense to call it `XpanseUserName`?

iskey:
> @baixinsui since in the credentials context there are multiple usernames involved, would it make sense to call it `XpanseUserName`?

From my point of view, we should only have the `userName` for the `xpanse` but no one else. Am I right?

swaroopar:
> > @baixinsui since in the credentials context there are multiple usernames involved, would it make sense to call it `XpanseUserName`?
> 
> From my point of view, we should only have the `userName` for the `xpanse` but nothing else. Am I right?

I mean, in case we have userName credential variable also, then it might be a little confusing to know which userName is exactly meant here.

iskey:
> > > @baixinsui since in the credentials context there are multiple usernames involved, would it make sense to call it `XpanseUserName`?
> > 
> > 
> > From my point of view, we should only have the `userName` for the `xpanse` but nothing else. Am I right?
> 
> I mean, in case we have userName credential variable also, then it might be a little confusing to know which userName is exactly meant here. Basically i'm referring to confusion between Xpanse userName and CSP userName

I think you are right, but how about `xpanseUser`? or just `user` for xpanse?
@baixinsui what's your point?

swaroopar:
> @swaroopar @iskey It is a good idea. and I think using xpanseUser is better. Because the name may not be unique in the user system, but we need a unique representation to represent the user. I will change 'userName' to 'xpanseUser'. is it ok?

@baixinsui yes, fine for me.

swaroopar (approved):
looks good.

PR #405, opened 2023-06-06
delete unused manage credential methods with suffix ‘ByServiceId’

### LINKED ISSUE #400: CredentialsApi - Rename methods
In the monitorApi, 
addCredentialByServiceId, updateCredentialByServiceId, deleteCredentialByServiceId  services can be renamed as addCredentialByUsernameForCsp, updateCredentialByUsernameForCsp and  deleteCredentialByUsernameForCsp respectively.

For every Xpanse user, there will be just one user specific to Csp provided. So these 3 API methods are enough.

Use the userName in the CreateCredential model to build the cache key.
https://github.com/swaroopar/osc/blob/0a50e2356071a7d2a7233e8411d67d75343bcb6f/modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/CreateCredential.java#L32
 
Note - 
1. in the `addCredential` method, we are already using the "userName" in the `createCredential `object. So these methods are redundant. Can also be removed.
2. getCredentialDefinitionsByServiceId, getAvailableCredentialTypesByServiceId, getCredentialCapabilitiesByServiceId, getCredentialOpenApiByServiceId - Looks like this method also can be removed. Because credential definitions are only per Csp. We do not have anything service Id specific. Or we can rename them to `ByUser`. In my opinion these can be deleted.
we must keep all credential configurations only per Csp.

baixinsui:
fix #400


swaroopar (approved):
looks good.

PR #407, opened 2023-06-06
fix db error when destroying services

### LINKED ISSUE #406: Fix database exception  when destroying service
When destroying the service,  the following exception occur occasionally:
org.springframework.dao.InvalidDataAccessApiUsageException:
            No EntityManager with actual transactionavailable for current thread - cannot reliably process 'persist' call;
           nested exception isjavax.persistence.TransactionRequiredException: No EntityManager with actualtransaction available for current thread - cannot reliably process 'persist'call

baixinsui:
fix #406


swaroopar (approved):
looks good.

PR #411, opened 2023-06-07
update metrics return multiple points for HuaweiCloud

jinyangyang222:


plugins/huaweicloud/src/main/java/org/eclipse/xpanse/orchestrator/plugin/huaweicloud/monitor/utils/HuaweiCloudToXpanseDataModelConverter.java
```diff
@@ -164,10 +164,10 @@ public List<Metric> convertBatchListMetricDataResponseToMetric(
                                 datapointForBatchMetrics.sort(
```
baixinsui, line 164: The returned data is inherently ordered, and there is no need to sort the list again.

jinyangyang222: Okay, I have made the changes.

iskey (approved):
looks good.

PR #415, opened 2023-06-07
allow optional credential variables

### LINKED ISSUE #414: add optional credential variables to allow openstack optional configuration
Currently all credential variables are mandatory.
We must also support optional variables so that conditional variables can be passed for some CSPs.

swaroopar:
fixes #414

### DIFF
--- a/modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/CredentialVariable.java
+++ b/modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/CredentialVariable.java
@@ -17,36 +17,50 @@ import lombok.Data;
 @Data
 public class CredentialVariable {
 
-    /**
-     * The name of the CredentialVariable.
-     */
     @NotNull
     @Schema(description = "The name of the CredentialVariable,"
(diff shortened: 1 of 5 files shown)


modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/CredentialVariable.java
```diff
@@ -17,36 +17,52 @@
...
-    /**
-     * The value of the CredentialVariable.
-     */
+    @Schema(description = "If the variable is mandatory. If is optional then the credential "
+            + "completeness check will ignore this variable. It is upto the plugin to decide what "
+            + "needs to be done if this optional credential variable is present.",
+            defaultValue = "true")
+    private final boolean isMandatory;
+
+    private String newString;
```
baixinsui, line 36: What is this field for?

swaroopar: @baixinsui it was a wrong commit. Removed it now.

iskey (approved):
looks good.

PR #418, opened 2023-06-07
return error when exception occurs

### LINKED ISSUE #417: Service Destroy - Returns success always
the service destroy method currently returns success always. 
The exceptions are simply ignored.

swaroopar:
fixes #417

### DIFF
--- a/modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/utils/DeployEnvironments.java
+++ b/modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/utils/DeployEnvironments.java
@@ -12,21 +12,20 @@ import org.eclipse.xpanse.modules.deployment.deployers.terraform.DeployTask;
 import org.eclipse.xpanse.modules.models.enums.DeployVariableKind;
 import org.eclipse.xpanse.modules.models.resource.DeployVariable;
 import org.eclipse.xpanse.modules.models.resource.Flavor;
+import org.springframework.stereotype.Component;
 
 /**
  * Environment variables utils for deployment.
  */
+@Component
 public class DeployEnvironments {
-
-    private DeployEnvironments() {
-    }
 
     /**
      * Get environment variables for deployment.
      *
      * @param task the context of the task.
      */
-    public static Map<String, String> getEnv(DeployTask task) {
+    public Map<String, String> getEnv(DeployTask task) {
         Map<String, String> variables = new HashMap<>();
         Map<String, String> request = task.getCreateRequest().getServiceRequestProperties();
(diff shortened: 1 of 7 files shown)


modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/utils/DeployEnvironments.java
```diff
@@ -12,21 +12,20 @@
 import org.eclipse.xpanse.modules.models.enums.DeployVariableKind;
 import org.eclipse.xpanse.modules.models.resource.DeployVariable;
 import org.eclipse.xpanse.modules.models.resource.Flavor;
+import org.springframework.stereotype.Component;
 
 /**
  * Environment variables utils for deployment.
  */
+@Component
```
iskey, line 20: I think no need to make `DeployEnvironments` a `Component`, The methods here are sort of helpers for environments variables, no need to create an object.

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/utils/DeployEnvironments.java
```diff
@@ -12,21 +12,20 @@
...
-    private DeployEnvironments() {
-    }
-
     /**
      * Get environment variables for deployment.
      *
      * @param task the context of the task.
      */
-    public static Map<String, String> getEnv(DeployTask task) {
+    public Map<String, String> getEnv(DeployTask task) {
```
iskey, line 28: I feel like better to keep it as-is .

iskey (changes requested):
some minor comments added.

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/utils/DeployEnvironments.java
swaroopar: @iskey I made it a component so that it is possible to mock and test it without too much efforts. 
And it is also a singleton (by default) so no overhead we have here.

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/utils/DeployEnvironments.java
swaroopar: same here.

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/utils/DeployEnvironments.java
iskey: Ok, It makes sense, but I feel it is a bit strange always.

iskey (approved):
looks good.

PR #444, opened 2023-06-13
remove method getMetricTypes

### LINKED ISSUE #442: Remove method getMetricTypes
Method getMetricTypes returns the list of enum values. We do not need this. 
We have the enum defined in the interface for this purpose and all clients must simply use it.

jinyangyang222:
fixed #442


swaroopar (approved):
LGTM

PR #446, opened 2023-06-13
add isSensitive field to CredentialVariable

### LINKED ISSUE #420: Add new isSensitive field to CredentialVariable
- [ ] To the `CredentialVariable` data model, add a new mandatory field boolean field called `isSensitive `
- [ ] ~~In the GET methods which returns  `CredentialVariable`, the `value` field must be masked for in the JSON response when `isSensitive  `flag is true.~~ > To be done with user stories which implements GET methods.

swaroopar:
fixes #420


iskey (approved):
looks good.

PR #448, opened 2023-06-13
add onlyLastKnownMetric query parameter

### LINKED ISSUE #395: Monitor API - Add new parameter to return just the real time metric
1. Extend getMetricsByServiceId and getMetricsByResourceId with new optional boolean parameter - onlyLastKnownMetric

- if this value is true -> then all other parameters must be ignored and we return only the most recent metric available from backend. 
- if false or not provided -> then all other parameters must be considered.

swaroopar:
fixes #395


iskey (approved):
looks good.

PR #453, opened 2023-06-14
delete credentials only for specific type

### LINKED ISSUE #451: deleteCredential - add credential type parameter
Currently deleteCredential method deletes credentials of all types. 
we must restrict to delete only the specific type. 

- [ ] add new credential type query parameter
- [ ] parameter to be made mandatory

swaroopar:
fixes #451


iskey (approved):
looks good to me.

PR #454, opened 2023-06-14
remove method

### LINKED ISSUE #452: remove credentialsApi method getCredentialDefinitionsByCsp
getCredentialDefinitionsByCsp method currently returns the customer credentials stored in the credential cache and not the definition. 
This method is wrongly named and also we have user story #447 to implement the actual GET credentials method. 

So this existing method can be removed.

swaroopar:
fixes #452


swaroopar:
this method is already renamed in #455 . Closing this.

Issue #447, opened 2023-06-13
CredentialsApi - Add method to GET credentials stored


swaroopar:
In the credentials API, currently we do not have any method which returns the configured credentials for a CSP by an Xpanse user

- [x] New GET method - `getCredentials`
- [x] Mandatory query parameters - `csp`, `xpanseUser`
- [x] fields with `isSensitive` must have the value masked in the response. Return value as `*****` in such cases.
- [x] for credential variables whose value is null - we must read the value from OS environment variables and return it. So that the user knows which value would be used during the deployment or monitoring. This will also provide the end user an option to override with his custom value when the user finds out that this value is wrong.

jinyangyang222:
I'll do this.

jinyangyang222:
I'll do this.

jinyangyang222:
@swaroopar Hi swaroopar,The method  -getCredentials you mentioned above is the same as the existing method -getCredentialDefinitionsByCsp, but for the Wang Li front-end, it now requires a method that only passes the userName parameter.

swaroopar:
Hi @jinyangyang222, 
ideally `getCredentialDefinitionsByCsp `-> will only return the definitions by Csp, which is nothing but the type of credentials supported by CSP. This is not user specific and it will not have the actual credentials configured by the end user. 

the new  `getCredentials` will now returned the credentials configured by the end user for a CSP and for a specific type. 
The response model for this also depends a bit on #423

swaroopar:
@jinyangyang222  in getCredentials -> we do not need the definition. We need the values of the credential variables for that specific user.

swaroopar:
@jinyangyang222 we are using  GET method internally (here)[https://github.com/eclipse-xpanse/xpanse/blob/f0a113902fa70ef0edcf0045d43a95770d3a14a2/modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/credential/CredentialCenter.java#L179]
We must extend this to have credential type parameter and expose it  via REST API. 
Also add type from all current calling methods.

jinyangyang222:
okay,thank you @swaroopar,I will do this.

swaroopar:
Hi @iskey 
[...]

PR #455, opened 2023-06-14
Fix wrong type usage for credentials.

### LINKED ISSUE #449: Rename the `CredentialDefinition` to `CredentialVariables`
- [ ] As the `CredentialDefinition` is not a good name for one kind of `Credential`, We are going to rename it to `CredentialVariables`.

iskey:
fix #449 
related #447


swaroopar (approved):
looks good.

PR #459, opened 2023-06-14
allow plugin to decide the credential type required

### LINKED ISSUE #458: first available credentials used by orchestrator
Currently the first available metric from the cache is simply used by orchestrator for fetching metrics data. 
Instead the type of credential must be decided by CSP for each functionality.

swaroopar:
fixes #458


modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/monitor/Monitor.java
iskey, line 1: Seems lots of imports are unused then, after the refactoring.

iskey (approved):
looks good, only one minor comment was added.

modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/monitor/Monitor.java
swaroopar: fixing this. Was expecting the checkStyle to catch this error.
Will check that too.

PR #460, opened 2023-06-14
RestTemplate add Spring-retry and config log output

baixinsui:
#445

### DIFF
--- a/plugins/flexibleengine/src/main/java/org/eclipse/xpanse/orchestrator/plugin/flexibleengine/monitor/utils/RetryTemplateService.java
+++ b/plugins/flexibleengine/src/main/java/org/eclipse/xpanse/orchestrator/plugin/flexibleengine/monitor/utils/RetryTemplateService.java
@@ -1,0 +1,208 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * SPDX-FileCopyrightText: Huawei Inc.
+ *
+ */
+
+package org.eclipse.xpanse.orchestrator.plugin.flexibleengine.monitor.utils;
+
+import com.huaweicloud.sdk.ces.v1.model.BatchListMetricDataResponse;
+import com.huaweicloud.sdk.ces.v1.model.ListMetricsResponse;
+import com.huaweicloud.sdk.ces.v1.model.ShowMetricDataResponse;
+import com.huaweicloud.sdk.core.internal.model.KeystoneListProjectsResponse;
+import com.huaweicloud.sdk.core.internal.model.Project;
+import java.util.Arrays;
+import java.util.Map;
+import java.util.Objects;
+import java.util.stream.Collectors;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.http.Header;
+import org.apache.http.client.methods.HttpRequestBase;
+import org.springframework.context.annotation.Bean;
+import org.springframework.http.HttpEntity;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpMethod;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+import org.springframework.retry.RetryCallback;
+import org.springframework.retry.RetryContext;
+import org.springframework.retry.RetryListener;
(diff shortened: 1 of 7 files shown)


plugins/flexibleengine/src/main/java/org/eclipse/xpanse/orchestrator/plugin/flexibleengine/monitor/utils/RetryTemplateService.java
```diff
@@ -0,0 +1,168 @@
...
+     *
+     * @param httpRequestBase httpRequestBase
+     * @param requestBody     requestBody
+     * @return BatchListMetricDataResponse
+     */
+    @Retryable(retryFor = RestClientException.class, maxAttempts = RETRY_TIMES,
+            backoff = @Backoff(delay = DELAY_MILLISECONDS, multiplier = DELAY_MULTIPLIER))
+    public BatchListMetricDataResponse batchQueryMetricData(HttpRequestBase httpRequestBase,
+                                                            String requestBody) {
+        HttpEntity<String> entity = new HttpEntity<>(requestBody,
```
swaroopar, line 120: everywhere -> can we call it requestEntity instead of just entity? so that is is more clear.

plugins/flexibleengine/src/main/java/org/eclipse/xpanse/orchestrator/plugin/flexibleengine/monitor/utils/RetryTemplateService.java
```diff
@@ -0,0 +1,168 @@
...
+     * Batch query metric data using RestTemplate with Spring-Retry.
+     *
+     * @param httpRequestBase httpRequestBase
+     * @param requestBody     requestBody
+     * @return BatchListMetricDataResponse
+     */
+    @Retryable(retryFor = RestClientException.class, maxAttempts = RETRY_TIMES,
+            backoff = @Backoff(delay = DELAY_MILLISECONDS, multiplier = DELAY_MULTIPLIER))
+    public BatchListMetricDataResponse batchQueryMetricData(HttpRequestBase httpRequestBase,
+                                                            String requestBody) {
```
swaroopar, line 119: Can we have the actual request object here and then covert it to String inside this method?

plugins/flexibleengine/src/main/java/org/eclipse/xpanse/orchestrator/plugin/flexibleengine/monitor/utils/RetryTemplateService.java
```diff
@@ -0,0 +1,168 @@
...
+    private static final double DELAY_MULTIPLIER = 1.5D;
+    private final RestTemplate restTemplate = new RestTemplate();
+
+    /**
+     * Query project info using RestTemplate with Spring-Retry.
+     *
+     * @param httpRequestBase httpRequestBase
+     * @return Project
+     */
+    @Retryable(retryFor = RestClientException.class, maxAttempts = RETRY_TIMES,
```
swaroopar, line 54: can we also add a `RetryListener` which logs when error happens and retry is being performed?

swaroopar (changes requested):
comments added.

plugins/flexibleengine/src/main/java/org/eclipse/xpanse/orchestrator/plugin/flexibleengine/monitor/utils/RetryTemplateService.java
baixinsui: Ok.

plugins/flexibleengine/src/main/java/org/eclipse/xpanse/orchestrator/plugin/flexibleengine/monitor/utils/RetryTemplateService.java
baixinsui: Ok.

swaroopar (approved):
looks good.

Issue #423, opened 2023-06-08
getCredentialCapabilitiesByCsp - ReturnType does not match


swaroopar:
- [x] Return type of getCredentialCapabilitiesByCsp method is `AbstractCredentialInfo` but since this is an abstract class, what we return is actually `CredentialDefinition`. But the OpenApi document only shows that `AbstractCredentialInfo` as the return type. Clients cannot work with this.
- [x] Also we return `CredentialVariable` with value as null. This is misleading.

iskey:
`CredentialDefinition` is just one sort of the CredentialInfo object similar to INPUT, I think this should not be the only one. So I feel like keeping it as-is would be better.

For the first one, we can use `@Schema(anyOf = {CredentialDefinition.class})` to export the API document like this:
![image](https://github.com/eclipse-xpanse/xpanse/assets/1907997/483cd2a7-906a-47f2-9faa-330e2c9f1e5d)

For the second one, I don't get you, do you mean the `CredentialVariable` object is null or the `value` field?

swaroopar:
@iskey thanks for checking. 
If @AnyOf works then its perfect and we can use it. But we must check if for example, our UI client generator is able to generate client properly with it. In my past experience, AnyOf, and AllOf were pretty new features on OpenAPI and we had issues on client generation. 

For second point, value field is always null.

iskey:
@swaroopar , agree, we should check it works well first.

For the second point, would it be good to issue a new class object without the `value` field? or any magic way for it?

Alice1319:
I'll take this

iskey:
The second point is still open.

iskey:
> @iskey i feel the cleanest way would be to have a new type defined. May be we can remove the value from CredentialVariable and create another type which extends CredentialVariable and adds a the value field?

Yes, it makes sense.

swaroopar:
done. Both implemented.

PR #461, opened 2023-06-14
add CredentialValue.

Alice1319:
fixes second point of #423


swaroopar (changes requested):
@Alice1319 the getCredentialCapabilitiesByCsp method is still showing the value field.
only the 'getCredential' should have the value field in response. and 'addCredential' and 'updateCredential' must allow the value field in request.

swaroopar (approved):
Looks good

PR #463, opened 2023-06-14
update getCredentials method

jinyangyang222:
fiexd #447

### DIFF
--- a/modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/credential/CredentialCenter.java
+++ b/modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/credential/CredentialCenter.java
@@ -104,9 +104,11 @@ public class CredentialCenter {
         if (Objects.nonNull(requestedCredentialType)) {
             AbstractCredentialInfo abstractCredentialInfo =
                     this.credentialsStore.getCredential(csp, requestedCredentialType, xpanseUser);
-
             if (Objects.nonNull(abstractCredentialInfo)) {
+                checkNullParamAndFillValueFromEnv(abstractCredentialInfo);
                 abstractCredentialInfos.add(abstractCredentialInfo);
+            } else {
+                addCredentialInfoFromEnv(csp, abstractCredentialInfos);
             }
             return abstractCredentialInfos;
         } else {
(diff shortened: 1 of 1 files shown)


swaroopar (changes requested):
comments added.

modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/credential/CredentialCenter.java
```diff
@@ -319,4 +326,35 @@ private boolean isAnyMandatoryCredentialVariableMissing(
                         && Objects.isNull(credentialVariable.getValue()));
     }
 
+    private void checkNullParamAndFillValueFromEnv(AbstractCredentialInfo abstractCredentialInfo) {
+        if (abstractCredentialInfo.getType().equals(CredentialType.VARIABLES)) {
+            CredentialVariables credentialVariables =
+                    (CredentialVariables) abstractCredentialInfo;
+            for (CredentialVariable variable : credentialVariables.getVariables()) {
+                if (Objects.isNull(variable.getValue())
+                        || Objects.equals("null", variable.getValue())) {
```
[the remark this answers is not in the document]
jinyangyang222: fixed.

modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/credential/CredentialCenter.java
```diff
@@ -319,4 +326,35 @@ private boolean isAnyMandatoryCredentialVariableMissing(
...
+                        || Objects.equals(variable.getValue(), "null")) {
+                    variable.setValue(System.getenv(variable.getName()));
+                }
+            }
+        }
+    }
+
+    private void addCredentialInfoFromEnv(Csp csp,
+                                          List<AbstractCredentialInfo> abstractCredentialInfos) {
+        AbstractCredentialInfo credentialInfoFromEnv = findCredentialInfoFromEnv(csp);
```
swaroopar, line 344: findCredentialInfoFromEnv - can return null. 
So we need a check here to add only non-null values here.

swaroopar (changes requested):
@jinyangyang222 small comment added.

modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/credential/CredentialCenter.java
jinyangyang222: ok ok, I made an oversight, it's fixed.

swaroopar (approved):
@jinyangyang222 thank you. Looks good.

PR #465, opened 2023-06-14
use caffeine cache for credentials cache

### LINKED ISSUE #398: Use spring cache mechanism for credentials cache
currently the credentials cache is implemented as simple map and is cleared as part of the new credential creation method.
We ideally need the expired cache to be cleared as soon as the validity is expired. 
This can be achieved using [spring cache][(https://docs.spring.io/spring-boot/docs/2.1.6.RELEASE/reference/html/boot-features-caching.html](https://docs.spring.io/spring-framework/reference/integration/cache/annotations.html))

- [x]  Enable spring caching in the runtime
- [x]  It is enough to use  `ConcurrentHashMap` which is the default cache manager, if it supports automatic cache eviction. Otherwise check other implementations which provide this. 
- [x] Move all write, reads via the cache. 
- [x] Add cache eviction based on the cache expiry. 
- [x] Remove the `expiredTime` field as the evicted caches are removed automatically from the spring cache framework. https://github.com/eclipse-xpanse/xpanse/blob/fbd30b683dde869ca3680794fb53ba118d90f9a9/modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/CredentialDefinition.java#L38
- [x] unit tests

P.S - once this is implemented, we can plan to move metrics cache also to this.

swaroopar:
fixes #398


modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/cache/CaffeineCredentialCacheManager.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+@Component
+public class CaffeineCredentialCacheManager {
+
+    private static final Cache<CredentialCacheKey, AbstractCredentialInfo> credentialsCache =
+            credentialsCache();
+
+    private static Cache<CredentialCacheKey, AbstractCredentialInfo> credentialsCache() {
+        return Caffeine.newBuilder()
+                .expireAfter(new CredentialCacheExpiry())
```
baixinsui, line 24: can we add removeListener() to output  log？

modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/credential/CredentialCenter.java
```diff
@@ -278,23 +279,10 @@ private CredentialVariables createCredentialVariablesObject(
                 createCredential.getXpanseUser(),
                 createCredential.getName(), createCredential.getDescription(),
                 createCredential.getType(), createCredential.getVariables());
-        if (Objects.isNull(createCredential.getTimeToLive())) {
-            createCredential.setTimeToLive(DEFAULT_TIMEOUT_SECONDS);
-        }
-        credential.setExpiredTime(
-                System.currentTimeMillis() + createCredential.getTimeToLive() * 1000);
+        credential.setTimeToLive(Integer.toUnsignedLong(createCredential.getTimeToLive()));
```
baixinsui, line 282: We can unify the type of  the filed 'timeToLive' in both classes AbstractCredentialInfo and CreateCredential.

baixinsui (commented):
Comments add.

modules/orchestrator/src/main/java/org/eclipse/xpanse/orchestrator/credential/CredentialCenter.java
swaroopar: fixed

baixinsui (approved):
Looks good

modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/cache/CaffeineCredentialCacheManager.java
```diff
@@ -0,0 +1,52 @@
...
+import com.github.benmanes.caffeine.cache.RemovalCause;
+import lombok.extern.slf4j.Slf4j;
+import org.eclipse.xpanse.modules.credential.AbstractCredentialInfo;
+import org.springframework.stereotype.Component;
+
+/**
+ * Class which instantiates credential cache and handles all read-write access to the cache.
+ */
+@Component
+@Slf4j
```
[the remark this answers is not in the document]
swaroopar: the usual false positive warnings from sonatype.

iskey (approved):
looks good.

PR #474, opened 2023-06-16
fix CredentialApi

### LINKED ISSUE #473: Fix CredentialApi
1.Constructor error in CredentiaVariables class.
2.Add check xpanseUser is not null.

jinyangyang222:
fixed #473

### DIFF
--- a/modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/CredentialCenter.java
+++ b/modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/CredentialCenter.java
@@ -7,6 +7,7 @@
 package org.eclipse.xpanse.modules.credential;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Objects;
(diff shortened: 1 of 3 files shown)


swaroopar:
@jinyangyang222 pls fix merge conflicts

modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/CredentialCenter.java
```diff
@@ -131,6 +131,9 @@ public String getCredentialOpenApiUrl(Csp csp, CredentialType type) {
      * @return the credentials.
      */
     public List<AbstractCredentialInfo> getCredentialsByUser(String xpanseUser) {
+        if (StringUtils.isBlank(xpanseUser)) {
```
swaroopar, line 134: Hi @jinyangyang222 
in the getCredentialsByUser  API method -> we can make user parameter as "required".
And here instead of returning null, we must return empty list.

swaroopar (changes requested):
comments added.

swaroopar (approved):
LGTM

PR #476, opened 2023-06-16
Refactor modules

### LINKED ISSUE #475: Refactor modules
As the modules are in chaos, we must do some refactoring things about the modules now.

iskey:
fix #475


plugin/src/main/java/org/eclipse/xpanse/plugin/.gitignore
swaroopar, line 1: this file can be removed.

swaroopar (approved):
LGTM

PR #481, opened 2023-06-16
refactor data model packages

### LINKED ISSUE #479: package restructuring - data models
- [ ] All data models should by default reside in "models" module. 
- [ ] Only the data modules that is needed by plugins must be defined in the plugin/orchestrator module. 
- [ ] APIs must translate the data models from the API to the models required by Plugins.

swaroopar:
fix #479


iskey (approved):
lgtm

PR #485, opened 2023-06-16
split OrchestratorApi

### LINKED ISSUE #478: Split OrchestratorApi class and rename plugin module
- [ ] Rename the plugin module  to orchestrator -> and rename the existing orchestrator to "register"
- [ ] Split OrchestratorApi -> ServiceRegisterApi, ServiceDeployerApi
- [ ] Keep ServiceRegisterApi implementation in "register" module
- [ ] Keep "ServiceDeployerApi" in deployment module. 

Note - this must be done after #475 is merged.

iskey:
fix #478


modules/api/src/main/java/org/eclipse/xpanse/api/config/ResponseValidator.java
```diff
@@ -34,7 +34,7 @@ public ResponseValidator(@Autowired Validator validator) {
     }
 
     @AfterReturning(
-            pointcut = "execution(public * org.eclipse.xpanse.api.OrchestratorApi.*(..))",
+            pointcut = "execution(public * org.eclipse.xpanse.api.ServiceDeployerApi.*(..))",
```
swaroopar, line 37: here we need validation for all APIs. We need the pointcut to execute on all *Api classes. 
Looks like it was a bug that we missed MonitorApi also until now.

modules/api/pom.xml
```diff
@@ -66,5 +66,11 @@
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-aop</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.eclipse.xpanse.modules</groupId>
+            <artifactId>monitor</artifactId>
+            <version>1.0.0-SNAPSHOT</version>
```
swaroopar, line 72: replace with ${project.version}

swaroopar (changes requested):
minor comments added. 
Just another small thing. Package name of classes in "orchestrator" package still has "plugin" in it. Instead of orchestrator

iskey:
> minor comments added. Just another small thing. Package name of classes in "orchestrator" package still has "plugin" in it. Instead of orchestrator

all comments are fixed.

swaroopar (approved):
looks good

PR #488, opened 2023-06-18
fill default value field only for API calls

swaroopar:
Default value for the "value" field in credential definition must be filled only for the getCredentialCapabilitiesByCsp API method. 

In the current version, the value is set always and it is causing null checks to always fail.

Related to #423


iskey (approved):
looks good.

PR #493, opened 2023-06-19
fix cred return type and uri

### LINKED ISSUE #464: Fix return types for add, update, delete credentials
Currently the add, update and delete are returning boolean value which is not useful and also confusing for clients. 
Make these  methods return void and HTTP response code 204 (NO_CONTENT)

If there are errors, we must simply update the HTTP response code instead of sending boolean values.

swaroopar:
fixes #464


iskey (approved):
lgtm.

PR #498, opened 2023-06-20
Cache Monitor lastKnownMetric using Caffeine Cache

baixinsui:


modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/deployers/terraform/utils/SystemCmd.java
```diff
@@ -133,11 +135,6 @@ private void readProcessOutput(Process process, SystemCmdResult systemCmdResult)
         ExecutorService threadToReadStdErr = Executors.newSingleThreadExecutor();
         Future<String> stdErrFuture =
                 threadToReadStdErr.submit(() -> readStream(stdErrorReader, contextMap));
-
-        while (!stdOutFuture.isDone() || !stdErrFuture.isDone()) {
-            log.debug("Command output and error streams are still being read.");
```
swaroopar, line 138: @baixinsui this is just for debug level. If you feel it is too much for debug level too, may be we can move it to trace instead of removing it completely?

baixinsui: Ok. When the build task fails, even the log file package cannot be downloaded successfully due to too many logs and poor network connectivity

swaroopar: ah ok. for me this issue was not there until now. 
We can keep this log only with trace. or we leave it with debug and ensure the tests run only with "info" level.

baixinsui: I removed the commit about the SystemCmd

swaroopar (approved):
LGTM

PR #499, opened 2023-06-20
add monitorResourceType variable to monitor return resource

### LINKED ISSUE #497: Monitor - Add MonitorResourceType variable to monitor return resource
We should keep the resource names of monitor input parameters and returned results consistent. For example, if the input parameter is "cpu", the return parameter should also be "cpu". HuaweiCloud and FlexibleEngine currently needs to be modified.

jinyangyang222:
fixed #497

### DIFF
--- a/modules/models/src/main/java/org/eclipse/xpanse/modules/models/monitor/Metric.java
+++ b/modules/models/src/main/java/org/eclipse/xpanse/modules/models/monitor/Metric.java
@@ -12,6 +12,7 @@ import java.util.Map;
 import lombok.Data;
 import org.eclipse.xpanse.modules.models.monitor.enums.MetricType;
 import org.eclipse.xpanse.modules.models.monitor.enums.MetricUnit;
+import org.eclipse.xpanse.modules.models.monitor.enums.MonitorResourceType;
 
 /**
  * One kind of metric.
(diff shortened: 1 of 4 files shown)


modules/models/src/main/java/org/eclipse/xpanse/modules/models/monitor/Metric.java
```diff
@@ -30,6 +31,10 @@ public class Metric {
     @Schema(description = "The type of the metric.")
     MetricType type;
 
+    @NotNull
+    @Schema(description = "The resources type of the monitor.")
```
swaroopar, line 35: @jinyangyang222 description should be "the resource type  of the metric."

jinyangyang222: ok, fixed.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/monitor/Metric.java
```diff
@@ -30,6 +31,10 @@ public class Metric {
     @Schema(description = "The type of the metric.")
     MetricType type;
 
+    @NotNull
+    @Schema(description = "The resource type of the metric.")
+    MonitorResourceType monitorResourceType;
+
     @NotNull
     @Schema(description = "The unit of the metric.")
     MetricUnit unit;
```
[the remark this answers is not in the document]
swaroopar: @sonatype-lift ignore

swaroopar (approved):
LGTM

PR #507, opened 2023-06-22
remove unused dependency

swaroopar:
Fix to remove this security finding - https://github.com/eclipse-xpanse/xpanse/security/dependabot/1


iskey (approved):
looks good.

PR #511, opened 2023-06-23
GET to return only from cache

### LINKED ISSUE #510: return only cached credentials info in getCredentialsByUser
Currently the getCredentialsByUser returns complete joined credentials. This seem to have other side effects.
hence return only whatever is stored in cache.

swaroopar:
fixes #510


iskey (approved):
LGTM.

PR #520, opened 2023-06-25
fix cache query when optional query parameters are missing

### LINKED ISSUE #519: Wrong cache query when optional parameters are not passed
When optional parameters are not passed in the API, then we pass also null to cache query and this leads to no cache matches and the response is always null.

swaroopar:
fixes #519


iskey (approved):
LGTM

Issue #532, opened 2023-06-26
Add tests to increase test coverage of models module


swaroopar:
Add valid unit tests to all possible code in database module in order to increase overall test coverage.

jinyangyang222:
I will take this.

PR #540, opened 2023-06-27
add tests of models module

jinyangyang222:
related to #532


swaroopar (approved):
Thanks for this. LGTM.

PR #546, opened 2023-06-28
add new custom exception

### LINKED ISSUE #544: Add new exception for register server update
Currently we throw `IllegalArgumentException` when update of an already registered service is rejected after validations. Add a new custom exception for the same.

swaroopar:
fixes #544


iskey (approved):
LGTM

PR #560, opened 2023-07-03
display credential type in register, update and catalog UI

### LINKED ISSUE #528: UI - display credential type in register, update and catalog UI
We have added a credential type field in OCL - https://github.com/eclipse-xpanse/xpanse/blob/main/modules/models/src/main/java/org/eclipse/xpanse/modules/models/service/register/Deployment.java#L34
We must  now display this in different places

- [ ] Register UI
- [ ] Service update UI
- [ ] In the service catalog of CSP users.

WangLiNaruto:
fixed #528
image
image
image


swaroopar (approved):
LGTM

Issue #562, opened 2023-07-04
Enable spring security with oauth2


swaroopar:
- [ ] Add new spring profile called - `zitadel`. This is necessary to ensure that we can enable security only when the feature is fully available and also switch between different identity providers. 
- [ ] Add new `security` maven module under `modules`.
- [ ] Add OIDC (oauth2) based authentication using spring security. 
- [ ] Security bean to be initialized only when `zitadel `profile is activated.
- [ ] Use 'OpaqueToken'
- [ ] Protect all `/xpanse/` URI with oauth. 
- [ ] Return 401 if authentication fails. 
- [ ] Document how to start the application by passing `client-secret` variable.

Note - Spring security filter need not do any redirection. It must  only validate the provided authorization header and return error 401 if the token validation fails. 
Depends on  #563

baixinsui:
I will take it.

PR #575, opened 2023-07-05
Add pbkdf2 + spring security for sensitive field

### LINKED ISSUE #554: Add pbkdf2 + spring security for sensitive field

WangLiNaruto:
Fixed #554

### DIFF
--- a/modules/common/pom.xml
+++ b/modules/common/pom.xml
@@ -28,5 +28,9 @@
             <version>${project.version}</version>
             <scope>compile</scope>
         </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-security</artifactId>
+        </dependency>
     </dependencies>
 </project>
--- a/modules/api/pom.xml
+++ b/modules/api/pom.xml
@@ -77,5 +77,15 @@
             <version>${instancio.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-test</artifactId>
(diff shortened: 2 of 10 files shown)


modules/common/src/main/java/org/eclipse/xpanse/common/security/config/WebSecurityConfig.java
```diff
@@ -0,0 +1,27 @@
...
+
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.security.crypto.password.Pbkdf2PasswordEncoder;
+
+/**
+ * Spring Security Config.
+ */
+@Configuration
+public class WebSecurityConfig {
```
swaroopar, line 17: Can we make this class name - SensitiveFieldEncoderConfig?
because we wil have spring web security for other purpose also soon.

modules/common/pom.xml
```diff
@@ -28,5 +28,9 @@
             <version>${project.version}</version>
             <scope>compile</scope>
         </dependency>
+        <dependency>
```
swaroopar, line 31: I would prefer to have this classes under new maven module "security". 
It will be added in #562 userstory. 
We can either add it now with your userstory or we can just inform @baixinsui to move this to the security module whenever he creates it.

swaroopar (changes requested):
minor comments added

modules/common/pom.xml
WangLiNaruto: @swaroopar  sure,Let do that way after the PR of xinsui is merged

swaroopar (approved):
LGTM

PR #580, opened 2023-07-07
fix error in cache when masked isSensitive fields

### LINKED ISSUE #578: Fix error in cache when masked isSensitive fields
Fix the value with isSensitive in the cache, and the value in the previous method cache will also be masked.

jinyangyang222:
fixed #578

### DIFF
--- a/modules/credential/src/test/java/org/eclipse/xpanse/modules/credential/CredentialCenterTest.java
+++ b/modules/credential/src/test/java/org/eclipse/xpanse/modules/credential/CredentialCenterTest.java
@@ -6,6 +6,7 @@
 package org.eclipse.xpanse.modules.credential;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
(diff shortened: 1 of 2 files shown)


swaroopar (approved):
LGTM

modules/credential/src/test/java/org/eclipse/xpanse/modules/credential/CredentialCenterTest.java
```diff
@@ -258,7 +259,7 @@ void testGetCredentials_WithRequestedCredentialType() {
 
         credentialsStore.getCredential(csp, requestedCredentialType, xpanseUser);
 
-        assertEquals(Collections.singletonList(credential), result);
+        assertNotEquals(Collections.singletonList(credential), result);
```
swaroopar, line 262: If the list is not same, we can now add assertions to validate if the values inside the list are same.

swaroopar (changes requested):
extend test assertions pls.

swaroopar (approved):
LGTM

modules/credential/src/test/java/org/eclipse/xpanse/modules/credential/CredentialCenterTest.java
jinyangyang222: Okay, I added it.

PR #583, opened 2023-07-10
enable spring security oauth2 on api-client and swagger-ui

### LINKED ISSUE #562: Enable spring security with oauth2

### LINKED ISSUE #565: Enable Oauth2 on Swagger-UI
Swagger-UI must have Oauth2 enabled to provider login-ui redirection and execute all further requests with authorization

- [ ] Add configuration to enable authorization on swagger-ui. Below is the sample configuration
- [ ] Ensure the authentication is enabled only when `zitadel` profile is activated. 
- [ ] add a 'web' application on Zitadel for swagger-UI with correct redirect URL
- [ ] Document the application creation process.
- [ ] Provide documentation to override values of clientid, authurl and token url with startup.

```java
@SecurityScheme(name = "security_auth", type = SecuritySchemeType.OAUTH2,
        in = SecuritySchemeIn.HEADER,
        scheme = "bearer",
        flows = @OAuthFlows(authorizationCode = @OAuthFlow(
                authorizationUrl = "${springdoc.oAuthFlow.authorizationUrl}"
                , tokenUrl = "${springdoc.oAuthFlow.tokenUrl}", scopes = {
                @OAuthScope(name = "openid"),
                @OAuthScope(name = "email"),
                @OAuthScope(name = "profile"),
        })))
@Configuration
public class OpenApiConfig {
}
````

```yaml
springdoc:
  show-login-endpoint: true
  swagger-ui:
    oauth:
      clientId: xxxx
      use-pkce-with-authorization-code-grant: true
  oAuthFlow:
    authorizationUrl: http://localhost:8080/oauth/v2/authorize
    tokenUrl: http://localhost:8080/oauth/v2/token
```

baixinsui:
fixed #562
fixed #565

### DIFF
--- a/modules/security/src/main/java/org/eclipse/xpanse/modules/security/config/ZitadelOpenApiOauth2Config.java
+++ b/modules/security/src/main/java/org/eclipse/xpanse/modules/security/config/ZitadelOpenApiOauth2Config.java
@@ -1,0 +1,51 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * SPDX-FileCopyrightText: Huawei Inc.
+ *
+ */
+
+package org.eclipse.xpanse.modules.security.config;
+
+import io.swagger.v3.oas.annotations.OpenAPIDefinition;
+import io.swagger.v3.oas.annotations.enums.SecuritySchemeType;
+import io.swagger.v3.oas.annotations.info.Info;
+import io.swagger.v3.oas.annotations.security.OAuthFlow;
+import io.swagger.v3.oas.annotations.security.OAuthFlows;
+import io.swagger.v3.oas.annotations.security.OAuthScope;
+import io.swagger.v3.oas.annotations.security.SecurityRequirement;
+import io.swagger.v3.oas.annotations.security.SecurityScheme;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Profile;
+
+
+/**
+ * Configuration springdoc security oauth2.
+ */
+@Profile("zitadel")
+@OpenAPIDefinition(
+        info = @Info(
+                title = "Xpanse API",
+                description = "RESTful Services to interact with Xpanse runtime",
+                version = "${app.version}"
+        ),
+        security = @SecurityRequirement(name = "OAuth2 Flow",
+                scopes = {"openid", "email", "profile"})
+)
+@SecurityScheme(
+        name = "OAuth2 Flow",
+        type = SecuritySchemeType.OAUTH2,
+        flows = @OAuthFlows(
+                authorizationCode = @OAuthFlow(
+                        authorizationUrl = "${springdoc.oAuthFlow.authorizationUrl}",
+                        tokenUrl = "${springdoc.oAuthFlow.tokenUrl}",
+                        scopes = {
+                                @OAuthScope(name = "openid"),
+                                @OAuthScope(name = "email"),
+                                @OAuthScope(name = "profile")
+                        }
+                )
+        )
+)
+@Configuration
+public class ZitadelOpenApiOauth2Config {
+}
--- a/runtime/src/main/resources/application-zitadel.properties
+++ b/runtime/src/main/resources/application-zitadel.properties
@@ -1,0 +1,18 @@
+#
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Huawei Inc.
+#
+#
+# spring security oauth2 config
+logging.level.org.springframework.web=info
+logging.level.org.springframework.security=debug
+# set authorization server endpoint
+spring.security.oauth2.resourceserver.opaquetoken.introspection-uri=${authorization-server-endpoint}/oauth/v2/introspect
+spring.security.oauth2.resourceserver.opaquetoken.client-id=${authorization-api-client-id}
+spring.security.oauth2.resourceserver.opaquetoken.client-secret=${authorization-api-client-secret}
+# springdoc openapi security oauth2 config
+springdoc.show-login-endpoint=true
+springdoc.swagger-ui.oauth.use-pkce-with-authorization-code-grant=true
+springdoc.swagger-ui.oauth.clientId=${authorization-swagger-ui-client-id}
+springdoc.oAuthFlow.authorizationUrl=${authorization-server-endpoint}/oauth/v2/authorize
+springdoc.oAuthFlow.tokenUrl=${authorization-server-endpoint}/oauth/v2/token
--- a/README.md
+++ b/README.md
@@ -11,22 +11,20 @@
   </a>
 </p>
 
-Xpanse is an Open Source project allowing to easily implement native managed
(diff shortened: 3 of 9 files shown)


modules/security/src/main/java/org/eclipse/xpanse/modules/security/config/WebSecurityConfig.java
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * Configuration applied on all web endpoints defined for this
+ * application. Any configuration on specific resources is applied
+ * in addition to these global rules.
+ */
+@Slf4j
+@Profile("zitadel")
+@Configuration
+@EnableWebSecurity
+public class WebSecurityConfig {
```
swaroopar, line 27: Since it is only zitadel specific config, can we add 'Zitadel' to the classname?

baixinsui: OK

runtime/src/main/resources/application.properties
```diff
@@ -3,6 +3,8 @@
 # SPDX-FileCopyrightText: Huawei Inc.
 #
 #
+
+spring.profiles.active=zitadel
```
swaroopar, line 7: it must not be activated by default, It must only be activated on demand.

runtime/src/main/resources/application-zitadel.properties
```diff
@@ -0,0 +1,22 @@
...
+authorization-api-client-id=221665786966638595@eclipse-xpanse
+authorization-api-client-secret=2S7yZR4jWauyE3wLHM1h5asc0WNWdBUXAw2Ill3Pmu0qM4D3LWBCwikXbsG81ycl
+spring.security.oauth2.resourceserver.opaquetoken.introspection-uri=${authorization-server-endpoint}/oauth/v2/introspect
+spring.security.oauth2.resourceserver.opaquetoken.client-id=${authorization-api-client-id}
+spring.security.oauth2.resourceserver.opaquetoken.client-secret=${authorization-api-client-secret}
+# springdoc openapi security oauth2 config
+authorization-useragent-client-id=221664356859969539@eclipse-xpanse
+springdoc.show-login-endpoint=true
+springdoc.swagger-ui.oauth.use-pkce-with-authorization-code-grant=true
+springdoc.swagger-ui.oauth.clientId=${authorization-useragent-client-id}
```
swaroopar, line 20: we can directly assign the value here.

runtime/src/main/resources/application.properties
baixinsui: OK.

modules/security/src/main/java/org/eclipse/xpanse/modules/security/config/OpenApiOauth2Config.java
```diff
@@ -0,0 +1,49 @@
...
+                        tokenUrl = "${springdoc.oAuthFlow.tokenUrl}",
+                        scopes = {
+                                @OAuthScope(name = "openid"),
+                                @OAuthScope(name = "email"),
+                                @OAuthScope(name = "profile")
+                        }
+                )
+        )
+)
+@Configuration
```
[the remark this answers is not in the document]
baixinsui: OK.

runtime/src/main/resources/application-zitadel.properties
swaroopar: @baixinsui  ok. then we can keep it as it is. Just that we can replace the `authorization-useragent-client-id` with `authorization-swagger-ui-client-id`

swaroopar (changes requested):
comments added.

runtime/src/main/resources/application-zitadel.properties
baixinsui: OK.

runtime/src/main/resources/application-zitadel.properties
```diff
@@ -0,0 +1,22 @@
+#
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Huawei Inc.
+#
+#
+# spring security oauth2 config
+logging.level.org.springframework.web=info
+logging.level.org.springframework.security=debug
+# set authorization server endpoint
+authorization-server-endpoint=https://iam.xpanse.site
```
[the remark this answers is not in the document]
swaroopar: @baixinsui this point is still not fixed?

modules/security/src/main/java/org/eclipse/xpanse/modules/security/config/ZitadelOpenApiOauth2Config.java
```diff
@@ -0,0 +1,52 @@
...
+import io.swagger.v3.oas.annotations.security.SecurityScheme;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Profile;
+
+
+/**
+ * Configuration springdoc security oauth2.
+ */
+//@Profile("zitadel")
```
swaroopar, line 25: this must be uncommented?

swaroopar (changes requested):
comments added.

runtime/src/main/resources/application-zitadel.properties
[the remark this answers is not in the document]
baixinsui: chrome_Z9oQmaVgHS
I think it is same as this to keep these properties.

modules/security/src/main/java/org/eclipse/xpanse/modules/security/config/ZitadelOpenApiOauth2Config.java
baixinsui: I will fix this error.

runtime/src/main/resources/application-zitadel.properties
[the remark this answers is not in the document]
swaroopar: @baixinsui its a different topic here. 
What i meant is, we must remove our test Zitadel configuration from here and move it to a different profile called 'zitadel-testbed'. So anyone can simply activate both `zitadel` and `zitadel-testbed` profiles to use the test bed, In production, they will have to activate only `zitadel` profile and then pass other properties as env variables or startup parameters

swaroopar (changes requested):
comment added.

swaroopar (approved):
LGTM

PR #590, opened 2023-07-12
Enable authority check on APIs

### LINKED ISSUE #567: REST API - Enable RBAC
- [ ] Add 3 roles to `eclipse-xpanse` project on  Zitadel - admin, user, csp
- [ ] Add authorization required on API - at class level or at method level
- [ ] Validate the 'token' claims contains the list of roles allocated to user under claim name `urn:zitadel:iam:org:project:roles` - https://zitadel.com/docs/apis/openidoauth/claims
- [ ] Proceed only when the required role is available in the list of role the user has in the claim `urn:zitadel:iam:org:project:roles`.
- [ ] If role is allowed to user, then return 403 error.
- [ ] Ensure the role based authorization is executed only when `zitadel` profile is enabled.
- [ ] If no role is received in the claims, then assume `user` roles.

Note - we might have to use the customizer for extracting roles - https://docs.spring.io/spring-security/reference/reactive/oauth2/resource-server/opaque-token.html#webflux-oauth2resourceserver-opaque-authorization-extraction

baixinsui:
fixed #567


modules/security/src/main/java/org/eclipse/xpanse/modules/security/config/ZitadelAuthoritiesOpaqueTokenIntrospector.java
```diff
@@ -0,0 +1,94 @@
...
+import org.springframework.security.oauth2.core.DefaultOAuth2AuthenticatedPrincipal;
+import org.springframework.security.oauth2.core.OAuth2AuthenticatedPrincipal;
+import org.springframework.security.oauth2.server.resource.introspection.NimbusOpaqueTokenIntrospector;
+import org.springframework.security.oauth2.server.resource.introspection.OpaqueTokenIntrospector;
+import org.springframework.util.CollectionUtils;
+
+/**
+ * Customize the ZitadelAuthoritiesOpaqueTokenIntrospector implements OpaqueTokenIntrospector.
+ */
+@Slf4j
```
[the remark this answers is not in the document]
baixinsui: @sonatype-lift ignoreall

modules/api/src/main/java/org/eclipse/xpanse/api/MonitorApi.java
```diff
@@ -34,6 +35,7 @@
 @RestController
 @Validated
 @RequestMapping("/xpanse")
+@Secured({"admin", "user"})
```
swaroopar, line 38: is it possible to keep roles in some constants page or enum and then use it everywhere?

modules/api/src/main/java/org/eclipse/xpanse/api/ServiceRegisterApi.java
```diff
@@ -54,6 +55,7 @@
 @RestController
 @RequestMapping("/xpanse")
 @CrossOrigin
+@Secured({"admin", "csp"})
```
swaroopar, line 58: is this annotation ignored by Spring when `zitadel` profile is not enabled?

swaroopar (changes requested):
@baixinsui minor comment added.

modules/api/src/main/java/org/eclipse/xpanse/api/MonitorApi.java
baixinsui: Ok.

modules/api/src/main/java/org/eclipse/xpanse/api/ServiceRegisterApi.java
baixinsui: Yes. It only works when the @EnableMethodSecurity(securedEnabled = true) is enabled.

PR #591, opened 2023-07-13
Add pbkdf2 encode/decode for deployment sensitive variables

### LINKED ISSUE #557: Add AES encode/decode for deployment sensitive variables

WangLiNaruto:
Fixed #557


swaroopar (changes requested):
same question here as well. where and how is the decoding handled?

iskey:
> same question here as well. where and how is the decoding handled?

My fault, I misremembered the pbkdf2 and AES, we should choose AES here.

WangLiNaruto:
looks good

iskey (changes requested):
Please turn to AES instead.

PR #592, opened 2023-07-13
Add pbkdf2 encode/decode for credential sensitive field

### LINKED ISSUE #558: Add AES encode/decode for credential sensitive field

WangLiNaruto:
Fixed #558

### DIFF
--- a/modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/CredentialCenter.java
+++ b/modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/CredentialCenter.java
@@ -27,6 +27,7 @@ import org.eclipse.xpanse.modules.models.service.common.enums.Csp;
 import org.eclipse.xpanse.modules.orchestrator.OrchestratorPlugin;
 import org.eclipse.xpanse.modules.orchestrator.PluginManager;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.security.crypto.password.Pbkdf2PasswordEncoder;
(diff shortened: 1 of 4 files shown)


swaroopar (changes requested):
@WangLiNaruto we are only encoding the value and storing. How/where are we handling the decoding?

WangLiNaruto:
> @WangLiNaruto we are only encoding the value and storing. How/where are we handling the decoding?

@swaroopar Yes, we just encoded sensitive fields and stored

WangLiNaruto:
> @WangLiNaruto we are only encoding the value and storing. How/where are we handling the decoding?

modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/CredentialCenter.java
```diff
@@ -174,7 +179,15 @@ public void addCredential(CreateCredential createCredential) {
...
                             createCredential.getCsp(), createCredential.getType()));
         }
-        createCredential(credential);
+    }
+
+    private void encodeSensitiveVariables(CreateCredential createCredential) {
+        List<CredentialVariable> variables = createCredential.getVariables();
+        variables.stream().forEach(variable -> {
+            if (!Objects.isNull(variable) && variable.getIsSensitive()) {
+                variable.setValue(passwordEncoder.encode(variable.getValue()));
```
baixinsui, line 188: When variable.getValue() is null, encode (null) will throw NullPointerException，We should add a judgment condition Objects.isNull(variable.getValue()) ;

PR #613, opened 2023-07-19
Move health API  && Add health check of backend systems contain Identity provider and Database

### LINKED ISSUE #604: move health API to a separate class
currently health API is part of the ServiceRegisterApi class.

- [ ] Move it to another class called `AdminServicesApi `class.
- [ ] Rename the method from health to `healthCheck`

### LINKED ISSUE #605: Add identity provider to healthcheck method
When Authentication is enabled, 
the healthcheck method must additionally check the status of the IdentityProvider and return its status

- [ ] Extend `SystemStatus` object with a new type called `IdentityProviderStatus`
- [ ] `IdentityProviderStatus` -> endpoint: string, status: HealthStatus, ProviderType: new Enum, with one value -> Zitadel
- [ ] endpoint -> should return whatever is read from the spring properties.
- [ ] HealthStatus should be based on response of https://zitadel.com/docs/apis/observability/health

### LINKED ISSUE #617: Add database to healthcheck method
Similar to #605 , we must add Database to the healthcheck method. 
EndPoint msust be  the full jdbc connection url. 

This depends on #605

baixinsui:
fixed #604 
fixed #605
fixed #617


modules/api/src/main/java/org/eclipse/xpanse/api/AdminServicesApi.java
```diff
@@ -0,0 +1,81 @@
...
+import org.springframework.web.bind.annotation.RestController;
+
+/**
+ * Admin services Api.
+ */
+@Slf4j
+@CrossOrigin
+@RestController
+@RequestMapping("/xpanse")
+@Secured({ROLE_ADMIN})
```
swaroopar, line 42: @baixinsui this is different to roles mentioned in line 60.

modules/api/src/main/java/org/eclipse/xpanse/api/AdminServicesApi.java
```diff
@@ -0,0 +1,81 @@
...
+
+    /**
+     * Method to find out the current state of the system.
+     *
+     * @return Returns the current state of the system.
+     */
+    @Tag(name = "Admin", description = "APIs for administrating Xpanse")
+    @GetMapping(value = "/health", produces = MediaType.APPLICATION_JSON_VALUE)
+    @ResponseStatus(HttpStatus.OK)
+    @Secured({ROLE_ADMIN, ROLE_CSP, ROLE_USER})
```
swaroopar, line 60: @baixinsui 
I think I made a mistake while analyzing this topic. 
For  the 'user' and 'csp' role, we must only return the 'SystemStatus' without the status of other integrated systems such as  Identitiy provider. 

Only for role 'admin' we must return full healthcheck response with IdentityProviderStatus. 

If you agree, we can rename the old `health` method to `systemStatus` which returns only SystemStatus without IdentityProviderStatus. 
add new method healthCheck with a new data model 'HealthCheck' which has SystemStatus + IdentityProviderStatus

swaroopar (changes requested):
@baixinsui comments added.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/admin/IdentityProviderStatus.java
```diff
@@ -0,0 +1,33 @@
...
+import jakarta.validation.constraints.NotNull;
+import lombok.Data;
+import org.eclipse.xpanse.modules.models.admin.enums.HealthStatus;
+import org.eclipse.xpanse.modules.models.admin.enums.IdentityProviderType;
+
+/**
+ * Describes health status of the identity provider.
+ */
+@Data
+public class IdentityProviderStatus {
```
swaroopar, line 19: Even here there is a small change in the requirement. 
Instead of separate type for each 'integrated' component, we need a common type that can fit all types of systems. 

BackendSystem ->
Name(string) - value should be 'zitadel' if Zitadel is enabled, 
Endpoint(string)
SystemStatus
SystemTypeEnum -> current values - IdentityProvider, Database

the HealthCheck object must contain a List<BackendSystem>

sorry for the confusion.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/admin/IdentityProviderStatus.java
```diff
@@ -0,0 +1,33 @@
...
+
+    @NotNull
+    @Schema(description = "The type of identity provider.")
+    private IdentityProviderType identityProviderType;
```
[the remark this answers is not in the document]
baixinsui: @sonatype-lift ignore

modules/api/src/main/java/org/eclipse/xpanse/api/AdminServicesApi.java
baixinsui: Ok. I will restore the old method with path 'system/status'  for 'csp' and 'role', update the new method for 'admin'

modules/models/src/main/java/org/eclipse/xpanse/modules/models/admin/IdentityProviderStatus.java
baixinsui: a List ? which object should be a list? the error message?

modules/api/src/main/java/org/eclipse/xpanse/api/AdminServicesApi.java
swaroopar: @baixinsui in my opinion, the end users and csp users need not know internal details of the implementation. 
we can show them only IAM status is OK or NOK. But end point details need not be revelated to them.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/admin/IdentityProviderStatus.java
swaroopar: list of backend system. We just need one generic type which covers identity provider and database

modules/api/src/main/java/org/eclipse/xpanse/api/AdminServicesApi.java
baixinsui: Ok. I don't think it's necessary to add new method. In this method, we can obtain the roles of current user. If roles contains' admin', we will return endpoint and detail information

modules/models/src/main/java/org/eclipse/xpanse/modules/models/admin/IdentityProviderStatus.java
baixinsui: Ok.

Alice1319:
looks good

swaroopar (approved):
LGTM

modules/api/src/main/java/org/eclipse/xpanse/api/AdminServicesApi.java
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    private void processShownFields(BackendSystemStatus backendSystemStatus) {
+        CurrentUserInfo currentUserInfo = null;
+        List<String> profileSet = Arrays.asList(activeProfiles.split(","));
+        if (profileSet.contains(IdentityProviderType.ZITADEL.toValue())) {
+            currentUserInfo = ZitadelUserAuthenticationConverter.getCurrentUserInfo();
+        }
+        boolean allFieldsShown = Objects.nonNull(currentUserInfo) && !CollectionUtils.isEmpty(
+                currentUserInfo.getRoles()) && currentUserInfo.getRoles().contains(ROLE_ADMIN);
```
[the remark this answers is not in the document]
baixinsui: @sonatype-lift ignoreall

PR #624, opened 2023-07-21
Display of roles required for a service in Swagger-UI

### LINKED ISSUE #616: Display of roles required for a service in Swagger-UI
As we have now introduced RBAC, it will be informative to also display the 'allowed roles' for each method in the Swagger UI.

Some options - https://stackoverflow.com/questions/76153735/how-to-display-required-user-roles-in-for-openapi-ui

baixinsui:
fixed #616

### DIFF
--- a/modules/api/src/main/java/org/eclipse/xpanse/api/AdminServicesApi.java
+++ b/modules/api/src/main/java/org/eclipse/xpanse/api/AdminServicesApi.java
@@ -10,6 +10,7 @@ import static org.eclipse.xpanse.modules.models.security.constant.RoleConstants.
 import static org.eclipse.xpanse.modules.models.security.constant.RoleConstants.ROLE_CSP;
 import static org.eclipse.xpanse.modules.models.security.constant.RoleConstants.ROLE_USER;
 
+import io.swagger.v3.oas.annotations.Operation;
 import io.swagger.v3.oas.annotations.tags.Tag;
 import jakarta.annotation.Resource;
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


modules/security/src/main/java/org/eclipse/xpanse/modules/security/zitadel/config/UserRoleDescriptionCustomizer.java
```diff
@@ -0,0 +1,44 @@
...
+import org.springframework.security.access.annotation.Secured;
+import org.springframework.stereotype.Component;
+import org.springframework.web.method.HandlerMethod;
+
+/**
+ * Custom Operation of Swagger-UI to show required roles.
+ */
+@Profile("zitadel")
+@Component
+public class UserRoleDescriptionCustomizer implements OperationCustomizer {
```
swaroopar, line 23: @baixinsui do we really need this new annotation?
Can we not simply add the description when `secured` annotation is available in class or method level?

swaroopar (changes requested):
comment added

modules/security/src/main/java/org/eclipse/xpanse/modules/security/zitadel/config/UserRoleDescriptionCustomizer.java
baixinsui: I removed it and appended the required role to description of Operation.
image

swaroopar (approved):
LGTM

PR #625, opened 2023-07-21
fix opening service openAPI fails with 401

### LINKED ISSUE #621: opening service openAPI fails with 401
We currently provide an option to open the the OpenAPI documentation of the registered service from UI.  But this is not working after authentication was introduced. 

We currently use `window.open()` to open the API page but it looks like this method does not allow providing authentication header .
https://stackoverflow.com/questions/35332673/is-that-possible-to-send-header-information-in-window-openurl-method

image

baixinsui:
fix #621


swaroopar (approved):
LGTM

PR #642, opened 2023-07-24
fix identity provider status error

### LINKED ISSUE #639: Check Health of Identity Provider 'NOK' status is error.
Pz8Q03lWP9
Bug: Status of  Identity Provider is always 'NOK' but  it's actually 'OK'

baixinsui:
fixed #639


swaroopar (approved):
LGTM

PR #646, opened 2023-07-24
remove listCategories method

### LINKED ISSUE #629: remove listCategories method
listCategories currently returns the enum values of `Category` as list. 
This is not needed. All consumers of the API must directly refer to the enum. 
This method must be removed. 

PS - UI was using this service earlier but is now changed to use the enum directly.

jinyangyang222:
fixes #629


baixinsui (approved):
LGTM.

swaroopar (approved):
LGTM

PR #652, opened 2023-07-25
fix wrong documentation of mandatory field

### LINKED ISSUE #649: wrong documentation of mandatory field
In the  `getAvailableServicesTree` method, the category field is documented as optional. But this is wrong. Even OpenAPI ignores this config. All path variables are by default only mandatory. 

We must remove this config to maintain uniformity. 

https://github.com/eclipse-xpanse/xpanse/blob/df42f452bdb54190c65d7614c10faef4abcc368b/modules/api/src/main/java/org/eclipse/xpanse/api/ServiceDeployerApi.java#L239

jinyangyang222:
fixed #649


swaroopar (approved):
LGTM

PR #666, opened 2023-07-27
ServiceCatalogApi- fix availableServiceDetails method return empty object when the incoming ID is incorrect

### LINKED ISSUE #665: ServiceCatalogApi-  availableServiceDetails method, When the incoming ID is incorrect, an empty object is returned;
When the incoming ID is incorrect, an empty object is returned, like this.
[

](url)

Should ServiceNotRegisteredException be thrown at the database layer when the incoming id is incorrect,like this.
[

](url)

jinyangyang222:
fixed #665


swaroopar (approved):
LGTM

Issue #667, opened 2023-07-27
Csp  Category-- Throw an exception for unsupported values during serialization


jinyangyang222:
Csp -- Define new exception "UnsupportedCspException" 
Category --  Define new exception "UnsupportedCategoryException" 

In the current Csp and Category classes, if the passed in value is not supported, serialization will return null.
Throw this exception during serialization when an unsupported value is passed in.

iskey:
maybe a generic exception for enumerate is a better choice.

jinyangyang222:
> maybe a generic exception for enumerate is a better choice.

ok, I will fix it.

PR #670, opened 2023-07-28
fix deserialization bugs for unsupported enumerate values

jinyangyang222:
related to #667

### DIFF
--- a/modules/models/src/main/java/org/eclipse/xpanse/modules/models/common/exceptions/UnsupportedEnumValueException.java
+++ b/modules/models/src/main/java/org/eclipse/xpanse/modules/models/common/exceptions/UnsupportedEnumValueException.java
@@ -1,0 +1,16 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * SPDX-FileCopyrightText: Huawei Inc.
+ */
+
+package org.eclipse.xpanse.modules.models.common.exceptions;
+
+/**
+ * Exception thrown when there is an error during enum deserialization.
+ */
+public class UnsupportedEnumValueException extends RuntimeException {
+
+    public UnsupportedEnumValueException(String message) {
+        super(message);
+    }
+}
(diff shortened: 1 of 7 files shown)


jinyangyang222:
@swaroopar Hi swaroopar, Please review this, if this small modification is ok, I will fix all enums and tests.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/common/exceptions/UnsupportedEnumValueException.java
```diff
@@ -0,0 +1,16 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * SPDX-FileCopyrightText: Huawei Inc.
+ */
+
+package org.eclipse.xpanse.modules.models.common.exceptions;
+
+/**
+ * Exception thrown when there is an error during enum serialization.
```
swaroopar, line 9: Should be deseralization.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/common/exceptions/UnsupportedEnumValueException.java
```diff
@@ -0,0 +1,16 @@
...
+ */
+public class UnsupportedEnumValueException extends RuntimeException {
+
+    public UnsupportedEnumValueException(String message) {
+        super("Enums serialization error:" + message);
```
swaroopar, line 14: No need of this extra text. Just pass the ''message" string.

swaroopar (changes requested):
@jinyangyang222 looks good. Just minor comment added. We must still add exception handler for this new exception. Either with this PR or the next PRs for the same topic.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/common/exceptions/UnsupportedEnumValueException.java
jinyangyang222: ok.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/common/exceptions/UnsupportedEnumValueException.java
jinyangyang222: okay.

jinyangyang222:
> @jinyangyang222 looks good. Just minor comment added. We must still add exception handler for this new exception. Either with this PR or the next PRs for the same topic.

okay,exception handler added.

swaroopar (approved):
LGTM

PR #672, opened 2023-07-28
fix deserialization bugs for unsupported enumerate values second time

### LINKED ISSUE #667: Csp  Category-- Throw an exception for unsupported values during serialization

jinyangyang222:
fixed #667


swaroopar (approved):
LGTM

PR #682, opened 2023-08-01
fix bug in get credential during deployment

### LINKED ISSUE #676: credentials in deployed variables are masked
Deployment is using wrong method to fetch credentials. This is loading masked sensitive variables. 

[org/eclipse/xpanse/modules/deployment/utils/DeployEnvironments.java:150](https://github.com/eclipse-xpanse/xpanse/blob/ad65e6de5a1e77d42fb13652a2d93f5055a9932b/modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/utils/DeployEnvironments.java#L150)

the correct method would be this https://github.com/eclipse-xpanse/xpanse/blob/ad65e6de5a1e77d42fb13652a2d93f5055a9932b/modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/CredentialCenter.java#L231 which loads the variables from all sources.

jinyangyang222:
fixed #676


swaroopar (approved):
LGTM

PR #687, opened 2023-08-01
fix POST API return 403 when Unauthorized

### LINKED ISSUE #679: API returns HTTP 403 even when no authentication header is not passed
When authentication header is not passed or if it is invalid, we must return HTTP 401 unauthenticated. 
Only when the RBAC check fails, we must return HTTP 403 unauthorized.

baixinsui:
fixed #679


baixinsui:
chrome_hGkCDiGszV

swaroopar (approved):
LGTM

PR #723, opened 2023-08-09
remove dependency on openapi-generator-cli-version.jar

### LINKED ISSUE #705: Remove dependency on openapi-generator-cli-version.jar
Currently, we download the  `openapi-generator-cli` jar during the runtime if it is done found or in case of docker image, we build an image by copying the jar into it. We must remove this dependency so that we can build a plain java docker image. 

- [ ] As part of the maven build, add a step that downloads the jar file to resources folder. Eg., https://stackoverflow.com/questions/2741806/maven-downloading-files-from-url
- [ ] At the runtime, the file is only read from the `resources` directly and copied outside the jar if the file does not exist. 
- [ ] If this works, remove the dependency in the docker file too.

jinyangyang222:
fixes #705 

Now Windows, Linux, and Docker are all working.
[
08 11-windows
](url)
[
08 11-linux
](url)
[
08 11-docker
](url)


modules/common/src/main/java/org/eclipse/xpanse/common/openapi/OpenApiUtil.java
```diff
@@ -129,6 +130,29 @@ public boolean downloadClientJar(String workdir) throws IOException {
...
+     */
+    public File getClientJarFromTargetPath() {
+        String jarFileName = "openapi-generator-cli.jar";
+        String rootPath = System.getProperty("user.dir");
+        int modulesIndex = rootPath.indexOf("modules");
+        if (modulesIndex > 1) {
+            rootPath = rootPath.substring(0, modulesIndex);
+        }
+        String filePath;
+        if (rootPath.contains("runtime") && rootPath.contains("target")) {
```
swaroopar, line 146: just confused about this implementation. 
If the openapi-generator-cli.jar is inside the final run-time jar.. then there is no need to check in target folder and the target folder does not even exist when the jar is compiled. 

What you must check is.. copy the runtime jar file to some other location.. and see if the runtime jar automatically uses the file from the "resources" inside jar.

swaroopar (changes requested):
@jinyangyang222 this is is not the exact expected solution. Please see my comments.

modules/common/src/main/java/org/eclipse/xpanse/common/openapi/OpenApiUtil.java
jinyangyang222: @swaroopar  Because we need to consider Windows/Linux and testing classes, the rootPaths they obtain here are different

swaroopar (approved):
LGTM

PR #735, opened 2023-08-11
Add ROLE_ISV and replace the usage of the ROLE_CSP to ROLE_ISV.

### LINKED ISSUE #713: API RBAC - Rename role from CSP to ISV
Rename RBAC role from CSP to ISV.
idToken will also now return role as ISV. hence change the RBAC validations accordingly.
update documentation.

iskey:
fix #713


swaroopar (approved):
LGTM

PR #750, opened 2023-08-14
Disable jacoco by default,Enable only in CI pipeline

### LINKED ISSUE #740: disable jacoco by default. Enable only in CI pipeline
Currently jacoco plugin runs always when maven build is executed. This is not necessary. 

- [ ] Introduce a flag to disable the test coverage by default. 
- [ ] Enable this only in the ci pipeline here https://github.com/eclipse-xpanse/xpanse/blob/main/.github/workflows/coverage.yml#L27

jinyangyang222:
fixes #740

### DIFF
--- a/runtime/pom.xml
+++ b/runtime/pom.xml
@@ -18,6 +18,7 @@
         <relativePath>../pom.xml</relativePath>
     </parent>
     <properties>
+        <jacoco.skip>true</jacoco.skip>
         <docker.skip>true</docker.skip>
         <image.name>xpanse</image.name>
     </properties>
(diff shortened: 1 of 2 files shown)


runtime/pom.xml
```diff
@@ -228,7 +228,8 @@
                 </executions>
                 <configuration>
                     <outputDirectory>${project.parent.basedir}/target/coverage-reports</outputDirectory>
-                    <dataFile>${project.parent.basedir}/jacoco.exec</dataFile>
+                    <dataFile>${project.parent.basedir}/target/coverage-reports/jacoco.exec</dataFile>
+                    <skip>true</skip>
```
swaroopar, line 232: hi @jinyangyang222 instead of directly using the "skip" flag.. can you please link it to a maven property?
Similar to our current flag `docker.skip`. Then it is more clear in the Maven command, what exactly we are skipping or not skipping.

jinyangyang222: fixed

swaroopar (approved):
LGTM

PR #756, opened 2023-08-16
Delete workspace of task after Saving data successfully

### LINKED ISSUE #754: Delete the Terraform workspace after successfully saving the data to the DB.
Delete the file and then save the data to the database. In case of abnormal database connection, the data cannot be saved and the original file is also deleted, making it impossible to retrieve the data for manual recovery. So we should do delete the workspace after saving the data to the DB.

baixinsui:
fixed #754


swaroopar (approved):
LGTM

PR #795, opened 2023-08-24
reduce xpanse docker image size

### LINKED ISSUE #781: reduce size of xpanse docker image
-currently the terraform-boot docker image uses the eclipse-temurin:17 image which is too big.

- [ ]  Change this to use the alpine and also only the "jre" installed base images.
- [ ] Remove the fixed aes_sec file in the image and instead create the private key at the start of the container. https://stackoverflow.com/questions/42857897/execute-a-script-before-cmd
- [ ] Remove references of maven `jib` plugin in the POM xmls. This is not used anymore.

jinyangyang222:
fixes #781 

[
08 24-docker_private_key
](url)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,27 +1,30 @@
-FROM eclipse-temurin:17
+FROM eclipse-temurin:17-jre-alpine
 
 ENV TERRAFORM_VERSION=1.4.4
 ENV XPANSE_USER=xpanse
 ENV XPANSE_HOME_DIR=/home/xpanse
 
 # Install dependencies
-RUN apt-get update && \
-    apt-get install unzip wget -y
+RUN apk update && \
+    apk add --no-cache unzip wget openssh-keygen openssl
 
 # Install Terraform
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,27 +1,29 @@
...
 USER ${XPANSE_USER}
 WORKDIR ${XPANSE_HOME_DIR}
 
 # Copy jar file
 COPY --chown=${XPANSE_USER}:${XPANSE_USER} runtime/target/xpanse-runtime-*.jar xpanse-runtime.jar
 
-# Generate random key
-RUN tr -dc 'a-zA-Z0-9' < /dev/urandom | head -c 32 > aes_sec
+# Generate RSA private key
+RUN openssl genpkey -algorithm RSA -out /home/xpanse/private_key.pem
```
swaroopar, line 26: @jinyangyang222 this will still create the key in the image. 
We need something that will create the key only when the container i started.

swaroopar (changes requested):
comments added.

swaroopar (approved):
@jinyangyang222 this looks good. I will merge it. Will test it on our testbed and get back to you in case of any questions.

PR #796, opened 2023-08-24
Automatically generate terraform-boot API client

WangLiNaruto:
Automatically generate terraform-boot API client


swaroopar (changes requested):
@WangLiNaruto i'm a little confused now. Where are we exactly using the client which YangYang generated in this?

modules/api/src/main/java/org/eclipse/xpanse/api/ServiceDeployV2Api.java
```diff
@@ -0,0 +1,102 @@
...
+
+/**
+ * REST interface methods for processing Service.
+ */
+@Slf4j
+@RestController
+@Profile("remote-deploy")
+@RequestMapping("/xpanse")
+@CrossOrigin
+public class ServiceDeployV2Api {
```
swaroopar, line 42: we need to use the same existing APIs. and based on the profile.. we must be able to switch between different terraform deployment. 
Please note that even this profile based selection is also only temporary. Once we have terraform-boot fully working, we will move to that completely and delete all terraform execution within xpanse runtime. 
So its fine even if you can think of any temporary/manual solution to start the xpanse runtime to use terraform-boot for deployment

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/deployers/terraform/request/TerraformDeployRequest.java
```diff
@@ -0,0 +1,52 @@
...
+import jakarta.validation.constraints.NotNull;
+import java.util.HashMap;
+import java.util.Map;
+import lombok.Data;
+import org.eclipse.xpanse.modules.models.service.common.enums.Csp;
+
+/**
+ * Data model for the terraform deploy requests.
+ */
+@Data
```
swaroopar, line 18: i dont think we need these data classes

WangLiNaruto:
> i'm a little confused now. Where are we exactly using the client which YangYang generated in this?

This is the code that generates the terraform-boot client. I have not yet submitted the code used in the terraform-boot client code. I will submit that code after waiting for the PR to be merged.

swaroopar (approved):
LGTM

PR #798, opened 2023-08-24
add user friendly enum string values

swaroopar:
changing enum string values so that it can be directly used in UI/API


jinyangyang222 (approved):
looks good

iskey (approved):
LGTM

PR #815, opened 2023-08-30
add terraformBoot healthCheck to xpanse healthCheck background

### LINKED ISSUE #774: add terraform-boot health check to xpanse healthcheck background
- [ ] use the /health method of the terraform-boot to check if terraform boot is API is reachable.
- [ ] Add terraform-boot backend to the current list of backends that checked in Xpanse's healthcheck API

jinyangyang222:
fixes #774

### DIFF
--- a/modules/api/src/main/java/org/eclipse/xpanse/api/AdminServicesApi.java
+++ b/modules/api/src/main/java/org/eclipse/xpanse/api/AdminServicesApi.java
@@ -18,6 +18,9 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Objects;
 import lombok.extern.slf4j.Slf4j;
+import org.eclipse.xpanse.modules.deployment.deployers.terraform.terraformboot.ApiClient;
+import org.eclipse.xpanse.modules.deployment.deployers.terraform.terraformboot.api.TerraformApi;
+import org.eclipse.xpanse.modules.deployment.deployers.terraform.terraformboot.model.TerraformBootSystemStatus;
 import org.eclipse.xpanse.modules.models.security.model.CurrentUserInfo;
 import org.eclipse.xpanse.modules.models.system.BackendSystemStatus;
 import org.eclipse.xpanse.modules.models.system.SystemStatus;
@@ -51,6 +54,12 @@ public class AdminServicesApi {
 
     @Value("${spring.datasource.url:jdbc:h2:file:./testdb}")
     private String dataSourceUrl;
+
+    @Resource
+    private TerraformApi terraformApi;
+
+    @Resource
+    private ApiClient apiClient;
 
     /**
      * Method to find out the current state of the system.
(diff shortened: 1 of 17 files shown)


modules/api/src/main/java/org/eclipse/xpanse/api/AdminServicesApi.java
```diff
@@ -46,6 +49,8 @@
 @Secured({ROLE_ADMIN})
 public class AdminServicesApi {
 
+    private final TerraformApi terraformApi = new TerraformApi();
```
swaroopar, line 52: @jinyangyang222 I see there is some issue with the generated classes. The generated classes must be spring beans and simply be `autowired`. 
<generatorName>java</generatorName> -> must be `spring` for the classes to be generated as spring beans.

Can you please check this?

modules/api/src/main/java/org/eclipse/xpanse/api/AdminServicesApi.java
```diff
@@ -89,10 +94,39 @@ private List<BackendSystemStatus> checkHealthOfBackendSystem() {
...
+    }
+
+    private BackendSystemStatus getTerraformBootStatus() {
+        BackendSystemStatus terraformBootStatus = new BackendSystemStatus();
+        terraformBootStatus.setBackendSystemType(BackendSystemType.TERRAFORM_BOOT);
+        terraformBootStatus.setName(BackendSystemType.TERRAFORM_BOOT.toValue());
+        if (isTerraformBootApiAccessible()) {
+            terraformBootStatus.setHealthStatus(HealthStatus.OK);
+            ApiClient apiClient = new ApiClient();
+            terraformBootStatus.setEndpoint(apiClient.getBasePath());
```
swaroopar, line 123: we must also customize the base path config. 

The basepath must be read from spring-properties and then added to the ApiClient. 

The generated class has the default url and this must not be directly used.

swaroopar (changes requested):
comments added.

modules/api/src/main/java/org/eclipse/xpanse/api/AdminServicesApi.java
jinyangyang222: ok,I'll check this.

modules/api/src/main/java/org/eclipse/xpanse/api/AdminServicesApi.java
```diff
@@ -46,12 +48,27 @@
...
+    private final TerraformApi terraformApi;
+
+    private final ApiClient apiClient;
+
+    /**
+     * The constructor.
+     */
+    public AdminServicesApi(IdentityProviderManager identityProviderManager,
+                            TerraformApi terraformApi, ApiClient apiClient,
+                            @Value("${apiClient.basePath:}") String baseUrl) {
```
swaroopar, line 65: @jinyangyang222 we must not update the basepath here. Instead we need to update the bean directly at startup and simply use it everywhere, 
example - https://github.com/Zomzog/openapi-spring-client-token/blob/master/front/src/main/java/bzh/zomzog/demo/openapispringclienttoken/front/config/BackendClientConfig.java

runtime/src/main/resources/application-terraformboot-local.properties
```diff
@@ -0,0 +1,6 @@
+#
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Huawei Inc.
+#
+#
+apiClient.basePath=http://localhost:9090
```
swaroopar, line 6: property name - terraformboot.endpoint

runtime/src/main/resources/application-terraformboot-local.properties
```diff
@@ -0,0 +1,6 @@
+#
```
swaroopar, line 1: we can just have one profile - application-terraformboot.properties and enable it by default. 
Pls also check once with WangLi.. i think he also has added a new profile in this PR.

swaroopar (changes requested):
comments added

modules/api/src/main/java/org/eclipse/xpanse/api/AdminServicesApi.java
jinyangyang222: ok,fixed

swaroopar (approved):
LGTM

PR #819, opened 2023-08-31
disable generating timestamp and exclude terraform-boot from code coverage

### LINKED ISSUE #810: exclude terraform-boot generate client code from code coverage
We must exclude the terraform-boot generate client code from code coverage report. 
Currently it is included and this is causing the report to show less coverage.

### LINKED ISSUE #814: terraform-boot client - disable generating time stamp
Currently all terraform-boot client generated files have a "Generated" annotation with date. This is causing all files to be updated whenever the client is newly generated. 

Check for options to disable this timestamp recorded in the classes.
Hint - https://github.com/swagger-api/swagger-codegen/issues/2834

jinyangyang222:
fixes #814 
fixes #810


modules/deployment/src/main/resources/terraformbootApi.json
```diff
@@ -35,8 +35,8 @@
           "required": true
         },
         "responses": {
-          "400": {
```
iskey, line 38: it seems only the order of some contents has changed in this file, can you remove this from this PR?

modules/deployment/pom.xml
```diff
@@ -85,7 +85,7 @@
             <plugin>
                 <groupId>org.openapitools</groupId>
                 <artifactId>openapi-generator-maven-plugin</artifactId>
-                <version>6.6.0</version>
+                <version>6.4.0</version>
```
iskey, line 88: why this change?

modules/common/src/main/java/org/eclipse/xpanse/common/openapi/config/RestTemplateConfig.java
```diff
@@ -0,0 +1,23 @@
...
+ */
+
+package org.eclipse.xpanse.common.openapi.config;
+
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.web.client.RestTemplate;
+
+/**
+ * Configuration class to define a RestTemplate bean.
```
iskey, line 14:
```suggestion
 * Configuration class to create a RestTemplate bean.
```

iskey (changes requested):
comments added

modules/deployment/pom.xml
jinyangyang222: The classes generated in version 6.6.0 do not have @ Component annotations, only in version 6.4.0. The client we need to generate is a spring bean.

modules/deployment/src/main/resources/terraformbootApi.json
jinyangyang222: This file was copied from the swagger api-docs of the terraform-boot. If the terraform boot changes, it must also be changed here.

modules/deployment/pom.xml
swaroopar: @jinyangyang222 can you keep the 6.6.0 version and add "generateClientAsBean" config as true? Pls see if this works.

jinyangyang222: @iskey fixed， @swaroopar thank you very much, it's ok.

iskey (approved):
LGTM

PR #821, opened 2023-09-01
Get metadata of user and Add namespace in ServiceTemplate

baixinsui:
fixed https://github.com/eclipse-xpanse/xpanse/issues/792
related https://github.com/eclipse-xpanse/xpanse-ui/pull/183


swaroopar (approved):
LGTM

PR #827, opened 2023-09-04
fix monitor for HuaweiCloud and FlexibleEngine

### LINKED ISSUE #800: Metrics for HuaweiCloud need to be updated
# Metrics for HuaweiCloud need to be updated

image

image

jinyangyang222:
fixes #800

### DIFF
--- a/plugins/flexibleengine/src/main/java/org/eclipse/xpanse/plugins/flexibleengine/monitor/utils/FlexibleEngineMetricsService.java
+++ b/plugins/flexibleengine/src/main/java/org/eclipse/xpanse/plugins/flexibleengine/monitor/utils/FlexibleEngineMetricsService.java
@@ -29,6 +29,7 @@ import org.eclipse.xpanse.modules.models.credential.enums.CredentialType;
 import org.eclipse.xpanse.modules.models.monitor.Metric;
 import org.eclipse.xpanse.modules.models.monitor.enums.MonitorResourceType;
 import org.eclipse.xpanse.modules.models.monitor.exceptions.ClientApiCallFailedException;
+import org.eclipse.xpanse.modules.models.monitor.exceptions.MetricsDataNotYetAvailableException;
 import org.eclipse.xpanse.modules.models.service.common.enums.Csp;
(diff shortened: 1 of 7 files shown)


plugins/flexibleengine/src/main/java/org/eclipse/xpanse/plugins/flexibleengine/monitor/utils/FlexibleEngineMetricsService.java
```diff
@@ -339,9 +339,16 @@ private Map<MonitorResourceType, MetricInfoList> getTargetMetricsMap(
...
-        }
-        return targetMetricsMap;
+            return targetMetricsMap;
 
+        } else {
+            log.error(
+                    "No monitor metrics available for the service, "
+                            + "Please wait for 3-5 minutes and try again.");
+            throw new ClientApiCallFailedException(
+                    "No monitor metrics available for the  service, "
```
swaroopar, line 349: looks like there is an extra space before the word "service". Same in both places.

plugins/flexibleengine/src/main/java/org/eclipse/xpanse/plugins/flexibleengine/monitor/utils/FlexibleEngineMetricsService.java
```diff
@@ -339,9 +339,16 @@ private Map<MonitorResourceType, MetricInfoList> getTargetMetricsMap(
...
             }
-        }
-        return targetMetricsMap;
+            return targetMetricsMap;
 
+        } else {
+            log.error(
+                    "No monitor metrics available for the service, "
+                            + "Please wait for 3-5 minutes and try again.");
+            throw new ClientApiCallFailedException(
```
swaroopar, line 348: ClientApiCallFailedException -> will return HTTP code 502 to client which is not correct. 
If we really need an exception to be returned here, then we must define something new that will return HTTP 400.

swaroopar (changes requested):
pls check the exception type.

swaroopar (approved):
LGTM

PR #833, opened 2023-09-05
Add API to query Csp list

baixinsui:
fixed https://github.com/eclipse-xpanse/xpanse/issues/830
related https://github.com/eclipse-xpanse/xpanse-ui/pull/186

### DIFF
--- a/modules/api/src/main/java/org/eclipse/xpanse/api/AdminServicesApi.java
+++ b/modules/api/src/main/java/org/eclipse/xpanse/api/AdminServicesApi.java
@@ -11,22 +11,26 @@ import static org.eclipse.xpanse.modules.models.security.constant.RoleConstants.
 import static org.eclipse.xpanse.modules.models.security.constant.RoleConstants.ROLE_USER;
 
 import io.swagger.v3.oas.annotations.Operation;
+import io.swagger.v3.oas.annotations.Parameter;
 import io.swagger.v3.oas.annotations.tags.Tag;
 import jakarta.annotation.Resource;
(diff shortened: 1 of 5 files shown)


modules/api/src/main/java/org/eclipse/xpanse/api/AdminServicesApi.java
```diff
@@ -81,6 +89,29 @@ public SystemStatus healthCheck() {
...
 
+
+    /**
+     * List supported cloud service provider.
+     *
+     * @return Returns list of cloud service provider.
+     */
+    @Tag(name = "Admin", description = "APIs for administrating Xpanse")
+    @Operation(description = "List cloud service provider.")
+    @GetMapping(value = "/csps", produces = MediaType.APPLICATION_JSON_VALUE)
```
swaroopar, line 100: @baixinsui URI is better to be just /csp

modules/api/src/main/java/org/eclipse/xpanse/api/AdminServicesApi.java
```diff
@@ -81,6 +89,29 @@ public SystemStatus healthCheck() {
...
+    @ResponseStatus(HttpStatus.OK)
+    @Secured({ROLE_ADMIN, ROLE_ISV, ROLE_USER})
+    public List<Csp> listCsps(
```
swaroopar, line 103: method name can be `getCsps`

swaroopar (changes requested):

swaroopar (approved):
LGTM

PR #843, opened 2023-09-06
remove port hard coding in the webhook url

### LINKED ISSUE #837: remove port hard coding in the webhook url
The port from the default config must be removed. We must read the port value from the server.port property. 

https://github.com/eclipse-xpanse/xpanse/blob/668cdbe2bc7fba0f83be4be2875c17f7dd7d2551/modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/deployers/terraform/config/TerraformBootConfig.java#L43

WangLiNaruto:
Fixed #837
image


swaroopar (approved):
LGTM

PR #852, opened 2023-09-07
update property name terraform-boot.endpoint

### LINKED ISSUE #841: update property name terraform-boot.endpoint
The current property name is not uniform. 
https://github.com/eclipse-xpanse/xpanse/blob/668cdbe2bc7fba0f83be4be2875c17f7dd7d2551/runtime/src/main/resources/application-terraform-boot.properties#L10

Make it `terraform.boot.endpoint` and update docs if necessary.

jinyangyang222:
fixes #841


swaroopar (approved):
LGTM

PR #854, opened 2023-09-07
Enable terraform-boot backed in health conditionally

### LINKED ISSUE #848: Enable terraform-boot backed in health conditionally
currently we consider terraform-boot as a backend in xpanse health check always. This must be changed. 
This backend must be considered only when `terraform-boot` profile is active. 
Check the list of active profiles and based on that decide if the terraform-boot health must be checked.

jinyangyang222:
fixes #848


swaroopar (approved):
LGTM

PR #855, opened 2023-09-07
exception handling for errors returned from terraform boot API

### LINKED ISSUE #840: exception handling for errors returned from terraform boot API
For errors returned from terraform boot API, currently the exception is not handled at all.
Exceptions of type `Response` must be caught and we must throw `TerraformBootRequestFailedException` with the error message coming from terraformboot `Response` object.

### LINKED ISSUE #847: terraform-boot does not call webhook in case of failure
Currently in case of deployment/destroy failures, terraform-boot is throwing exceptions. Instead all exceptions must be caught and the exception details must be returned in the webbook. 
This is currently causing the UI to assume that the deployment is still progress..

WangLiNaruto:
Fixed #840
Fixed #847
image


swaroopar:
@WangLiNaruto something is not fine with the screenshot. If you look at the terraform-local implementation, we will also show the exact error coming from terraform. We need that same behavior with terraform-boot too.

swaroopar (approved):
LGTM

PR #861, opened 2023-09-08
update resttemplate logging format using interceptor

### LINKED ISSUE #844: update resttemplate logging format using interceptor
Currently the rest template is using the default logging which is not optimal and does not match to our other log lines. Instead we must provide a custom log interceptor.

- [ ] Add a custom log interceptor for both terraform boot and flexible engine Rest template. The log lines must look similar to how we log our own HTTP request and response. Just one line for request, one for response. 
Example - https://objectpartners.com/2018/03/01/log-your-resttemplate-request-and-response-without-destroying-the-body/
- [ ] Remove this existing config https://github.com/eclipse-xpanse/xpanse/blob/668cdbe2bc7fba0f83be4be2875c17f7dd7d2551/runtime/src/main/resources/application.properties#L24
- [ ] Log request and response details with 'info' level. 
- [ ] Ensure no logging is done by the 'RestTemplate' package. We log only from our interceptor classes. 
- [ ] We can also define one interceptor and use it in both terraform-boot and flexible engine client.

jinyangyang222:
fixes #844 

[
09-11-log
](url)


modules/logging/src/main/java/org/eclipse/xpanse/modules/logging/config/RestTemplateConfig.java
```diff
@@ -0,0 +1,39 @@
...
+import org.eclipse.xpanse.modules.logging.RestTemplateLoggingInterceptor;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.web.client.RestTemplate;
+
+/**
+ * Configuration class to create a RestTemplate bean.
+ */
+@Configuration
+public class RestTemplateConfig {
```
swaroopar, line 19: It is better to have dedicated RestTemplate beans for each of our REST clients. This way, we can control it better. 
We need to have one in terraformboot, one for ZitadelIdentityProviderService and one for flexibleEngine.

in fact, the terraformboot is using its own version of RestTemplate - https://github.com/eclipse-xpanse/xpanse/blob/dae90cdf53fb8979a0f23643291a6cca84408702/modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/deployers/terraform/terraformboot/ApiClient.java#L644

We must update the interceptor directly  on the ApiClient bean. 

We must remove the RestTemplateConfig class and add the interceptors on each of the RestTemplates which we are creating.

swaroopar (changes requested):
pls also add screenshots of how the log looks like now. It will be easy to review the PR.

modules/logging/src/main/java/org/eclipse/xpanse/modules/logging/config/RestTemplateConfig.java
jinyangyang222: updated.

swaroopar (approved):
@jinyangyang222 great work. Looks good. Merging.

PR #873, opened 2023-09-12
Update variable validator format to follow JSONSchema/openapi specifications

### LINKED ISSUE #866: Update variable validator format to follow JSONSchema/openapi specifications
Currently the validator for a deployment variable is configured as  pipe separated string which is not very straight forward. Also the `VariableValidator` is not exposed at all. 

https://github.com/eclipse-xpanse/xpanse/blob/ec910cd39b8c5fbcafa124e196bcb41b66b4e555/modules/models/src/main/java/org/eclipse/xpanse/modules/models/servicetemplate/DeployVariable.java#L52

- [ ] Make validator a map of <VariableValidator, String>.
- [ ] Rename `validator` field in `DeployVariable` to `valueSchema` and document the field as detail as possible.
- [ ] Make the `VariableValidator` enum have values same as the JSON values-> https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-validation-00 -> we can use the values in OpenApi schema object which is a superset of JSON schema but consider the next point. Validating the field will not be straightforward. 
- [ ] Consider using any existing JSON validation libraries for data validation instead of custom validation - https://github.com/networknt/json-schema-validator

P.S - Once this is implemented, we will also have the similar OpenApi validation on UI for the deploy variables.

jinyangyang222:
fixes #866 

when deployVariable validator fail, error like this:
Screenshot_15


modules/models/src/main/java/org/eclipse/xpanse/modules/models/service/utils/DeployVariableValidator.java
```diff
@@ -204,129 +129,65 @@ public Map<VariableValidator, String> getValidatorMap(String variableName, Strin
...
-            }
-            // check input value of variables is valid by validator of Ocl context.
-            Map<String, Map<VariableValidator, String>> variableInfoMap = getVariableValidatorMap(
-                    deployVariables);
-            if (variableInfoMap.isEmpty() || userPutKeys.isEmpty()) {
+        JsonSchemaObject jsonObjectSchema = buildJsonObjectSchema(deployVariables);
+
+        try {
+            String schemaJson = jsonMapper.writeValueAsString(jsonObjectSchema);
+            JsonSchemaFactory factory = JsonSchemaFactory.getInstance(SpecVersion.VersionFlag.V4);
```
swaroopar, line 140: we can use the latest version.

swaroopar (changes requested):
@jinyangyang222 you are definitely in the right direction. Exactly what we are looking for. I have added comments.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/service/utils/DeployVariableValidator.java
```diff
@@ -204,129 +129,65 @@ public Map<VariableValidator, String> getValidatorMap(String variableName, Strin
...
-            if (!ignoredKeys.isEmpty()) {
-                throw new InvalidDeploymentVariableException(
-                        String.format("Required keys %s  of deploy "
-                                + " variables not found in deploy property", ignoredKeys));
-            }
-            // check input value of variables is valid by validator of Ocl context.
-            Map<String, Map<VariableValidator, String>> variableInfoMap = getVariableValidatorMap(
-                    deployVariables);
-            if (variableInfoMap.isEmpty() || userPutKeys.isEmpty()) {
+        JsonSchemaObject jsonObjectSchema = buildJsonObjectSchema(deployVariables);
```
swaroopar, line 136: write a new spring bean -> ServiceVariablesJsonSchemaGenerator and move the logic there.

modules/models/pom.xml
```diff
@@ -91,5 +91,16 @@
             <version>3.1.2</version>
             <scope>compile</scope>
         </dependency>
+        <dependency>
+            <groupId>com.networknt</groupId>
+            <artifactId>json-schema-validator</artifactId>
+            <version>1.0.87</version>
```
swaroopar, line 97: move the version to root pom.

swaroopar (changes requested):

modules/models/src/main/java/org/eclipse/xpanse/modules/models/service/utils/DeployVariableValidator.java
```diff
@@ -204,129 +129,65 @@ public Map<VariableValidator, String> getValidatorMap(String variableName, Strin
...
-                            if (userLong > validatorLong) {
-                                isValid = false;
-                            }
-                        }
-                    }
-                } else {
-                    isValid = false;
-                }
-            }
+    private JsonSchemaObject buildJsonObjectSchema(List<DeployVariable> deployVariables) {
```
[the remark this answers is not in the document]
jinyangyang222: Implemented, pls check.

modules/database/src/main/java/org/eclipse/xpanse/modules/database/servicetemplate/JsonObjectSchemaEntity.java
```diff
@@ -0,0 +1,57 @@
...
+import lombok.Data;
+import org.eclipse.xpanse.modules.database.common.CreateModifiedTime;
+import org.hibernate.annotations.Type;
+
+/**
+ * Represents the JSON_OBJECT_SCHEMA table in the database.
+ */
+@Entity
+@Table(name = "JSON_OBJECT_SCHEMA")
+@Data
```
swaroopar, line 31: @jinyangyang222 we do not need a new table like this. We must just serialize the JSON schema object and write to the existing service entity table  into a new column.

swaroopar (changes requested):

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java
```diff
@@ -566,4 +571,30 @@ private void maskSensitiveFields(DeployServiceEntity deployServiceEntity) {
             }
         }
     }
+
+    private JsonObjectSchema convertJsonObjectSchemaEntityToJsonObjectSchema(JsonObjectSchemaEntity entity) {
```
swaroopar, line 575: we just need to generate the JSON schema object as a JSON string and store in DB.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/service/utils/ServiceVariablesJsonSchemaGenerator.java
```diff
@@ -0,0 +1,111 @@
...
+    public boolean isVariableValid(List<DeployVariable> deployVariables,
+                                   Map<String, String> deployProperty,
+                                   JsonObjectSchema jsonObjectSchema) {
+        if (CollectionUtils.isEmpty(deployVariables) || CollectionUtils.isEmpty(deployProperty)) {
+            return true;
+        }
+
+        try {
+            String schemaJson = jsonMapper.writeValueAsString(jsonObjectSchema);
+            JsonSchemaFactory factory = JsonSchemaFactory.getInstance(SpecVersion.VersionFlag.V4);
```
swaroopar, line 54: is V4 the latest version?

modules/models/src/main/java/org/eclipse/xpanse/modules/models/servicetemplate/utils/JsonObjectSchema.java
```diff
@@ -0,0 +1,24 @@
...
+import java.util.Map;
+import lombok.Data;
+
+/**
+ * The class is used to describe the structure and validation rules of JSON data.
+ */
+@Data
+public class JsonObjectSchema {
+
+    private final String type = "object";
```
swaroopar, line 18: this can be a static field.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/service/utils/ServiceVariablesJsonSchemaGenerator.java
```diff
@@ -0,0 +1,111 @@
...
+    private final ObjectMapper jsonMapper = new ObjectMapper();
+
+    /**
+     * Check validation of deploy property map by list of deployVariables in registered service.
+     *
+     * @param deployVariables list of deployVariables in registered service
+     * @param deployProperty  deploy property map
+     * @return isValid
+     */
+    public boolean isVariableValid(List<DeployVariable> deployVariables,
```
swaroopar, line 45: here we never return false. In case of failures, we return exceptions. So better to change return type to void and method name to validateDeployVariables

modules/models/src/main/java/org/eclipse/xpanse/modules/models/service/utils/ServiceVariablesJsonSchemaGenerator.java
```diff
@@ -0,0 +1,111 @@
...
+import lombok.extern.slf4j.Slf4j;
+import org.eclipse.xpanse.modules.models.common.exceptions.VariableInvalidException;
+import org.eclipse.xpanse.modules.models.servicetemplate.DeployVariable;
+import org.eclipse.xpanse.modules.models.servicetemplate.utils.JsonObjectSchema;
+import org.springframework.stereotype.Component;
+import org.springframework.util.CollectionUtils;
+
+@Slf4j
+@Component
+public class ServiceVariablesJsonSchemaGenerator {
```
swaroopar, line 31: this bean handles both generation and valiation. we can split it into two beans or call the class name something generic.

swaroopar (changes requested):

modules/database/src/main/java/org/eclipse/xpanse/modules/database/servicetemplate/ServiceTemplateEntity.java
```diff
@@ -69,4 +69,10 @@ public class ServiceTemplateEntity extends CreateModifiedTime {
     @Column(name = "SERVICE_REGISTRATION_STATE")
     @Enumerated(EnumType.STRING)
     private ServiceRegistrationState serviceRegistrationState;
+
+    @Column(name = "JSON_OBJECT_SCHEMA_StRING", columnDefinition = "json", nullable = false)
```
swaroopar, line 73: column name -> SERVICE_VARIABLES_JSON_SCHEMA

modules/database/src/main/java/org/eclipse/xpanse/modules/database/servicetemplate/ServiceTemplateEntity.java
```diff
@@ -69,4 +69,10 @@ public class ServiceTemplateEntity extends CreateModifiedTime {
...
+    @Type(value = JsonType.class)
+    @Convert(converter = ObjectJsonConverter.class)
+    private String jsonObjectSchemaString;
```
swaroopar, line 76: please see if we can use JsonObjectSchema type instead of string. 
Also see in the h2-console, how the data is stored. Should be same as OCL column.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/service/utils/ServiceVariablesJsonSchemaValidator.java
```diff
@@ -0,0 +1,65 @@
...
+     *
+     * @param deployVariables list of deployVariables in registered service
+     * @param deployProperty  deploy property map
+     */
+    public void validateDeployVariables(List<DeployVariable> deployVariables,
+                                        Map<String, String> deployProperty,
+                                        String jsonObjectSchemaStr) {
+        if (!CollectionUtils.isEmpty(deployVariables) && !CollectionUtils.isEmpty(deployProperty)) {
+            try {
+                Locale.setDefault(Locale.ENGLISH);
```
swaroopar, line 46: this wil overwrite for the entire JVM. Instead use the 'SchemaValidatorsConfig' option in the link which i gave you.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/service/utils/ServiceVariablesJsonSchemaValidator.java
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * Check validation of deploy property map by list of deployVariables in registered service.
+     *
+     * @param deployVariables list of deployVariables in registered service
+     * @param deployProperty  deploy property map
+     */
+    public void validateDeployVariables(List<DeployVariable> deployVariables,
+                                        Map<String, String> deployProperty,
+                                        String jsonObjectSchemaStr) {
```
swaroopar, line 43: Should be the JsonObjectSchema instead of String.

modules/servicetemplate/src/main/java/org/eclipse/xpanse/modules/servicetemplate/ServiceTemplateManage.java
```diff
@@ -324,4 +341,16 @@ private void validateTerraformScript(Ocl ocl) {
         }
     }
 
+    private String generateJsonObjectSchemaString(List<DeployVariable> variables) {
```
swaroopar, line 344: Everywhere we need the JsonObjectSchema directly instead of string.

swaroopar (changes requested):

modules/servicetemplate/src/main/java/org/eclipse/xpanse/modules/servicetemplate/ServiceTemplateManage.java
jinyangyang222: The above comments have been fixed, pls check.

modules/api/src/main/java/org/eclipse/xpanse/api/exceptions/handler/CommonExceptionHandler.java
```diff
@@ -170,4 +171,16 @@ public Response handleMethodArgumentTypeMismatchException(
         return Response.errorResponse(ResultType.UNPROCESSABLE_ENTITY,
                 Collections.singletonList(ex.getMessage()));
     }
+
+    /**
```
swaroopar, line 175: please add this to DeploymentExceptionHandler  class. same for the test class also.

swaroopar (changes requested):

modules/api/src/main/java/org/eclipse/xpanse/api/exceptions/handler/CommonExceptionHandler.java
jinyangyang222: fixed.

samples/huawei-Kafka.yml
```diff
@@ -73,14 +73,17 @@ deployment:
       description: The admin password of all nodes in the Kafka cluster. If the value is empty, will create a random password.
```
swaroopar, line 73: please update all other sample files too.

swaroopar (changes requested):

swaroopar (approved):
LGTM

PR #892, opened 2023-09-15
use correct timestamp format

### LINKED ISSUE #891: issue with all timestamps stored in DB
Currently created and updated time of all entities are stored as `Date` which is just the unix timeformat and this is causing wrong times to be displayed when the timezones are different.

swaroopar:
fixes #891


iskey (approved):
LGTM

PR #911, opened 2023-09-22
validate only variables which are provided by end user

swaroopar:
fixes https://github.com/eclipse-xpanse/xpanse/issues/909


iskey (approved):
LGTM

PR #912, opened 2023-09-22
implement new migrate service using terraform-local

### LINKED ISSUE #887: implement new migrate service using terraform-local
- [ ] Implement a new REST POST method `migrate`.
- [ ] URI to have the 'id' of the service to be migrated. (same as `destroy` service)
- [ ] Request body to have the `CreateRequest` object. (same as `deploy` service)
- [ ] Extend the Service DB entity to store the original ID of the service. We must use this later for deleting the original service.
- [ ] Return the ID of the new service. (Same as the  `deploy` service)
- [ ] Terraform-local -> start a async process, run deploy first and once deploy is successful, run the destroy. We must not have any code duplication. We must simply use the existing methods.

WangLiNaruto:
Fixed #887

### DIFF
--- a/modules/api/pom.xml
+++ b/modules/api/pom.xml
@@ -83,5 +83,15 @@
             <artifactId>spring-security-test</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>io.nflow</groupId>
+            <artifactId>nflow-engine</artifactId>
+            <version>${nflow.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.xpanse.modules</groupId>
+            <artifactId>workflow</artifactId>
+            <version>${project.version}</version>
+        </dependency>
     </dependencies>
 </project>
--- a/modules/api/src/main/java/org/eclipse/xpanse/api/ServiceDeployerApi.java
+++ b/modules/api/src/main/java/org/eclipse/xpanse/api/ServiceDeployerApi.java
@@ -6,14 +6,15 @@
 
 package org.eclipse.xpanse.api;
 
-
 import static org.eclipse.xpanse.modules.models.security.constant.RoleConstants.ROLE_ADMIN;
 import static org.eclipse.xpanse.modules.models.security.constant.RoleConstants.ROLE_USER;
 
+import io.nflow.engine.service.WorkflowInstanceService;
+import io.nflow.engine.workflow.instance.WorkflowInstance;
+import io.nflow.engine.workflow.instance.WorkflowInstanceFactory;
 import io.swagger.v3.oas.annotations.Operation;
 import io.swagger.v3.oas.annotations.Parameter;
 import io.swagger.v3.oas.annotations.tags.Tag;
-import jakarta.annotation.Resource;
 import jakarta.validation.Valid;
 import java.util.Collections;
 import java.util.List;
@@ -33,6 +34,8 @@ import org.eclipse.xpanse.modules.models.service.view.ServiceDetailVo;
 import org.eclipse.xpanse.modules.models.service.view.ServiceVo;
 import org.eclipse.xpanse.modules.orchestrator.deployment.DeployTask;
 import org.eclipse.xpanse.modules.orchestrator.deployment.Deployment;
+import org.eclipse.xpanse.modules.nflow.definition.ServiceMigrationWorkflow;
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.MediaType;
 import org.springframework.security.access.annotation.Secured;
@@ -58,9 +61,20 @@ import org.springframework.web.bind.annotation.RestController;
 @Secured({ROLE_ADMIN, ROLE_USER})
 public class ServiceDeployerApi {
 
-    @Resource
-    private DeployService deployService;
+    private final DeployService deployService;
 
+    private final WorkflowInstanceService workflowInstances;
+
+    private final WorkflowInstanceFactory workflowInstanceFactory;
+
+    @Autowired
+    public ServiceDeployerApi(DeployService deployService,
+            WorkflowInstanceService workflowInstances,
+            WorkflowInstanceFactory workflowInstanceFactory) {
+        this.deployService = deployService;
+        this.workflowInstances = workflowInstances;
+        this.workflowInstanceFactory = workflowInstanceFactory;
+    }
 
     /**
      * Get details of the managed service by id.
@@ -177,6 +191,37 @@ public class ServiceDeployerApi {
         this.deployService.purgeService(deployment, deployTask);
         String successMsg = String.format("Purging task for service with ID %s has started.", id);
         return Response.successResponse(Collections.singletonList(successMsg));
+    }
+
+    /**
+     * Start a task to migrate the deployed service using id
+     *
+     * @param deployRequest the managed service to create.
+     * @param oldId            ID of deployed service
+     */
+    @Tag(name = "Service", description = "APIs to manage the service instances")
+    @Operation(description = "Start a task to migrate the deployed service using id.")
+    @PostMapping(value = "/services/migrate/{id}", produces = MediaType.APPLICATION_JSON_VALUE)
+    @ResponseStatus(HttpStatus.ACCEPTED)
+    public UUID migrate(@Valid @RequestBody CreateRequest deployRequest,
+            @PathVariable("id") String oldId) {
+        log.info("Migrate managed service with id {}", oldId);
+        UUID newId = UUID.randomUUID();
+        workflowInstances.insertWorkflowInstance(
+                getWorkflowInstance(oldId, deployRequest, newId));
+        return newId;
+    }
+
+    private WorkflowInstance getWorkflowInstance(String oldId,
+            CreateRequest deployRequest, UUID newId) {
+        return workflowInstanceFactory.newWorkflowInstanceBuilder()
+                .setType(ServiceMigrationWorkflow.TYPE)
+                .putStateVariable(ServiceMigrationWorkflow.NEW_ID, newId)
+                .putStateVariable(ServiceMigrationWorkflow.DEPLOY_REQUEST, deployRequest)
+                .putStateVariable(ServiceMigrationWorkflow.OLD_ID, oldId)
+                .setBusinessKey(newId.toString())
+                .setExternalId(newId.toString())
+                .build();
     }
 
     private ServiceQueryModel getServiceQueryModel(Category category, Csp csp,
--- a/modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java
+++ b/modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java
@@ -191,6 +191,16 @@ public class DeployService {
      */
     @Async("taskExecutor")
     public void asyncDeployService(Deployment deployment, DeployTask deployTask) {
+        deployService(deployment, deployTask);
+    }
+
+    /**
+     * Method to deploy service.
+     *
+     * @param deployment deployment
+     * @param deployTask deployTask
+     */
+    public void deployService(Deployment deployment, DeployTask deployTask) {
         MDC.put(TASK_ID, deployTask.getId().toString());
         DeployServiceEntity deployServiceEntity = getNewDeployServiceTask(deployTask);
         DeployResult deployResult = new DeployResult();
@@ -300,6 +310,16 @@ public class DeployService {
      */
     @Async("taskExecutor")
     public void asyncDestroyService(Deployment deployment, DeployTask deployTask) {
+        destroyService(deployment, deployTask);
+    }
+
+    /**
+     * Method to destroy service.
+     *
+     * @param deployment deployment
+     * @param deployTask deployTask
+     */
+    public void destroyService(Deployment deployment, DeployTask deployTask) {
         MDC.put(TASK_ID, deployTask.getId().toString());
         DeployServiceEntity deployServiceEntity =
                 deployServiceStorage.findDeployServiceById(deployTask.getId());
--- a/modules/workflow/pom.xml
+++ b/modules/workflow/pom.xml
@@ -1,0 +1,74 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <artifactId>modules</artifactId>
+    <groupId>org.eclipse.xpanse</groupId>
+    <version>1.0.0-SNAPSHOT</version>
+  </parent>
+  <groupId>org.eclipse.xpanse.modules</groupId>
+  <artifactId>workflow</artifactId>
+  <properties>
+    <maven.compiler.source>17</maven.compiler.source>
+    <maven.compiler.target>17</maven.compiler.target>
+  </properties>
+
+  <dependencies>
+    <dependency>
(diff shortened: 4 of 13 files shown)


runtime/src/main/resources/application.properties
```diff
@@ -6,12 +6,18 @@
 server.port=8080
 server.shutdown=graceful
 spring.lifecycle.timeout-per-shutdown-phase=300s
-spring.profiles.active=zitadel,zitadel-testbed,terraform-boot
+spring.profiles.active=zitadel,zitadel-testbed,nflow.db.h2
```
swaroopar, line 9: need a simple profile name.

runtime/src/main/resources/application.properties
```diff
@@ -6,12 +6,18 @@
...
 spring.banner.location=classpath:banner.txt
-spring.datasource.url=jdbc:h2:file:./testdb;DB_CLOSE_ON_EXIT=FALSE
-spring.datasource.driverClassName=org.h2.Driver
-spring.datasource.username=xpanse
-spring.datasource.password=Xpanse@2023
+# spring.datasource.url=jdbc:h2:file:./testdb;DB_CLOSE_ON_EXIT=FALSE
```
swaroopar, line 11: uncomment these properties.

runtime/src/main/resources/application.properties
```diff
@@ -6,12 +6,18 @@
...
+# spring.datasource.driverClassName=org.h2.Driver
+# spring.datasource.username=xpanse
+# spring.datasource.password=Xpanse@2023
+nflow.db.h2.driver=org.h2.Driver
+nflow.db.h2.url=jdbc:h2:file:./testdb;DB_CLOSE_ON_EXIT=FALSE
```
swaroopar, line 16: we can use same database for the nflow DB as well.

runtime/src/main/resources/application.properties
```diff
@@ -39,4 +45,5 @@ terraform.provider.aws.version=~> 4.0
 terraform.provider.openstack.version=>= 1.48.0
 terraform.provider.flexibleengine.version=>= 1.30.0
 terraform.provider.huaweicloud.version=~> 1.51.0
-terraform.provider.scs.version=~> 1.51.0
\ No newline at end of file
+terraform.provider.scs.version=~> 1.51.0
+logging.level.root=warn
```
swaroopar, line 49: this is needed for which system? we must control all via log config xml.

runtime/src/main/resources/application.properties
```diff
@@ -6,12 +6,18 @@
...
 spring.banner.location=classpath:banner.txt
-spring.datasource.url=jdbc:h2:file:./testdb;DB_CLOSE_ON_EXIT=FALSE
-spring.datasource.driverClassName=org.h2.Driver
-spring.datasource.username=xpanse
-spring.datasource.password=Xpanse@2023
+# spring.datasource.url=jdbc:h2:file:./testdb;DB_CLOSE_ON_EXIT=FALSE
+# spring.datasource.driverClassName=org.h2.Driver
+# spring.datasource.username=xpanse
+# spring.datasource.password=Xpanse@2023
+nflow.db.h2.driver=org.h2.Driver
```
swaroopar, line 15: pls add relevant config for mariadb as well.

modules/workflow/src/main/java/org/eclipse/xpanse/modules/workflow/definition/ServiceMigrationWorkflow.java
```diff
@@ -0,0 +1,90 @@
...
+        log.warn("Stopping managed service with id {}", oldId);
+        DeployTask deployTask = new DeployTask();
+        deployTask.setId(UUID.fromString(oldId));
+        Deployment deployment = this.deployService.getDestroyHandler(deployTask);
+        deployService.destroyService(deployment, deployTask);
+        return NextAction.moveToState(DONE, "destroy service");
+    }
+
+    public void done(StateExecution execution) {
+        log.warn("migrate process ended");
```
swaroopar, line 83: log level should be info. 
Also the ID of the order should be in the log msg.

modules/workflow/src/main/java/org/eclipse/xpanse/modules/workflow/definition/ServiceMigrationWorkflow.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    public void error(StateExecution execution) {
+        log.info("migrate process error");
```
swaroopar, line 87: log level should be error.
Also the ID of the order should be in the log msg.

runtime/pom.xml
```diff
@@ -148,6 +148,12 @@
             <artifactId>archunit-junit5</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>io.nflow</groupId>
+            <artifactId>nflow-rest-api-common</artifactId>
+            <version>9.0.0</version>
+            <scope>compile</scope>
```
swaroopar, line 155: compile is the default scope. you can remove it from everywhere.

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java
```diff
@@ -300,6 +310,16 @@ public Deployment getDestroyHandler(DeployTask deployTask) {
      */
     @Async("taskExecutor")
     public void asyncDestroyService(Deployment deployment, DeployTask deployTask) {
+        destroyService(deployment, deployTask);
+    }
+
+    /**
+     * Method to deploy service.
```
swaroopar, line 317: method to destroy service.

modules/workflow/pom.xml
```diff
@@ -0,0 +1,77 @@
...
+  <parent>
+    <artifactId>modules</artifactId>
+    <groupId>org.eclipse.xpanse</groupId>
+    <version>1.0.0-SNAPSHOT</version>
+  </parent>
+
+  <modelVersion>4.0.0</modelVersion>
+  <artifactId>workflow</artifactId>
+
+  <properties>
```
swaroopar, line 14: we can remove these properties. they must be taken from parent pom.

modules/workflow/pom.xml
```diff
@@ -0,0 +1,77 @@
...
+    <maven.compiler.source>17</maven.compiler.source>
+    <maven.compiler.target>17</maven.compiler.target>
+  </properties>
+
+  <dependencies>
+    <dependency>
+      <groupId>io.nflow</groupId>
+      <artifactId>nflow-rest-api-spring-web</artifactId>
+      <version>9.0.0</version>
```
swaroopar, line 23: move version to parent pom.

swaroopar (changes requested):
comments added

modules/workflow/pom.xml
```diff
@@ -0,0 +1,77 @@
...
+    </dependency>
+    <dependency>
+      <groupId>org.eclipse.xpanse.modules</groupId>
+      <artifactId>models</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.projectlombok</groupId>
+      <artifactId>lombok</artifactId>
+      <scope>compile</scope>
```
swaroopar, line 33: remove compile scope from everywhere.

modules/workflow/pom.xml
```diff
@@ -0,0 +1,77 @@
...
+    </dependency>
+    <dependency>
+      <groupId>org.eclipse.xpanse.modules</groupId>
+      <artifactId>deployment</artifactId>
+      <version>1.0.0-SNAPSHOT</version>
```
swaroopar, line 38: version must be ${project.version} here.

modules/workflow/pom.xml
```diff
@@ -0,0 +1,77 @@
...
+            <phase>generate-resources</phase>
+            <goals>
+              <goal>unpack</goal>
+            </goals>
+            <configuration>
+              <artifactItems>
+                <artifactItem>
+                  <groupId>io.nflow</groupId>
+                  <artifactId>nflow-explorer</artifactId>
+                  <version>9.0.0</version>
```
swaroopar, line 64: version must be in parent pom

runtime/pom.xml
```diff
@@ -148,6 +148,12 @@
             <artifactId>archunit-junit5</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
swaroopar, line 151: its better we move this dependency in the workflow module pom itself.

modules/workflow/src/main/java/org/eclipse/xpanse/modules/workflow/definition/ServiceMigrationWorkflow.java
```diff
@@ -0,0 +1,90 @@
...
+ */
+@Slf4j
+@Component
+public class ServiceMigrationWorkflow extends WorkflowDefinition {
+
+    public static final String NEW_ID = "newId";
+    public static final String DEPLOY_REQUEST = "deployRequest";
+    public static final String OLD_ID = "oldId";
+
+    public static final String TYPE = "serviceMigrateProcess";
```
swaroopar, line 33: serviceMigrateProcess -> serviceMigrationWorkflow

modules/workflow/src/main/java/org/eclipse/xpanse/modules/workflow/definition/ServiceMigrationWorkflow.java
```diff
@@ -0,0 +1,90 @@
...
+    private static final WorkflowState ERROR = new State("error", WorkflowStateType.manual,
+            "Manual processing of failed migration");
+
+    @Autowired
+    private DeployService deployService;
+
+    public ServiceMigrationWorkflow() {
+        super(TYPE, DEPLOY_SERVICE, ERROR, WorkflowUtils.getWorkflowSettings());
+        permit(DEPLOY_SERVICE, DESTROY_SERVICE);
+        permit(DEPLOY_SERVICE, DONE);
```
swaroopar, line 54: should be permit(DESTROY_SERVICE, DONE); ?

swaroopar (changes requested):

modules/api/src/main/java/org/eclipse/xpanse/api/ServiceDeployerApi.java
```diff
@@ -61,6 +65,11 @@ public class ServiceDeployerApi {
     @Resource
     private DeployService deployService;
 
+    @Inject
```
swaroopar, line 68: please move everything to constructor

modules/workflow/src/main/java/org/eclipse/xpanse/modules/workflow/definition/ServiceMigrationWorkflow.java
```diff
@@ -0,0 +1,90 @@
...
+            new State("destroyService", WorkflowStateType.normal,
+                    "destroy service");
+
+    private static final WorkflowState DONE = new State("done", end,
+            "Migration process finished");
+
+    private static final WorkflowState ERROR = new State("error", WorkflowStateType.manual,
+            "Manual processing of failed migration");
+
+    @Autowired
```
swaroopar, line 48: move this to constructor.

modules/workflow/src/main/java/org/eclipse/xpanse/modules/workflow/definition/ServiceMigrationWorkflow.java
```diff
@@ -0,0 +1,90 @@
...
+    public static final String TYPE = "serviceMigrateProcess";
+
+    private static final WorkflowState DEPLOY_SERVICE =
+            new State("deployService", WorkflowStateType.start, "deploy service");
+
+    private static final WorkflowState DESTROY_SERVICE =
+            new State("destroyService", WorkflowStateType.normal,
+                    "destroy service");
+
+    private static final WorkflowState DONE = new State("done", end,
```
swaroopar, line 42: can you please also add a 'start' state?

modules/workflow/src/main/java/org/eclipse/xpanse/modules/workflow/definition/ServiceMigrationWorkflow.java
swaroopar, line 87: migration of service ${OLD_ID} completed successfully

runtime/src/main/resources/nflow-rest-api.properties
```diff
@@ -0,0 +1,9 @@
+#
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Huawei Inc.
+#
+#
+nflow.rest.cors.enabled=false
```
swaroopar, line 6: profile can be just 'nflow'

swaroopar (changes requested):

modules/api/pom.xml
```diff
@@ -83,5 +83,17 @@
...
+        <dependency>
+            <groupId>io.nflow</groupId>
+            <artifactId>nflow-engine</artifactId>
+            <version>9.0.0</version>
+            <scope>compile</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.xpanse</groupId>
+            <artifactId>workflow</artifactId>
+            <version>1.0.0-SNAPSHOT</version>
```
swaroopar, line 95: version should be project.version

modules/workflow/src/main/java/org/eclipse/xpanse/modules/workflow/definition/ServiceMigrationWorkflow.java
```diff
@@ -0,0 +1,90 @@
...
+import org.eclipse.xpanse.modules.workflow.utils.WorkflowUtils;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+/**
+ *
+ */
+@Slf4j
+@Component
+public class ServiceMigrationWorkflow extends WorkflowDefinition {
```
swaroopar, line 27: since we will have both terraform-boot and terraform-local.. i dont think this workflow will work also for terraform-boot. Better to enable this only for profiles - terraform-local and nflow

modules/workflow/src/main/java/org/eclipse/xpanse/modules/workflow/definition/ServiceMigrationWorkflow.java
```diff
@@ -0,0 +1,90 @@
+package org.eclipse.xpanse.modules.workflow.definition;
```
swaroopar, line 1: pls also add 'nflow' to package name. Just incase in the future, if we plan to support any other workflow frameworks.

swaroopar (changes requested):

runtime/src/main/resources/application.properties
WangLiNaruto: @swaroopar Hi Swaroop
image
According to the definition of nFlow, we need to activate its H2 or mariadb data source. The activation key of the H2 database is nflow.db.h2, which we cannot modify.

runtime/src/main/resources/nflow-rest-api.properties
WangLiNaruto: @swaroopar Hi Swaroop
 https://github.com/NitorCreations/nflow/blob/535348a22d159ca51e536a861c94ce17854881bb/nflow-rest-api-common/src/main/java/io/nflow/rest/config/NflowRestApiPropertiesConfiguration.java#L7 , This configuration file is the built-in configuration file when springboot integrates nFlow, not a Profile file. If we want to use nFlow, it must be configured. Springhuizidong goes to the classpath to find this file. We cannot use Profile to decide whether to use it or not.

runtime/src/main/resources/application.properties
swaroopar: @WangLiNaruto ok got it. then we can keep it as it is.

runtime/src/main/resources/nflow-rest-api.properties
swaroopar: @WangLiNaruto then can we skip having this file? We can directly put the properties in the application.properties and add the profile to the default profiles.

PR #914, opened 2023-09-26
run validation aspect only for API methods

swaroopar:
the earlier aspect was impacting all methods in api package and I assume this was causing the test-coverage job to fail. This a fix to test this error. unable to reproduce it locally.


iskey (approved):
LGTM

PR #921, opened 2023-09-28
upload runtime jar as release asset

swaroopar:
- [x] update job to handle version bump in correct sequence
- [x] upload runtime jar to release assets.
- [x] set default version to 0.0.1-SNAPSHOT so that the first major release is set to 1.0.0


iskey (approved):
LGTM

PR #934, opened 2023-10-09
implement new migrate service using terraform-local

### LINKED ISSUE #887: implement new migrate service using terraform-local

WangLiNaruto:
Fixed #887 
H2:
image
image

Mysql:
image
image


modules/workflow/src/main/resources/activiti.cfg.xml
```diff
@@ -0,0 +1,30 @@
+<beans xmlns="http://www.springframework.org/schema/beans"
```
swaroopar, line 1: would prefer to move this configuration as `@bean` or `@component`

modules/workflow/src/main/java/org/eclipse/xpanse/modules/workflow/migrate/MigrateDeployProcess.java
```diff
@@ -0,0 +1,37 @@
...
+@Component
+public class MigrateDeployProcess implements Serializable, JavaDelegate {
+
+    /**
+     * Methods when performing deployment tasks.
+     */
+    @Override
+    public void execute(DelegateExecution execution) {
+        String processInstanceId = execution.getProcessInstanceId();
+        RuntimeService runtimeService = SpringUtil.getObject(RuntimeService.class);
```
swaroopar, line 28: do we really need to use `SpringUtil` here to get the `RuntimeService` and `DeployService` beans?
Since `MigrateDeployProcess`  is also a spring bean, can we not simply autowire them.

modules/workflow/src/main/java/org/eclipse/xpanse/modules/workflow/migrate/MigrateDeployProcess.java
```diff
@@ -0,0 +1,37 @@
...
+        DeployService deployService = SpringUtil.getObject(DeployService.class);
+        Map<String, Object> variables =
+                runtimeService.getVariables(processInstanceId);
+        UUID newId = (UUID) variables.get("newId");
+        CreateRequest createRequest = (CreateRequest) variables.get("createRequest");
+        boolean isDeploySuccess = deployService.deployService(newId, createRequest);
+        runtimeService.setVariable(processInstanceId, "isDeploySuccess", isDeploySuccess);
```
swaroopar, line 35: Should we handle any logic here or in the BPMN whenever we want to retry the job?

swaroopar (changes requested):
@WangLiNaruto This looks really neat and clean. Just some minor comments i have added. Additionally - 

1. in the screenshot which you put, the 'customerServiceName' column is blank in the migrated service. This must be filled. 
2. Is there an admin UI from activiti as well that is automatically deployed and can be used for viewing workflows?

modules/workflow/src/main/java/org/eclipse/xpanse/modules/workflow/migrate/MigrateDeployProcess.java
WangLiNaruto: @swaroopar  Hi swaroopar , I'm very sorry. I also automatically assembled RuntimeService and DeployService in MigrateDeployProcess at the beginning, but the instance objects of RuntimeService and DeployService obtained were null.

modules/workflow/src/main/java/org/eclipse/xpanse/modules/workflow/migrate/MigrateDeployProcess.java
WangLiNaruto: @swaroopar  sure...Our process tasks include deployment and destruction. The processing logic corresponding to the process is in MigrateDeployProcess and MigrateDestroyProcess. With the automatic flow after the process is started, we will automatically call the execute function in these two classes to execute our business logic. We Task logic can be processed there, or any logic processing class can be added to the task node in the flowchart BPMN

modules/workflow/src/main/java/org/eclipse/xpanse/modules/workflow/migrate/MigrateDeployProcess.java
swaroopar: @WangLiNaruto then there is some other problem. we must check why it is returned null.

modules/api/src/main/java/org/eclipse/xpanse/api/controllers/ServiceDeployerApi.java
```diff
@@ -179,6 +184,27 @@ public Response purge(@PathVariable("id") String id) {
...
+    @PostMapping(value = "/services/migrate/{id}", produces = MediaType.APPLICATION_JSON_VALUE)
+    @ResponseStatus(HttpStatus.ACCEPTED)
+    public UUID migrate(@PathVariable("id") String id,
+            @Valid @RequestBody CreateRequest deployRequest) {
+        UUID newId = UUID.randomUUID();
+        Map<String, Object> variable = new HashMap<>();
+        variable.put("id", id);
+        variable.put("newId", newId);
+        variable.put("createRequest", deployRequest);
+        processUtils.startProcess("migrate", variable);
```
swaroopar, line 204: here we must not wait for the complete process to finish. We must handover the process to an async thread and simply return the new ID immediately.

swaroopar (changes requested):

modules/workflow/src/main/java/org/eclipse/xpanse/modules/workflow/handle/migrate/MigrateDeployProcess.java
```diff
@@ -0,0 +1,43 @@
...
+@Component
+public class MigrateDeployProcess implements Serializable, JavaDelegate {
+
+    /**
+     * Methods when performing deployment tasks.
+     */
+    @Override
+    public void execute(DelegateExecution execution) {
+        String processInstanceId = execution.getProcessInstanceId();
+        RuntimeService runtimeService = SpringUtil.getObject(RuntimeService.class);
```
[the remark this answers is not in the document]
WangLiNaruto: I tried to autowire them directly, but a null pointer error was reported every time. I looked for related solutions.https://stackoverflow.com/questions/6474324/how-can-i-access-a-spring-bean-in-activiti-javadelegate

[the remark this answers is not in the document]
swaroopar: @WangLiNaruto According to this link, we must change the type of task - https://stackoverflow.com/questions/25894248/how-to-access-spring-beans-in-activiti-javadelegate-tasks

modules/workflow/src/main/java/org/eclipse/xpanse/modules/workflow/handle/migrate/MigrateDeployProcess.java
```diff
@@ -0,0 +1,54 @@
...
+ */
+@Slf4j
+@Component
+public class MigrateDeployProcess implements Serializable, JavaDelegate {
+
+    private static RuntimeService runtimeService;
+    private static DeployService deployService;
+
+    @Autowired
+    public void setRuntimeService(RuntimeService runtimeService) {
```
swaroopar, line 32: instead of setters, please add a constructor.

modules/workflow/src/main/java/org/eclipse/xpanse/modules/workflow/utils/SpringUtil.java
```diff
@@ -0,0 +1,35 @@
...
+import org.springframework.beans.BeansException;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.ApplicationContextAware;
+import org.springframework.stereotype.Component;
+
+/**
+ * spring tool class.
+ */
+@Component
+public class SpringUtil implements ApplicationContextAware {
```
swaroopar, line 18: we can delete this class.

runtime/src/main/resources/application.properties
```diff
@@ -6,7 +6,7 @@
 server.port=8080
 server.shutdown=graceful
 spring.lifecycle.timeout-per-shutdown-phase=300s
-spring.profiles.active=zitadel,zitadel-testbed,terraform-boot
+spring.profiles.active=zitadel,zitadel-testbed
```
swaroopar, line 9: we must enable terraform-boot by default.

swaroopar (changes requested):

modules/workflow/src/main/resources/processes/migrate.bpmn20.xml
```diff
@@ -0,0 +1,111 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
swaroopar, line 1: does this file really need 'bpmn20' to be appended to its name? i would prefer to have it without it.

modules/workflow/src/main/java/org/eclipse/xpanse/modules/workflow/config/ActivitiConfig.java
```diff
@@ -0,0 +1,95 @@
...
+    public ProcessEngineConfiguration processEngineConfiguration() throws IOException {
+        SpringProcessEngineConfiguration processEngineConfiguration =
+                new SpringProcessEngineConfiguration();
+        processEngineConfiguration.setDataSource(dataSource);
+        processEngineConfiguration.setTransactionManager(
+                (PlatformTransactionManager) transactionManager);
+        processEngineConfiguration.setDatabaseSchemaUpdate("true");
+        processEngineConfiguration.setDeploymentMode("single-resource");
+        Resource[] resources = new PathMatchingResourcePatternResolver().getResources(
+                ResourceLoader.CLASSPATH_URL_PREFIX + "processes/**.bpmn20.xml");
```
swaroopar, line 54: its better we load all files of type 'xml' under the processes folder.

swaroopar (changes requested):

modules/workflow/src/main/resources/processes/migrate.bpmn20.xml
WangLiNaruto: Make sure that the file ends with .bpmn20.xml or .bpmn, since otherwise the engine won’t pick up this file for deployment.
image

modules/workflow/src/main/java/org/eclipse/xpanse/modules/workflow/handle/migrate/MigrateDeployProcess.java
WangLiNaruto: This is very magical. If you use constructor injection, you will not be able to instantiate MigrateDeployProcess, but setter injection is no problem.
image

swaroopar (approved):
LGTM

PR #938, opened 2023-10-11
switch default database from mariadb to mysql

### LINKED ISSUE #937: switch default database from mariadb to mysql
In order to use a DB that supports all usecases, we must switch from MariaDB to MySql as our default supported DB. 

- [ ] replace application-mariadb.properties file with replace application-mysql.properties file. 
- [ ] Update contents to use MySql JDBC driver. 
- [ ] Update 'DatabaseType' Enum
- [ ] Update AbstractMariaDbIntegrationTest 
- [ ] Update jdbc driver dependency - https://github.com/eclipse-xpanse/xpanse/blob/d4dd559520a5da489678569c57e685909e93eba7/modules/database/pom.xml#L46

WangLiNaruto:
Fixed #937 
image
image


swaroopar (approved):
@WangLiNaruto Changes looks good to me. Do you know if the

iskey (approved):
LGTM

PR #944, opened 2023-10-13
Service templates data only return necessary fields for ordering service

### LINKED ISSUE #940: service catalog API must only return data that is necessary for ordering the service.
currently we use the 'UserAvailableServiceVo' data type for both service template API and service catalog. but not all data must be returned in the service catalog API which is for end users. 

- [ ] UserAvailableServiceVo -> Rename to ServiceTemplateDetailVo
- [ ] Create new data model -> `UserOrderableServiceVo' -> should be same as the existing UserAvailableServiceVo but without namespace, deployment, createTime, lastModifiedTime, serviceRegistrationState, flavors -> must be without properties. 
- [ ] catalog API returns services only which are in state registered and updated only. 
- [ ] Test changes with the UI to see nothing breaks.

baixinsui:
fixed #940 
related https://github.com/eclipse-xpanse/xpanse-ui/pull/225


modules/models/src/main/java/org/eclipse/xpanse/modules/models/servicetemplate/view/UserOrderableServiceVo.java
```diff
@@ -0,0 +1,81 @@
...
+    @Schema(description = "The icon of the available service.")
+    private String icon;
+
+    @NotNull
+    @Schema(description = "The variables for the deployment, which will be passed to the deployer.")
+    private List<@Valid DeployVariable> variables;
+
+    @NotNull
+    @Schema(description = "The flavors of the available service.")
+    private List<@Valid Flavor> flavors;
```
swaroopar, line 75: Is it possible to return only flavor names instead of the properties also?

modules/models/src/main/java/org/eclipse/xpanse/modules/models/servicetemplate/view/UserOrderableServiceVo.java
```diff
@@ -0,0 +1,81 @@
...
+@Data
+@EqualsAndHashCode(callSuper = true)
+public class UserOrderableServiceVo extends RepresentationModel<UserOrderableServiceVo> {
+
+    @NotNull
+    @Schema(description = "The id of the available service.")
+    private UUID id;
+
+    @NotNull
+    @Schema(description = "The catalog of the available service.")
```
swaroopar, line 38: The cateagory....

modules/models/src/main/java/org/eclipse/xpanse/modules/models/servicetemplate/view/UserOrderableServiceVo.java
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * The view model for the end-user to query the available services.
+ */
+@Data
+@EqualsAndHashCode(callSuper = true)
+public class UserOrderableServiceVo extends RepresentationModel<UserOrderableServiceVo> {
+
+    @NotNull
+    @Schema(description = "The id of the available service.")
```
swaroopar, line 34: available service. -> orderable service everywhere

swaroopar (changes requested):
small comments added.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/servicetemplate/view/UserOrderableServiceVo.java
baixinsui: Whether to modify the filed in the related  API and Swagger description？
image

modules/models/src/main/java/org/eclipse/xpanse/modules/models/servicetemplate/view/UserOrderableServiceVo.java
baixinsui: fixed

modules/models/src/main/java/org/eclipse/xpanse/modules/models/servicetemplate/view/UserOrderableServiceVo.java
swaroopar: @baixinsui yes, we change this too.

baixinsui: fixed all.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/servicetemplate/view/UserOrderableServiceVo.java
baixinsui: updated

swaroopar (approved):
LGTM

PR #952, opened 2023-10-16
Refactor Flavor data model

### LINKED ISSUE #951: Refactor Flavor data model
The [Flavor](https://github.com/eclipse-xpanse/xpanse/blob/145070f719637cba655176f1f0fc701b28ae1c73/modules/models/src/main/java/org/eclipse/xpanse/modules/models/servicetemplate/Flavor.java#L19) type is currently shared between service template and orderable services. This must be changed. 

- [ ] Remove `properties` field from Flavor type and use it in the `UserOrderableServiceVo` type. 
- [ ] Create a new type `FlavorWithProperties` - which extends the Flavor type and additionally has `properties` field. Use it in the `ServiceTemplateDetailVo` type. 
- [ ] Check and update UI.

baixinsui:
fixed #951 
related https://github.com/eclipse-xpanse/xpanse-ui/pull/230

### DIFF
--- a/modules/api/src/main/java/org/eclipse/xpanse/api/controllers/ServiceCatalogApi.java
+++ b/modules/api/src/main/java/org/eclipse/xpanse/api/controllers/ServiceCatalogApi.java
@@ -23,6 +23,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.eclipse.xpanse.modules.database.servicetemplate.ServiceTemplateEntity;
 import org.eclipse.xpanse.modules.models.service.common.enums.Category;
 import org.eclipse.xpanse.modules.models.service.common.enums.Csp;
+import org.eclipse.xpanse.modules.models.servicetemplate.FlavorBasic;
 import org.eclipse.xpanse.modules.models.servicetemplate.query.ServiceTemplateQueryModel;
 import org.eclipse.xpanse.modules.models.servicetemplate.view.UserOrderableServiceVo;
 import org.eclipse.xpanse.modules.servicetemplate.ServiceTemplateManage;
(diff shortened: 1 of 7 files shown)


modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/utils/DeployEnvironments.java
```diff
@@ -17,7 +17,7 @@
 import org.eclipse.xpanse.modules.models.service.common.enums.Csp;
 import org.eclipse.xpanse.modules.models.service.deploy.exceptions.FlavorInvalidException;
 import org.eclipse.xpanse.modules.models.servicetemplate.DeployVariable;
-import org.eclipse.xpanse.modules.models.servicetemplate.Flavor;
+import org.eclipse.xpanse.modules.models.servicetemplate.FlavorWithProperties;
```
iskey, line 20: I feel like keeping the original `Flavor` is better, with adding a new class named `FlavorWithoutProperties`/`FlavorBasic`.

iskey (changes requested):
comments added

swaroopar (changes requested):
We can go with the suggestion from Yichao.. FlavorBasic is fine for me.

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/utils/DeployEnvironments.java
baixinsui: Ok. I added  a new class `FlavorBasic` without the field `properties` for user querying.

modules/api/src/main/java/org/eclipse/xpanse/api/controllers/ServiceCatalogApi.java
```diff
@@ -166,7 +167,13 @@ private UserOrderableServiceVo convertToUserOrderableServiceVo(
             userOrderableServiceVo.setDescription(
                     serviceTemplateEntity.getOcl().getDescription());
             userOrderableServiceVo.setBilling(serviceTemplateEntity.getOcl().getBilling());
-            userOrderableServiceVo.setFlavors(serviceTemplateEntity.getOcl().getFlavors());
+            List<FlavorBasic> flavorBasics = serviceTemplateEntity.getOcl().getFlavors()
+                    .stream().map(flavorWithProperties -> {
+                        FlavorBasic flavorBasic = new FlavorBasic();
+                        BeanUtils.copyProperties(flavorWithProperties, flavorBasic);
```
swaroopar, line 173: @baixinsui flavorWithProperties must be just flavor now?

baixinsui: Updated.

iskey (approved):
LGTM

swaroopar (approved):

PR #955, opened 2023-10-18
update error handling in the migrate BPMN

### LINKED ISSUE #946: update error handling in the migrate BPMN
The newly added migration BPMN must be updated to work in the following way.

- [ ] Retry deploy and destroy only for a configured number of times. 
- [ ] After retries - mark the job as failed using error boundaries. 
- [ ] Also ensure the code handles complete job retries - i.e., when deploy is successful and destroy fails - then when the operator/admin retries the entire job, then deploy must be skipped and directly go to destroy. We can directly check the status of the service DB to decide if we must retry the deploy or if we can skip it.

WangLiNaruto:
Fixed #946


modules/models/src/main/java/org/eclipse/xpanse/modules/models/servicetemplate/Billing.java
```diff
@@ -17,7 +18,7 @@
  * Defines the billing model of the managed service.
  */
 @Data
-public class Billing {
+public class Billing implements Serializable {
```
baixinsui, line 22: Add `serialVersionUID` to avoid potential incompatibilities during deserialization.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/servicetemplate/CloudServiceProvider.java
```diff
@@ -17,7 +18,7 @@
  * Defines for the Cloud Service Provider.
  */
 @Data
-public class CloudServiceProvider {
+public class CloudServiceProvider implements Serializable {
```
baixinsui, line 22: Add serialVersionUID to avoid potential incompatibilities during deserialization.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/servicetemplate/DeployVariable.java
```diff
@@ -19,7 +20,7 @@
  * Defines for the deploy variable.
  */
 @Data
-public class DeployVariable {
+public class DeployVariable implements Serializable {
```
baixinsui, line 24: Add serialVersionUID to avoid potential incompatibilities during deserialization.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/servicetemplate/Deployment.java
```diff
@@ -19,7 +20,7 @@
  * Defines the Deployment.
  */
 @Data
-public class Deployment {
+public class Deployment implements Serializable {
```
baixinsui, line 24: Add serialVersionUID to avoid potential incompatibilities during deserialization.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/servicetemplate/Flavor.java
```diff
@@ -19,7 +20,7 @@
 @Data
 @ToString(callSuper = true)
 @EqualsAndHashCode(callSuper = true)
-public class Flavor extends FlavorBasic {
+public class Flavor extends FlavorBasic implements Serializable {
```
baixinsui, line 24: Add serialVersionUID to avoid potential incompatibilities during deserialization.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/servicetemplate/Ocl.java
```diff
@@ -26,7 +27,7 @@
 @Valid
 @Data
 @Slf4j
-public class Ocl {
+public class Ocl implements Serializable {
```
baixinsui, line 31: Add serialVersionUID to avoid potential incompatibilities during deserialization.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/servicetemplate/Region.java
```diff
@@ -9,13 +9,14 @@
...
 import jakarta.validation.constraints.NotNull;
+import java.io.Serializable;
 import lombok.Data;
 
 /**
  * The regions of the Cloud Service Provider.
  */
 @Data
-public class Region {
+public class Region implements Serializable {
```
baixinsui, line 20: Add serialVersionUID to avoid potential incompatibilities during deserialization.

modules/workflow/src/main/java/org/eclipse/xpanse/modules/workflow/bean/MigrateState.java
```diff
@@ -0,0 +1,31 @@
...
+package org.eclipse.xpanse.modules.workflow.bean;
+
+import com.fasterxml.jackson.annotation.JsonValue;
+
+/**
+ * Migration status enum.
+ */
+public enum MigrateState {
+    MIGRATING("migrating"),
+    MIGRATED("migrated");
```
swaroopar, line 16: Do we also need a migration failed status here?

modules/workflow/src/main/java/org/eclipse/xpanse/modules/workflow/handle/migrate/DeployFailedManualHandlerListener.java
```diff
@@ -0,0 +1,36 @@
...
+import org.activiti.engine.delegate.DelegateTask;
+import org.activiti.engine.delegate.TaskListener;
+import org.eclipse.xpanse.modules.workflow.consts.MigrateConstants;
+import org.springframework.beans.factory.annotation.Autowired;
+
+/**
+ * Monitoring class for manual processing after deployment failure and retry.
+ */
+@Slf4j
+public class DeployFailedManualHandlerListener implements TaskListener {
```
swaroopar, line 20: Should this class be annotated as a spring bean? Otherwise the auto wiring would not work

modules/workflow/src/main/java/org/eclipse/xpanse/modules/workflow/handle/migrate/DestroyFailedManualHandlerListener.java
```diff
@@ -0,0 +1,36 @@
...
+import org.activiti.engine.delegate.DelegateTask;
+import org.activiti.engine.delegate.TaskListener;
+import org.eclipse.xpanse.modules.workflow.consts.MigrateConstants;
+import org.springframework.beans.factory.annotation.Autowired;
+
+/**
+ * Monitoring class for manual processing after destroy failure and retry.
+ */
+@Slf4j
+public class DestroyFailedManualHandlerListener implements TaskListener {
```
swaroopar, line 20: Same here . Is this a spring bean?

swaroopar (changes requested):

modules/workflow/src/main/java/org/eclipse/xpanse/modules/workflow/bean/MigrateState.java
```diff
@@ -0,0 +1,32 @@
+/*
```
swaroopar, line 1: the bean package can be moved to the "models" module. Thats where all data models are present.

modules/workflow/src/main/java/org/eclipse/xpanse/modules/workflow/handle/migrate/DestroyFailedManualHandlerListener.java
```diff
@@ -0,0 +1,38 @@
...
+import org.springframework.stereotype.Component;
+
+/**
+ * Monitoring class for manual processing after destroy failure and retry.
+ */
+@Slf4j
+@Component
+public class DestroyFailedManualHandlerListener implements TaskListener {
+
+    private static RuntimeService runtimeService;
```
swaroopar, line 24: is there a specific reason for declaring all spring autowired components as static? if not, we must keep it only private and final. Static is not needed. 
This applies to all beans in the workflow module.

modules/workflow/src/main/java/org/eclipse/xpanse/modules/workflow/utils/WorkflowProcessUtils.java
```diff
@@ -47,4 +62,70 @@ public ProcessInstance startProcess(String processKey, Map<String, Object> varia
...
+     * Get the tasks that the user has done through the ID of the currently logged in user.
+     *
+     * @param userId the ID of the currently logged in user.
+     */
+    public List<WorkFlowTask> doneTasks(String userId) {
+        List<HistoricTaskInstance> list = historyService.createHistoricTaskInstanceQuery()
+                .taskAssignee(userId)
+                .finished()
+                .list();
+        return list.stream().map(task -> getWorkFlow(task)).collect(Collectors.toList());
```
swaroopar, line 86: we can use method reference instead of lambda here.

modules/api/src/main/java/org/eclipse/xpanse/api/controllers/WorkFlowApi.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    /**
+     * Complete tasks by task ID.
+     */
+    @Tag(name = "WorkFlow", description = "APIs to manage the WorkFlow")
+    @Operation(description = "Complete tasks by task ID.")
+    @PutMapping(value = "/workflow/task/{id}", produces = MediaType.APPLICATION_JSON_VALUE)
+    @ResponseStatus(HttpStatus.OK)
+    public void completeTask(@PathVariable("id") String taskId) {
```
swaroopar, line 91: we need here to allow user to either retry task or just close the BPMN. 
We can call the API to have "retryWorkflow" and one more method "closeWorkflow". 
The "closeWorkflow" method will simply close the BPMN.
and "retryworkflow" method should work like how you have already implemented.

swaroopar (changes requested):

modules/workflow/src/main/java/org/eclipse/xpanse/modules/workflow/handle/migrate/DestroyFailedManualHandlerListener.java
WangLiNaruto: Yes, it needs to be static. If it is not static, it will report an error.

modules/api/src/main/java/org/eclipse/xpanse/api/controllers/WorkFlowApi.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    /**
+     * Complete tasks based on task ID and set global process variables.
+     */
+    @Tag(name = "WorkFlow", description = "APIs to manage the WorkFlow")
+    @Operation(description = "Complete tasks by task ID and set global process variables .")
+    @PutMapping(value = "/workflow/task/{id}", produces = MediaType.APPLICATION_JSON_VALUE)
+    @ResponseStatus(HttpStatus.OK)
+    public void completeTask(@PathVariable("id") String taskId, @RequestBody Map<String,
```
swaroopar, line 93: @WangLiNaruto here having a generic map will make the method generic but will be confusing. 
It would be better to have something to have like this
1. manageFailedOrder -> which should have be a wrapper to this completeTask method.
2. the manageFailedOrder will have a flag -> retryOrder/closeOrder. Based on this value, we will call the completeTask method accordingly. 

Would this be possible?

swaroopar (changes requested):

baixinsui (approved):
LGTM

swaroopar (approved):
LGTM

PR #959, opened 2023-10-20
validate deploy variable value schema validation

swaroopar:
fixes https://github.com/eclipse-xpanse/xpanse/issues/958


iskey (approved):
LGTM

PR #961, opened 2023-10-20
extend migrate service for terraform-boot

### LINKED ISSUE #888: extend migrate service for terraform-boot
- [ ] Create a new webhook for migrate -> `/migrate/create` 
- [ ] For the create request sent to terraform boot -> send webhook `/migrate/create` -> 
- [ ]  `/migrate/create` -> should work exactly same as the current `deploy` webhook. But at the end also call the destroy async service. 
- [ ] in the destroy request, send the regular '/destroy' webhook. 
- [ ]  Retry deploy and destroy only for a configured number of times.
- [ ]  After retries - mark the job as failed using error boundaries.
- [ ]  Also ensure the code handles complete job retries - i.e., when deploy is successful and destroy fails - then when the operator/admin retries the entire job, then deploy must be skipped and directly go to destroy. We can directly check the status of the service DB to decide if we must retry the deploy or if we can skip it.

Similar to #887

WangLiNaruto:
Fixed #888 
image
image

### DIFF
--- a/modules/models/src/main/java/org/eclipse/xpanse/modules/models/workflow/WorkFlowTask.java
+++ b/modules/models/src/main/java/org/eclipse/xpanse/modules/models/workflow/WorkFlowTask.java
@@ -8,6 +8,7 @@ package org.eclipse.xpanse.modules.models.workflow;
 
 import io.swagger.v3.oas.annotations.media.Schema;
 import jakarta.validation.constraints.NotEmpty;
+import jakarta.validation.constraints.NotNull;
 import java.util.Date;
 import lombok.Data;
 
@@ -49,7 +50,7 @@ public class WorkFlowTask {
     @Schema(description = "The businessKey of the Process")
     private String businessKey;
 
-    @NotEmpty
+    @NotNull
     @Schema(description = "The create time of the task")
     private Date createTime;
 }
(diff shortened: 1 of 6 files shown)


modules/workflow/src/main/java/org/eclipse/xpanse/modules/workflow/handle/migrate/MigrateDeployProcess.java
```diff
@@ -53,18 +55,28 @@ public void execute(DelegateExecution execution) {
...
             log.info("Process instance: {} retry deployment service : {},RetryNum:{}",
                     processInstanceId, newId, deployRetryNum);
         }
-        MigrateRequest migrateRequest =
-                (MigrateRequest) variables.get(MigrateConstants.MIGRATE_REQUEST);
-        DeployRequest deployRequest = new DeployRequest();
-        BeanUtils.copyProperties(migrateRequest, deployRequest);
-        boolean isDeploySuccess = deployService.deployService(newId, deployRequest);
+        ServiceDeploymentState deploymentState =
+                deployService.getDeployServiceDetails(newId).getServiceDeploymentState();
```
swaroopar, line 73: if the deployment status is checked immediately, then the deployment is never successful. If i understand it right, we must wait for the webhook update and then proceed with the BPMN next steps. Not sure if this code is really handling that.

swaroopar (changes requested):

PR #967, opened 2023-10-24
extend migrate service for terraform-boot

### LINKED ISSUE #888: extend migrate service for terraform-boot

WangLiNaruto:
Fixed #888
image
image

### DIFF
--- a/modules/database/src/main/java/org/eclipse/xpanse/modules/database/service/DatabaseDeployServiceStorage.java
+++ b/modules/database/src/main/java/org/eclipse/xpanse/modules/database/service/DatabaseDeployServiceStorage.java
@@ -6,6 +6,8 @@
 
 package org.eclipse.xpanse.modules.database.service;
 
+import jakarta.persistence.EntityManager;
+import jakarta.persistence.PersistenceContext;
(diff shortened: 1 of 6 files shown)


swaroopar (changes requested):
@WangLiNaruto The idea is great. We can use this. We only have to check how to use Spring async to control threads started by completable futures.

swaroopar (approved):
LGTM

PR #985, opened 2023-10-31
Implement APIs for managing policy

baixinsui:
fixed https://github.com/eclipse-xpanse/xpanse/issues/901

### DIFF
--- a/modules/api/src/main/java/org/eclipse/xpanse/api/exceptions/handler/PolicyManageExceptionHandler.java
+++ b/modules/api/src/main/java/org/eclipse/xpanse/api/exceptions/handler/PolicyManageExceptionHandler.java
@@ -1,0 +1,80 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * SPDX-FileCopyrightText: Huawei Inc.
+ */
+
+package org.eclipse.xpanse.api.exceptions.handler;
+
+import java.util.Collections;
+import lombok.extern.slf4j.Slf4j;
+import org.eclipse.xpanse.modules.models.policy.exceptions.PoliciesEvaluationFailedException;
+import org.eclipse.xpanse.modules.models.policy.exceptions.PoliciesValidationFailedException;
+import org.eclipse.xpanse.modules.models.policy.exceptions.PolicyDuplicateException;
+import org.eclipse.xpanse.modules.models.policy.exceptions.PolicyNotFoundException;
+import org.eclipse.xpanse.modules.models.response.Response;
+import org.eclipse.xpanse.modules.models.response.ResultType;
+import org.springframework.core.Ordered;
+import org.springframework.core.annotation.Order;
+import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.annotation.ExceptionHandler;
+import org.springframework.web.bind.annotation.ResponseBody;
+import org.springframework.web.bind.annotation.ResponseStatus;
+import org.springframework.web.bind.annotation.RestControllerAdvice;
+
+/**
+ * Exception handler related to Credentials center for the REST API.
+ */
+@Slf4j
+@Order(Ordered.HIGHEST_PRECEDENCE)
+@RestControllerAdvice
+public class PolicyManageExceptionHandler {
+
+    /**
+     * Exception handler for PolicyNotFoundException.
+     */
+    @ExceptionHandler({PolicyNotFoundException.class})
+    @ResponseStatus(HttpStatus.BAD_REQUEST)
+    @ResponseBody
+    public Response handlePolicyNotFoundException(
+            PolicyNotFoundException ex) {
+        return Response.errorResponse(ResultType.POLICY_NOT_FOUND,
+                Collections.singletonList(ex.getMessage()));
+    }
+
+    /**
+     * Exception handler for PolicyDuplicateException.
+     */
+    @ExceptionHandler({PolicyDuplicateException.class})
+    @ResponseStatus(HttpStatus.BAD_REQUEST)
+    @ResponseBody
+    public Response handlePolicyDuplicateException(
+            PolicyDuplicateException ex) {
+        return Response.errorResponse(ResultType.POLICY_DUPLICATE,
+                Collections.singletonList(ex.getMessage()));
+    }
+
+    /**
+     * Exception handler for PoliciesValidationFailedException.
+     */
+    @ExceptionHandler({PoliciesValidationFailedException.class})
+    @ResponseStatus(HttpStatus.BAD_REQUEST)
+    @ResponseBody
+    public Response handlePoliciesValidationFailedException(
+            PoliciesValidationFailedException ex) {
+        return Response.errorResponse(ResultType.POLICY_VALIDATION_FAILED,
+                Collections.singletonList(ex.getMessage()));
+    }
+
+
+    /**
+     * Exception handler for PoliciesEvaluationFailedException.
+     */
+    @ExceptionHandler({PoliciesEvaluationFailedException.class})
+    @ResponseStatus(HttpStatus.BAD_REQUEST)
+    @ResponseBody
+    public Response handlePoliciesEvaluationFailedException(
+            PoliciesEvaluationFailedException ex) {
+        return Response.errorResponse(ResultType.POLICY_EVALUATION_FAILED,
+                Collections.singletonList(ex.getMessage()));
+    }
+}
--- a/modules/models/src/main/java/org/eclipse/xpanse/modules/models/policy/PolicyVo.java
+++ b/modules/models/src/main/java/org/eclipse/xpanse/modules/models/policy/PolicyVo.java
@@ -1,0 +1,74 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * SPDX-FileCopyrightText: Huawei Inc.
+ */
+
+package org.eclipse.xpanse.modules.models.policy;
+
+import com.fasterxml.jackson.annotation.JsonFormat;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import com.fasterxml.jackson.datatype.jsr310.ser.OffsetDateTimeSerializer;
+import io.swagger.v3.oas.annotations.Hidden;
+import io.swagger.v3.oas.annotations.media.Schema;
+import jakarta.validation.constraints.NotBlank;
+import jakarta.validation.constraints.NotEmpty;
+import jakarta.validation.constraints.NotNull;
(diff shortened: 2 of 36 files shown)


modules/policy/src/main/java/org/eclipse/xpanse/modules/policy/policyman/PolicyManager.java
```diff
@@ -0,0 +1,309 @@
...
+    }
+
+    /**
+     * Create new policy.
+     *
+     * @param createRequest create policy request.
+     * @return Returns created policy view object.
+     */
+    public PolicyVo addPolicy(PolicyCreateRequest createRequest) {
+        checkPolicyWhetherCreated(createRequest.getCsp(), createRequest.getPolicy());
```
swaroopar, line 119: Does it allow multiple policies for every CSP?

modules/policy/src/main/java/org/eclipse/xpanse/modules/policy/policyman/PolicyManager.java
```diff
@@ -0,0 +1,309 @@
...
+    private void validatePolicy(String policy) {
+        boolean valid = true;
+        String errorMsg = "";
+        ValidatePolicyList validatePolicyList = new ValidatePolicyList();
+        validatePolicyList.setPolicyList(List.of(policy));
+        try {
+            ValidateResponse validateResponse =
+                    policiesValidateApi.validatePoliciesPost(validatePolicyList);
+            log.info("Validate policies response:{}", validateResponse.toString());
+            if (!(Objects.nonNull(validateResponse.getIsSuccessful())
```
swaroopar, line 266: we can directly check for null here

modules/policy/src/main/java/org/eclipse/xpanse/modules/policy/policyman/PolicyManager.java
```diff
@@ -0,0 +1,309 @@
...
+    private void evaluatePolicies(List<String> policies, String input) {
+        boolean valid = true;
+        String errorMsg = "";
+        EvalCmdList cmdList = new EvalCmdList();
+        cmdList.setPolicyList(policies);
+        cmdList.setInput(input);
+        try {
+            EvalResult evalResult = policiesEvaluationApi.evaluatePoliciesPost(cmdList);
+            log.info("Evaluate policies response:{}", evalResult.toString());
+            if (!(Objects.nonNull(evalResult.getIsSuccessful())
```
swaroopar, line 291: we can directly check for null here

modules/policy/src/main/java/org/eclipse/xpanse/modules/policy/policyman/PolicyManager.java
```diff
@@ -0,0 +1,309 @@
...
+        }
+        if (!StringUtils.equals(getCurrentLoginUserId(), existedEntity.getUserId())) {
+            throw new AccessDeniedException(
+                    "No permissions to delete policy belonging to other users.");
+        }
+        policyStorage.deletePolicyById(id);
+    }
+
+
+    private String getCurrentLoginUserId() {
```
swaroopar, line 210: we have this method duplicated at many places. We must move it to one single common place.

modules/policy/src/main/java/org/eclipse/xpanse/modules/policy/policyman/PolicyManager.java
```diff
@@ -0,0 +1,309 @@
...
+        CurrentUserInfo currentUserInfo = identityProviderManager.getCurrentUserInfo();
+        if (Objects.nonNull(currentUserInfo)
+                && StringUtils.isNotBlank(currentUserInfo.getUserId())) {
+            return currentUserInfo.getUserId();
+        } else {
+            return "defaultUserId";
```
swaroopar, line 216: instead it is better to return Optional<String>

modules/policy/src/main/java/org/eclipse/xpanse/modules/policy/policyman/PolicyManager.java
```diff
@@ -0,0 +1,309 @@
...
+    }
+
+    /**
+     * Update policy.
+     *
+     * @param updateRequest update policy request.
+     * @return Returns updated policy view object.
+     */
+    public PolicyVo updatePolicy(PolicyUpdateRequest updateRequest) {
+        PolicyEntity existedEntity = policyStorage.findPolicyById(updateRequest.getId());
```
swaroopar, line 134: should be 'existingEntity'

modules/models/src/main/java/org/eclipse/xpanse/modules/models/policy/PolicyVo.java
```diff
@@ -0,0 +1,76 @@
...
+    private UUID id;
+
+    /**
+     * The id of user who created the policy.
+     */
+    @NotNull
+    @NotBlank
+    @NotEmpty
+    @Schema(description = "The id of user who created the policy.")
+    private String userId;
```
swaroopar, line 40: userId is not needed. Either we can remove it or hide it.

swaroopar (changes requested):

modules/policy/src/main/java/org/eclipse/xpanse/modules/policy/policyman/PolicyManager.java
baixinsui: It allows the same user to create multiple policies in the every CSP, but the policies created by the same user for the same CSP could not be duplicate.

modules/api/src/main/java/org/eclipse/xpanse/api/exceptions/handler/PolicyManageExceptionHandler.java
```diff
@@ -0,0 +1,80 @@
...
+
+    /**
+     * Exception handler for PolicyNotFoundException.
+     */
+    @ExceptionHandler({PolicyNotFoundException.class})
+    @ResponseStatus(HttpStatus.BAD_REQUEST)
+    @ResponseBody
+    public Response handlePolicyNotFoundException(
+            PolicyNotFoundException ex) {
+        return Response.errorResponse(ResultType.PLUGIN_NOT_FOUND,
```
swaroopar, line 40: ResultType must be POLICY_NOT_FOUND i guess.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/response/ResultType.java
```diff
@@ -48,7 +48,11 @@ public enum ResultType {
     TERRAFORM_BOOT_REQUEST_FAILED("Terraform Boot Request Failed"),
     METRICS_DATA_NOT_READY("Metrics Data Not Ready"),
     VARIABLE_VALIDATION_FAILED("Variable Validation Failed"),
-    VARIABLE_SCHEMA_DEFINITION_INVALID("Variable Schema Definition Invalid");
+    VARIABLE_SCHEMA_DEFINITION_INVALID("Variable Schema Definition Invalid"),
+    POLICY_NOT_FOUND("Policy Not Found"),
+    POLICY_DUPLICATE("Policy Duplicate"),
```
swaroopar, line 53: `Policy Duplicate` is better to be written as `Duplicate Policy`

modules/policy/src/main/java/org/eclipse/xpanse/modules/policy/policyman/PolicyManager.java
swaroopar: My expectation was that.. for every user, we must allow multiple policies per CSP.. Do you think we can achieve this easily? If not, let us discuss together with Yichao what he thinks.

modules/policy/src/main/java/org/eclipse/xpanse/modules/policy/policyman/PolicyManager.java
```diff
@@ -0,0 +1,305 @@
...
+    private void validatePolicy(String policy) {
+        boolean valid = true;
+        String errorMsg = "";
+        ValidatePolicyList validatePolicyList = new ValidatePolicyList();
+        validatePolicyList.setPolicyList(List.of(policy));
+        try {
+            ValidateResponse validateResponse =
+                    policiesValidateApi.validatePoliciesPost(validatePolicyList);
+            log.info("Validate policies response:{}", validateResponse.toString());
+            if (Objects.isNull(validateResponse.getIsSuccessful())
```
swaroopar, line 262: validateResponse.getIsSuccessful( ) -> is never null since this is a mandatory field. If policy-man is returning null here sometimes, then it is better we fix it in policy-man.

modules/policy/src/main/java/org/eclipse/xpanse/modules/policy/policyman/PolicyManager.java
```diff
@@ -0,0 +1,305 @@
...
+    private void evaluatePolicies(List<String> policies, String input) {
+        boolean valid = true;
+        String errorMsg = "";
+        EvalCmdList cmdList = new EvalCmdList();
+        cmdList.setPolicyList(policies);
+        cmdList.setInput(input);
+        try {
+            EvalResult evalResult = policiesEvaluationApi.evaluatePoliciesPost(cmdList);
+            log.info("Evaluate policies response:{}", evalResult.toString());
+            if (Objects.isNull(evalResult.getIsSuccessful())
```
swaroopar, line 287: same here. evalResult.getIsSuccessful() is never null.

runtime/src/test/java/org/eclipse/xpanse/runtime/modules/common/OpenApiGeneratorJarManageTest.java
```diff
@@ -42,6 +43,7 @@ void testGetOpenApiWorkdir() {
     }
 
     @Test
+    @Disabled
```
swaroopar, line 46: do we need to disable it really?

swaroopar (changes requested):

modules/policy/src/main/java/org/eclipse/xpanse/modules/policy/policyman/PolicyManager.java
baixinsui: Now the APIs allow each user to create multiple policies per CSP. No need to make any changes again.

runtime/src/test/java/org/eclipse/xpanse/runtime/modules/common/OpenApiGeneratorJarManageTest.java
baixinsui: Removed.

PR #987, opened 2023-10-31
Refactor the migration pages

### LINKED ISSUE #945: Refactor the migration pages with `/services/migrate/{id}`
Please note that this must be done only after the below two user stories are implemented. 
https://github.com/eclipse-xpanse/xpanse/issues/888
https://github.com/eclipse-xpanse/xpanse/issues/946

WangLiNaruto:
Fixed #945 
image
image


modules/workflow/src/main/java/org/eclipse/xpanse/modules/workflow/utils/WorkflowProcessUtils.java
```diff
@@ -95,7 +97,7 @@ public List<WorkFlowTask> doneTasks(String userId) {
      *
      * @param processInstanceId Process instance ID
      */
-    public String getWorkFlowState(String processInstanceId) {
+    public String getMigrateWorkFlowState(String processInstanceId) {
```
swaroopar, line 100: we must return an enum here instead of string. Then the client knows that only a certain states are possible.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/response/ResultType.java
```diff
@@ -37,6 +37,7 @@ public enum ResultType {
     ICON_PROCESSING_FAILED("Icon Processing Failed"),
     SERVICE_TEMPLATE_NOT_REGISTERED("Service Template Not Registered"),
     SERVICE_DEPLOYMENT_NOT_FOUND("Service Deployment Not Found"),
+    SERVICE_MIGRATE_NOT_FOUND("Service Migrate Not Found"),
```
swaroopar, line 40: Do we really need this? Does not seem to be used anywhere.

modules/workflow/src/main/java/org/eclipse/xpanse/modules/workflow/utils/WorkflowProcessUtils.java
```diff
@@ -110,12 +112,12 @@ public String getWorkFlowState(String processInstanceId) {
         if (instance == null) {
             if (variablesMap.containsKey(MigrateConstants.IS_RETRY_TASK)
```
swaroopar, line 113: i think i missed this in the previous MRs. I would prefer getting the status from DB and I think we are even updating the status in DB. It is better to use that instead of doing any calculation based on the "retry" flag.

modules/api/src/main/java/org/eclipse/xpanse/api/controllers/ServiceDeployerApi.java
```diff
@@ -199,13 +204,29 @@ public Response purge(@PathVariable("id") String id) {
...
+     * Get the status of service migration by service ID.
+     *
+     * @param id service ID.
+     */
+    @Tag(name = "Service", description = "APIs to manage the service instances")
+    @Operation(description = "Get the status of service migration by service ID")
+    @GetMapping(value = "/services/migration/state/{id}", produces =
+            MediaType.APPLICATION_JSON_VALUE)
+    @ResponseStatus(HttpStatus.OK)
+    public Response getMigrateServiceStateById(@PathVariable("id") String id) {
```
swaroopar, line 225: Even here it is better to define a new data model for the state. Even if it has just one field, it is ok.

swaroopar (changes requested):

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java
```diff
@@ -654,4 +661,31 @@ public void updateServiceStatus(UUID id, ServiceDeploymentState serviceDeploymen
         deployServiceEntity.setServiceDeploymentState(serviceDeploymentState);
         deployServiceStorage.storeAndFlush(deployServiceEntity);
     }
+
+    /**
+     * Get the migration status of a service by service ID.
+     */
+    public ServiceDeploymentState queryMigrateStateById(String id) {
+        String[] split = id.split("@");
```
swaroopar, line 669: we must avoid this for sure. 
The migration ID -> can be preferably the ID of the new service or worst case can be a completely new ID

swaroopar (changes requested):

modules/workflow/src/main/java/org/eclipse/xpanse/modules/workflow/handle/migrate/MigrateDeployProcess.java
```diff
@@ -52,8 +52,10 @@ public void execute(DelegateExecution execution) {
 
         String processInstanceId = execution.getProcessInstanceId();
         Map<String, Object> variables = runtimeService.getVariables(processInstanceId);
+        UUID id = (UUID) variables.get(MigrateConstants.ID);
         UUID newId = (UUID) variables.get(MigrateConstants.NEW_ID);
-        runtimeService.updateBusinessKey(processInstanceId, newId.toString());
+        runtimeService.updateBusinessKey(processInstanceId,
```
swaroopar, line 57: do we still need this change?

swaroopar (changes requested):
@WangLiNaruto Please also check the merge conflict.

swaroopar (approved):
LGTM

PR #994, opened 2023-11-03
Add new API methods for managing ISV cloud provider credentials

WangLiNaruto:


modules/api/src/main/java/org/eclipse/xpanse/api/controllers/IsvCloudCredentialsApi.java
```diff
@@ -0,0 +1,193 @@
...
+     * @return Returns list the available credential types of the cloud service provider.
+     */
+    @Tag(name = "Credentials Management",
+            description = "APIs for managing isv's cloud provider credentials")
+    @GetMapping(value = "/credential_types",
+            produces = MediaType.APPLICATION_JSON_VALUE)
+    @ResponseStatus(HttpStatus.OK)
+    @Operation(description = "Users in the ISV role get the list of available credential types of"
+            + " the cloud service provider.")
+    public List<CredentialType> getIsvCloudCredentialTypes(
```
swaroopar, line 71: This is a point that I missed to mention. This method and the next method (getIsvCloudCredentialCapabilities) are common to both ISV and end users. 
We can move it to another common API controller - CredentialsConfigApi. The definition of cloud credentials will be at just one place and will be used by both ISV and end users.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/common/exceptions/UserNoLoginException.java
```diff
@@ -0,0 +1,16 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ * SPDX-FileCopyrightText: Huawei Inc.
+ */
+
+package org.eclipse.xpanse.modules.models.common.exceptions;
+
+/**
+ * Exception is thrown when the information of the currently logged in person cannot be found.
+ */
+public class UserNoLoginException extends  RuntimeException {
```
swaroopar, line 11: Should be UserNotLoggedInException

modules/api/src/main/java/org/eclipse/xpanse/api/exceptions/handler/CommonExceptionHandler.java
```diff
@@ -170,4 +171,16 @@ public Response handleMethodArgumentTypeMismatchException(
...
                 Collections.singletonList(ex.getMessage()));
     }
+
+    /**
+     * Exception handler for UserNoLoginException.
+     */
+    @ExceptionHandler({UserNoLoginException.class})
+    @ResponseStatus(HttpStatus.UNAUTHORIZED)
+    @ResponseBody
+    public Response handleUserNoLoginException(
```
swaroopar, line 181: please also add a test for this.

modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/CredentialCenter.java
```diff
@@ -117,9 +117,6 @@ public String getCredentialOpenApiUrl(Csp csp, CredentialType type) {
      */
     public List<AbstractCredentialInfo> listCredentials(Csp csp, CredentialType type,
```
swaroopar, line 118: can you please confirm if the credential cache really works correctly when the userID is null?

swaroopar (changes requested):

modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/CredentialCenter.java
WangLiNaruto: I've looked at the code and I think it works fine when userId is null
image
image
image

swaroopar (approved):
LGTM

PR #996, opened 2023-11-07
Add tests for module policy

baixinsui:
fixed https://github.com/eclipse-xpanse/xpanse/issues/986

### DIFF
--- a/modules/policy/src/main/java/org/eclipse/xpanse/modules/policy/policyman/PolicyManager.java
+++ b/modules/policy/src/main/java/org/eclipse/xpanse/modules/policy/policyman/PolicyManager.java
@@ -117,9 +117,8 @@ public class PolicyManager {
      * @return Returns created policy view object.
      */
     public PolicyVo addPolicy(PolicyCreateRequest createRequest) {
-        checkPolicyWhetherCreated(createRequest.getCsp(), createRequest.getPolicy());
         validatePolicy(createRequest.getPolicy());
-
+        checkIfPolicyIsDuplicate(createRequest.getCsp(), createRequest.getPolicy());
         PolicyEntity newPolicy = conventToPolicyEntity(createRequest);
         PolicyEntity policyEntity = policyStorage.store(newPolicy);
         return conventToPolicyVo(policyEntity);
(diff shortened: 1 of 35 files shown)


modules/security/src/main/java/org/eclipse/xpanse/modules/security/IdentityProviderManager.java
```diff
@@ -43,33 +40,8 @@ public void loadActiveIdentityProviderServices() {
...
-                public String getAuthorizeUrl() {
-                    return null;
-                }
-
-                @Override
-                public TokenResponse getAccessToken(String code) {
-                    return null;
-                }
-            };
+            log.error("Not found any identity provider service is active.");
```
swaroopar, line 43: Log line can be -> No active identity providers found.

modules/policy/src/main/java/org/eclipse/xpanse/modules/policy/policyman/PolicyManager.java
```diff
@@ -211,7 +190,7 @@ public void deletePolicy(UUID id) {
         policyStorage.deletePolicyById(id);
     }
 
-    private void checkPolicyWhetherCreated(Csp csp, String policy) {
+    private void checkPolicyDuplicate(Csp csp, String policy) {
```
swaroopar, line 193: method name can be - `checkIfPolicyIsDuplicate`

swaroopar (changes requested):
Looks great. Just two minor text changes.

modules/policy/src/main/java/org/eclipse/xpanse/modules/policy/policyman/PolicyManager.java
baixinsui: OK

modules/security/src/main/java/org/eclipse/xpanse/modules/security/IdentityProviderManager.java
baixinsui: OK

swaroopar (approved):
LGTM

PR #997, opened 2023-11-07
Remove unused field userId

### LINKED ISSUE #991: remove userId from ServiceVo
The ServiceVo datamodel has userId field which must not be returned. This ID is also an internal ID of identity system and hence has no value in this response. 

https://github.com/eclipse-xpanse/xpanse/blob/8d06182c2b74fc77248b55db3fd5c0bf5ab50ed6/modules/models/src/main/java/org/eclipse/xpanse/modules/models/service/view/ServiceVo.java#L36

baixinsui:
fixed #991

### DIFF
--- a/modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java
+++ b/modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java
@@ -27,7 +27,6 @@ import org.eclipse.xpanse.modules.database.servicetemplate.ServiceTemplateEntity
 import org.eclipse.xpanse.modules.database.servicetemplate.ServiceTemplateStorage;
 import org.eclipse.xpanse.modules.database.utils.EntityTransUtils;
 import org.eclipse.xpanse.modules.deployment.deployers.terraform.terraformboot.model.TerraformResult;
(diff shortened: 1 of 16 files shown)


modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java
```diff
@@ -268,6 +267,7 @@ private List<DeployResourceEntity> getDeployResourceEntityList(
      * Get deployment and fill deployTask for destroy service task.
      *
      * @param deployTask the task of deploy managed service name.
+     * @param userId     the id of user who deployed the service.
      */
     public Deployment getDestroyHandler(DeployTask deployTask, String userId) {
```
swaroopar, line 272: @baixinsui Do we still need the "userId" parameter here?

swaroopar (changes requested):

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java
baixinsui: Removed

swaroopar (approved):
LGTM

PR #999, opened 2023-11-07
Add option in UI to manage ISV credentials

### LINKED ISSUE #982: Add option in UI to manage ISV credentials
- [ ] Similar to current credentials page in end user role, reuse the same components to manage ISV credentials. 
- [ ] All logic and code must be reused. only different API calls must be made based on the user role.

WangLiNaruto:
Fixed #982


swaroopar (approved):
LGTM

PR #1003, opened 2023-11-08
Update service deployer to support both end user and ISV credentials

### LINKED ISSUE #981: Update service deployer to support both end user and ISV credentials
- [ ] When the service template has `ServiceHostingType: SERVICE_VENDOR_HOSTED`, then the deployer must get the ISV credentials from the credential center and use it for deployment.
- [ ] Use it for terraform and terraform-boot deployments. 

Depends on https://github.com/eclipse-xpanse/xpanse/issues/979

WangLiNaruto:
Fixed #981


swaroopar (approved):
LGTM

PR #1011, opened 2023-11-08
allow same service to be hosted in both ways

swaroopar:
fixes https://github.com/eclipse-xpanse/xpanse/issues/1010


iskey (approved):
LGTM

PR #1012, opened 2023-11-09
Provide listDeployedServices service for ISV

### LINKED ISSUE #992: Provide listDeployedServices service for ISV
Similar to current listDeployedServices used by end user, we need a similar one for ISV Role. There by the ISV can know just statistics of how much of their services are really used. 

- [ ] Create a new API Controller for managing Service information for ISVs. 
- [ ] create new service listDeployedServicesOfIsv -> should be similar to the current listDeployedServices but we return only services belonging to the ISV of the user.
- [ ] Data model can be the same `ServiceVo`

WangLiNaruto:
Fixed #992

### DIFF
--- a/modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/deployers/terraform/TerraformBootDeployment.java
+++ b/modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/deployers/terraform/TerraformBootDeployment.java
@@ -199,7 +199,8 @@ public class TerraformBootDeployment implements Deployment {
     private Map<String, String> getEnvironmentVariables(DeployTask deployTask) {
         Map<String, String> envVariables = new HashMap<>();
         envVariables.putAll(this.deployEnvironments.getEnv(deployTask));
-        envVariables.putAll(this.deployEnvironments.getCredentialVariables(deployTask));
+        envVariables.putAll(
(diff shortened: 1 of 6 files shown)


modules/api/src/main/java/org/eclipse/xpanse/api/controllers/IsvServiceDeployApi.java
```diff
@@ -0,0 +1,100 @@
...
+            @Parameter(name = "serviceName", description = "name of the service")
+            @RequestParam(name = "serviceName", required = false) String serviceName,
+            @Parameter(name = "serviceVersion", description = "version of the service")
+            @RequestParam(name = "serviceVersion", required = false) String serviceVersion,
+            @Parameter(name = "serviceState", description = "deployment state of the service")
+            @RequestParam(name = "serviceState", required = false)
+                    ServiceDeploymentState serviceState) {
+        ServiceQueryModel query =
+                getServiceQueryModel(category, csp, serviceName, serviceVersion, serviceState);
+        return this.deployService.listDeployedServices(query);
```
swaroopar, line 73: here we must return only services belonging to the ISV to which the logged in user is belonging to.

swaroopar (changes requested):
since we have updated the URIs for existing service also, we must update the API client on the UI as well.

PR #1016, opened 2023-11-09
validate deployment plan with enabled policies

baixinsui:
fixes https://github.com/eclipse-xpanse/xpanse/issues/956

### DIFF
--- a/modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java
+++ b/modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java
@@ -27,6 +27,8 @@ import org.eclipse.xpanse.modules.database.servicetemplate.ServiceTemplateEntity
 import org.eclipse.xpanse.modules.database.servicetemplate.ServiceTemplateStorage;
 import org.eclipse.xpanse.modules.database.utils.EntityTransUtils;
 import org.eclipse.xpanse.modules.deployment.deployers.terraform.terraformboot.model.TerraformResult;
+import org.eclipse.xpanse.modules.models.policy.PolicyQueryRequest;
+import org.eclipse.xpanse.modules.models.policy.PolicyVo;
 import org.eclipse.xpanse.modules.models.service.common.enums.Csp;
 import org.eclipse.xpanse.modules.models.service.deploy.DeployRequest;
 import org.eclipse.xpanse.modules.models.service.deploy.DeployResource;
@@ -50,6 +52,7 @@ import org.eclipse.xpanse.modules.orchestrator.PluginManager;
 import org.eclipse.xpanse.modules.orchestrator.deployment.DeployResourceHandler;
 import org.eclipse.xpanse.modules.orchestrator.deployment.DeployTask;
 import org.eclipse.xpanse.modules.orchestrator.deployment.Deployment;
+import org.eclipse.xpanse.modules.policy.policyman.PolicyManager;
 import org.eclipse.xpanse.modules.security.IdentityProviderManager;
 import org.eclipse.xpanse.modules.security.common.AesUtil;
 import org.slf4j.MDC;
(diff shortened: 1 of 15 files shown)


modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java
```diff
@@ -360,6 +369,36 @@ public void purgeService(Deployment deployment, DeployTask deployTask) {
...
 
+    private void validateDeploymentWithPolicies(Deployment deployment, DeployTask deployTask) {
+
+        if (Objects.isNull(deployTask.getDeployRequest())) {
+            return;
+        }
+        String userId = deployTask.getDeployRequest().getUserId();
+        Csp csp = deployTask.getDeployRequest().getCsp();
+        if (StringUtils.isEmpty(userId) || Objects.isNull(csp)) {
+            log.info("Get user id:{} csp:{} from deploy task.", userId, csp);
```
swaroopar, line 380: Do we really need this check? Because these two fields will never be empty. 

If we need this, then the log line can be -> UserId is {} Csp is {}. Policy validation stop skipped.

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/deployers/terraform/TerraformExecutor.java
```diff
@@ -75,6 +75,16 @@ public SystemCmdResult tfPlan() {
         return executeWithVariables(new StringBuilder("terraform plan -input=false -no-color "));
     }
 
+    /**
+     * Executes terraform plan command and output.
+     *
+     * @return Returns result of SystemCmd executed.
+     */
+    public SystemCmdResult tfPlanWithOutPut() {
```
swaroopar, line 83: tfPlanWithOutPut -> tfPlanWithOutput (output should be one word.)

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/deployers/terraform/TerraformBootDeployment.java
```diff
@@ -112,9 +119,22 @@ public DeployValidationResult validate(Ocl ocl) {
                 terraformApi.validateWithScripts(getDeployWithScriptsRequest(ocl));
         DeployValidationResult result = new DeployValidationResult();
         BeanUtils.copyProperties(validate, result);
+        try {
```
swaroopar, line 122: in the line 121, we are already filling the values in `result`. Do we need still need this?

swaroopar (changes requested):
@baixinsui looks neat and simple. Just wrote some very minor comments.

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java
baixinsui: I removed this check.

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/deployers/terraform/TerraformExecutor.java
baixinsui: Renamed.

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/deployers/terraform/TerraformBootDeployment.java
baixinsui: The type of the filed `diagnostics` in these two class are different. The method `BeanUtils.copyProperties(validate, result)` could not copy this filed. I removed the ` copyProperties`  method.

swaroopar (approved):
LGTM

PR #1017, opened 2023-11-10
Provide listDeployedServices service for ISV

### LINKED ISSUE #992: Provide listDeployedServices service for ISV

WangLiNaruto:
Fixed #992


swaroopar (approved):
LGTM

PR #1019, opened 2023-11-11
block service hosting type update in template

swaroopar:
fixes https://github.com/eclipse-xpanse/xpanse/issues/1018


iskey (approved):
LGTM

PR #1030, opened 2023-11-14
Create new ISV user service -> getServiceDetailsByIdForIsv

### LINKED ISSUE #1024: Create new ISV user service -> getServiceDetailsByIdForIsv
- [ ] Create a new service getServiceDetailsByIdForIsv allowed only for ISV role. 
- [ ] Service similar to getServiceDetailsById. 
- [ ] Returns service details if the service hosting type is SERVICE_VENDOR. Otherwise return 404.
- [ ] Return the ServiceDetailVo if the user namespace is same as the service namespace. 
- [ ] If name space is not same, return access denied.

WangLiNaruto:
Fixed #1024

### DIFF
--- a/modules/api/src/main/java/org/eclipse/xpanse/api/exceptions/handler/DeploymentExceptionHandler.java
+++ b/modules/api/src/main/java/org/eclipse/xpanse/api/exceptions/handler/DeploymentExceptionHandler.java
@@ -14,6 +14,7 @@ import org.eclipse.xpanse.modules.models.service.deploy.exceptions.FlavorInvalid
 import org.eclipse.xpanse.modules.models.service.deploy.exceptions.InvalidDeploymentVariableException;
 import org.eclipse.xpanse.modules.models.service.deploy.exceptions.InvalidServiceStateException;
 import org.eclipse.xpanse.modules.models.service.deploy.exceptions.PluginNotFoundException;
+import org.eclipse.xpanse.modules.models.service.deploy.exceptions.ServiceDetailsNotAccessible;
 import org.eclipse.xpanse.modules.models.service.deploy.exceptions.ServiceNotDeployedException;
 import org.eclipse.xpanse.modules.models.service.deploy.exceptions.TerraformBootRequestFailedException;
 import org.eclipse.xpanse.modules.models.service.deploy.exceptions.TerraformExecutorException;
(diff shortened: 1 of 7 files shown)


modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java
```diff
@@ -709,4 +657,43 @@ public List<ServiceVo> listDeployedServicesOfIsv(ServiceQueryModel query) {
...
+     * Get deploy service detail by id.
+     *
+     * @param id ID of deploy service.
+     * @return serviceDetailVo
+     */
+    public ServiceDetailVo getServiceDetailsByIdForIsv(UUID id) {
+        DeployServiceEntity deployServiceEntity = getDeployServiceEntity(id);
+        ServiceHostingType serviceHostingType =
+                deployServiceEntity.getDeployRequest().getServiceHostingType();
+        if (ServiceHostingType.SERVICE_VENDOR == serviceHostingType) {
```
swaroopar, line 671: the check should be other way round. if hosting is not equal to SERVICE_VENDOR then reject the request.

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java
```diff
@@ -709,4 +657,43 @@ public List<ServiceVo> listDeployedServicesOfIsv(ServiceQueryModel query) {
...
+            String errorMsg = String.format("ServiceHostingType: %s of the Service with id %s not "
+                    + "found.", serviceHostingType, id);
+            log.error(errorMsg);
+            throw new ServiceHostingTypeNotFoundException(errorMsg);
```
swaroopar, line 675: if we want to define a new exception in this case, then ServiceDetailsNotAccessible 403 is a better name.

swaroopar (changes requested):

modules/api/src/main/java/org/eclipse/xpanse/api/exceptions/handler/DeploymentExceptionHandler.java
```diff
@@ -157,4 +158,16 @@ public Response handleVariableInvalidException(
...
                 Collections.singletonList(ex.getMessage()));
     }
+
+    /**
+     * Exception handler for ServiceDetailsNotAccessible.
+     */
+    @ExceptionHandler({ServiceDetailsNotAccessible.class})
+    @ResponseStatus(HttpStatus.FORBIDDEN)
+    @ResponseBody
+    public Response handleServiceHostingTypeNotFoundException(
```
swaroopar, line 168: @WangLiNaruto method name must be handleServiceDetailsNotAccessible

swaroopar (changes requested):

swaroopar (approved):
LGTM

PR #1032, opened 2023-11-15
Create new end user service -> getVendorHostedServiceDetailsById

### LINKED ISSUE #1025: Create new end user service -> getVendorHostedServiceDetailsById
- [ ] Create new data model -> VendorHostedServiceDetailsVo
- [ ] Extends ServiceVo.
- [ ] Similar to ServiceDetailVo but must not have resultMessage, deployResources
- [ ] Method returns only services whose service hosting type is -> SERVICE_VENDOR. If the service is called with the ID of the self-hosted service, return 404 NOT_FOUND

WangLiNaruto:
Fixed #1025

### DIFF
--- a/modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java
+++ b/modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java
@@ -44,6 +44,7 @@ import org.eclipse.xpanse.modules.models.service.query.ServiceQueryModel;
 import org.eclipse.xpanse.modules.models.service.utils.ServiceVariablesJsonSchemaValidator;
 import org.eclipse.xpanse.modules.models.service.view.ServiceDetailVo;
 import org.eclipse.xpanse.modules.models.service.view.ServiceVo;
+import org.eclipse.xpanse.modules.models.service.view.VendorHostedServiceDetailsVo;
 import org.eclipse.xpanse.modules.models.servicetemplate.DeployVariable;
 import org.eclipse.xpanse.modules.models.servicetemplate.enums.DeployerKind;
 import org.eclipse.xpanse.modules.models.servicetemplate.enums.SensitiveScope;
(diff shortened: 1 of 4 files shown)


modules/api/src/main/java/org/eclipse/xpanse/api/controllers/IsvServiceDeployApi.java
```diff
@@ -97,6 +98,22 @@ public ServiceDetailVo getServiceDetailsByIdForIsv(
...
+     */
+    @Tag(name = "Service", description = "APIs to manage the service instances")
+    @Operation(description = "Get deployed service details by id.")
+    @GetMapping(value = "/isv/service/vendor_hosted/{id}",
+            produces = MediaType.APPLICATION_JSON_VALUE)
+    @ResponseStatus(HttpStatus.OK)
+    public VendorHostedServiceDetailsVo getVendorHostedServiceDetailsById(
+            @Parameter(name = "id", description = "Task id of deployed service")
+            @PathVariable("id") String id) {
+        return this.deployService.getVendorHostedServiceDetailsById(UUID.fromString(id));
```
swaroopar, line 114: @WangLiNaruto this is a service for the end user to get details of vendor hosted service. So this must be allowed for role "user" and must be part of that controller class.

swaroopar (approved):
LGTM

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java
```diff
@@ -675,12 +676,28 @@ public ServiceDetailVo getServiceDetailsByIdForIsv(UUID id) {
...
     }
 
+    /**
+     * Get vendor hosted service detail by id.
+     *
+     * @param id ID of deploy service.
+     * @return VendorHostedServiceDetailsVo
+     */
+    public VendorHostedServiceDetailsVo getVendorHostedServiceDetailsById(UUID id) {
+        ServiceDetailVo serviceDetailVo = getServiceDetailsByIdForIsv(id);
```
swaroopar, line 694: also since this new method is for end users only, calling this existing method will not work since the end user will not have any namespace information.

swaroopar (approved):
LGTM

PR #1038, opened 2023-11-16
Handle new terraform resources (KeyPair,SecurityGroup,SecurityGroupRule)

### LINKED ISSUE #976: Extend Terraform Resource Handlers for new resources
Extend Terraform resource handlers on all plugins to handle new resources

- Security Rule
- Security Rule Group
- Key Pair

Ensure all useful properties are added to DeployResource - Properties map. For example, in case of KeyPair, the keys are important for the end user to access the resources.

baixinsui:
fixes #976 

7MMiOkheVH


plugins/scs/src/main/java/org/eclipse/xpanse/plugins/scs/enums/ScsTerraformResourceProperties.java
```diff
@@ -0,0 +1,230 @@
...
+     */
+    static class ScsKeyPairProperties extends DeployResourceProperties {
+
+        @Override
+        public DeployResourceKind getResourceKind() {
+            return DeployResourceKind.KEYPAIR;
+        }
+
+        @Override
+        public Map<String, String> getResourceProperties() {
```
swaroopar, line 180: in case of openstack and SCS, the public/private key has an error when we try to store it in DB.

image

swaroopar (changes requested):

plugins/scs/src/main/java/org/eclipse/xpanse/plugins/scs/enums/ScsTerraformResourceProperties.java
baixinsui: I updated the length of `p_value`.

swaroopar (approved):
LGTM

PR #1060, opened 2023-11-21
fix resources not destroyed on deploy failed

baixinsui:
fixes #https://github.com/eclipse-xpanse/xpanse/issues/1023


baixinsui:
The case `TerraformBoot` enabled is not fixed.

PR #1066, opened 2023-11-23
fix shown property of the resource security group rule is error

### LINKED ISSUE #1064: fix shown property of the resource security group rule is error.
The value of property `port_range_min` is error.
image

baixinsui:
fixes #1064


swaroopar (approved):
LGTM

PR #1067, opened 2023-11-23
Fix Migration service workflow blocked in the service deployment task

### LINKED ISSUE #1065: Fix Migration service workflow blocked in the service deployment task
The Migration service workflow blocked in the service deployment task. The new service instance was already but the old one cloud not be destroyed.
image
image
Caused by the `NullPointerException` :
image

baixinsui:
fixes #1065

### DIFF
--- a/modules/workflow/src/main/java/org/eclipse/xpanse/modules/workflow/handle/migrate/MigrateDeployProcess.java
+++ b/modules/workflow/src/main/java/org/eclipse/xpanse/modules/workflow/handle/migrate/MigrateDeployProcess.java
@@ -8,6 +8,7 @@ package org.eclipse.xpanse.modules.workflow.handle.migrate;
 
 import java.io.Serializable;
 import java.util.Map;
+import java.util.Objects;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import lombok.SneakyThrows;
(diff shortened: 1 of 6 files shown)


modules/workflow/src/main/java/org/eclipse/xpanse/modules/workflow/handle/migrate/MigrateDeployProcess.java
```diff
@@ -35,12 +38,12 @@ public class MigrateDeployProcess implements Serializable, JavaDelegate {
 
     @Autowired
     public void setRuntimeService(RuntimeService runtimeService) {
-        this.runtimeService = runtimeService;
+        MigrateDeployProcess.runtimeService = runtimeService;
```
swaroopar, line 41: @baixinsui I believe it better to keep all autowired beans as non-static. I had actually given the same comment when the original PR was opened as well. 
Could you please make the change to MigrateDeployProcess and MigrateDestroyProcess?

swaroopar (changes requested):

modules/workflow/src/main/java/org/eclipse/xpanse/modules/workflow/handle/migrate/MigrateDeployProcess.java
baixinsui: @swaroopar The workflow engine couldn't instantiate class after I change the static class to autowired bean.
image

PR #1073, opened 2023-11-27
fix: deployed resources could not be destroyed when deployment failed

### LINKED ISSUE #1023: When deployment failed the deployed resources will not be destroyed.
When deployment failed  caused  by the terraform command error,  the `resources` in the `deployResult`  is always empty  and the deployed resources will not be destroyed. We should remove judgment condition and always execute the destroy when the deployment failed.  The logic in terraform-boot should be checked  also.

image
image
image
image

baixinsui:
fixes #1023 
related https://github.com/eclipse-xpanse/terraform-boot/pull/32


modules/workflow/src/main/java/org/eclipse/xpanse/modules/workflow/handle/migrate/MigrateExportDataProcess.java
```diff
@@ -26,7 +26,7 @@ public class MigrateExportDataProcess implements Serializable, JavaDelegate {
 
     @Autowired
     public void setRuntimeService(RuntimeService runtimeService) {
-        this.runtimeService = runtimeService;
+        MigrateExportDataProcess.runtimeService = runtimeService;
```
swaroopar, line 29: please make these fields as non-static and autowire via constructor

modules/workflow/src/main/java/org/eclipse/xpanse/modules/workflow/handle/migrate/MigrateImportDataProcess.java
```diff
@@ -26,7 +26,7 @@ public class MigrateImportDataProcess implements Serializable, JavaDelegate {
 
     @Autowired
     public void setRuntimeService(RuntimeService runtimeService) {
-        this.runtimeService = runtimeService;
+        MigrateImportDataProcess.runtimeService = runtimeService;
```
swaroopar, line 29: please make these fields as non-static and autowire via constructor

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/deployers/terraform/TerraformExecutor.java
```diff
@@ -114,9 +114,7 @@ private SystemCmdResult executeWithVariables(StringBuilder command) {
         createVariablesFile();
         command.append(" -var-file=");
         command.append(VARS_FILE_NAME);
-        SystemCmdResult systemCmdResult = execute(command.toString());
-        // cleanUpVariablesFile();
```
swaroopar, line 118: i think the call to cleanupVariablesFile was commented out by mistake in the previous PRs.. can you please add this back?

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java
```diff
@@ -278,114 +348,87 @@ private List<DeployResourceEntity> getDeployResourceEntityList(
...
-        } else {
-            throw new InvalidServiceStateException(
-                    String.format("Service %s is not in the state allowed for purging.",
-                            deployServiceEntity.getId()));
+    private String getStoredStateContent(DeployServiceEntity deployServiceEntity) {
+        if (Objects.isNull(deployServiceEntity)
+                || CollectionUtils.isEmpty(deployServiceEntity.getPrivateProperties())
+                || StringUtils.isEmpty(
+                deployServiceEntity.getPrivateProperties().get(STATE_FILE_NAME))) {
+            throw new IllegalStateException("Can't find valid state context in stored service.");
```
swaroopar, line 406: Please throw any of our custom exception

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java
```diff
@@ -477,77 +543,70 @@ public VendorHostedDeployedServiceDetails getVendorHostedServiceDetailsByIdForEn
...
-            deployResourceStorage.deleteByDeployServiceId(deployServiceEntity.getId());
-        } else {
-            deployServiceEntity.setServiceDeploymentState(ServiceDeploymentState.DESTROY_FAILED);
+        DeployServiceEntity updatedDeployServiceEntity =
+                flushDestroyServiceEntity(destroyResult, deployServiceEntity);
+        if (ServiceDeploymentState.DESTROY_SUCCESS
+                == updatedDeployServiceEntity.getServiceDeploymentState()) {
+            Deployment deployment = getDeployment(updatedDeployServiceEntity.getDeployRequest()
+                    .getOcl().getDeployment().getKind());
+            deployment.deleteTaskWorkspace(taskId);
```
swaroopar, line 595: in case of terraform-boot, the workspace deletion is handled by terraform-boot. Do we really need this here?

swaroopar (changes requested):

modules/workflow/src/main/java/org/eclipse/xpanse/modules/workflow/handle/migrate/MigrateExportDataProcess.java
baixinsui: The workflow engine couldn't instantiate class after I change the static class to autowired bean.
image

modules/workflow/src/main/java/org/eclipse/xpanse/modules/workflow/handle/migrate/MigrateImportDataProcess.java
baixinsui: Same as above

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/deployers/terraform/TerraformExecutor.java
baixinsui: I restored it.

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java
baixinsui: Updated.

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java
baixinsui: It's  unnecessary and I removed it.

swaroopar (approved):
LGTM

PR #1075, opened 2023-11-27
add custom requestId header option to terraform-boot 

### LINKED ISSUE #1047: add custom requestId header option to terraform-boot
To allow request tracing from calling systems, terraform-boot must have an option to allow calling systems to provide their custom request Ids. 

1. Add new optional custom HTTP header - X-Custom-RequestId
2. If this value is present, insert MDC key from the new header. Otherwise generate new unique UUID -> https://github.com/eclipse-xpanse/terraform-boot/blob/main/src/main/java/org/eclipse/xpanse/terraform/boot/logging/CustomRequestIdGenerator.java#L21
3. If the header is present, then create the terraform directory using this value. 
4. Ensure the openApi/swagger-ui displays this header as optional correctly.

WangLiNaruto:
Fixed #1047
image
image


swaroopar (approved):
LGTM

PR #1081, opened 2023-11-30
When TerraformBoot enables authentication, Xpanse calls its APIs alwa…

### LINKED ISSUE #1043: When TerraformBoot enables authentication, Xpanse calls its APIs always return  401 unauthenticated exception.
When TerraformBoot enables authentication, Xpanse calls its APIs always return  401 unauthenticated exception.
image
image

WangLiNaruto:
Fixed #1043
related to https://github.com/eclipse-xpanse/terraform-boot/pull/37

image
image

image
image


modules/models/src/main/java/org/eclipse/xpanse/modules/models/security/model/CurrentUserInfoCacheHolder.java
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * User information cache.
+ */
+@Profile("zitadel")
+public class CurrentUserInfoCacheHolder {
+
+    public static final String CURRENT_USERINFO_CACHE_NAME = "CurrentUserinfo";
+
+    private static final Map<String, CurrentUserInfo> USERINFO_CACHE
```
baixinsui, line 21: Using this solution, the map seems always store only the latest putted user . Please test if the user information obtained in the sub thread is correct if multiple users access it simultaneously. If not correct, you can try to use the `ThreadLocal<CurrentUserInfo>` to obtain the user information in sub threads.

modules/security/src/main/java/org/eclipse/xpanse/modules/security/zitadel/ZitadelIdentityProviderService.java
```diff
@@ -172,6 +176,11 @@ public CurrentUserInfo getCurrentUserInfo() {
                     }
                 }
             }
+            OAuth2Token token =
+                    ((AbstractOAuth2TokenAuthenticationToken<?>) authentication).getToken();
+            currentUserInfo.setToken(
+                    ZitadelOauth2Constant.USER_TOKEN_PREFIX + token.getTokenValue());
+            CurrentUserInfoCacheHolder.getSingleton().setCurrentUserInfo(currentUserInfo);
```
baixinsui, line 183: If current solution using `ConcurrentHashMap` works well,  please move the set action to `IdentityProviderManager`. If change to use the new solution `ConcurrentHashMap`, delete the set acition.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/security/model/CurrentUserInfoCacheHolder.java
```diff
@@ -0,0 +1,44 @@
...
+            = new ConcurrentHashMap(0);
+
```
baixinsui, line 23: The initial capacity of map is `0`?

baixinsui (changes requested):
Some comment need to check

baixinsui (approved):
LGTM

modules/models/src/main/java/org/eclipse/xpanse/modules/models/security/model/CurrentUserInfoHolder.java
```diff
@@ -0,0 +1,38 @@
...
+
+package org.eclipse.xpanse.modules.models.security.model;
+
+import java.util.Objects;
+import org.springframework.context.annotation.Profile;
+
+/**
+ * User information holder.
+ */
+@Profile("zitadel")
```
swaroopar, line 15: do we need this profile annotation as this class is not a spring bean?

please also consider that if Zitdel profile is not enabled .. then there is no token available. So in this case, no auth header must be sent.

swaroopar (changes requested):

swaroopar (approved):
LGTM

PR #1120, opened 2023-12-11
Add APIs for ISV user managing policies belong to service template

baixinsui:
fixes https://github.com/eclipse-xpanse/xpanse/issues/1069


modules/api/src/main/java/org/eclipse/xpanse/api/controllers/ServicePolicyManageApi.java
```diff
@@ -0,0 +1,136 @@
...
+
+
+/**
+ * REST interface methods for service managing policies of the cloud service provider.
+ */
+@Slf4j
+@RestController
+@RequestMapping("/xpanse")
+@CrossOrigin
+@Secured({ROLE_ISV})
```
swaroopar, line 46: just to keep it uniform with other controllers, we can add ADMIN_ROLE as well here.

modules/policy/src/main/java/org/eclipse/xpanse/modules/policy/policyman/PolicyManager.java
```diff
@@ -44,16 +52,12 @@
 import org.springframework.web.client.RestClientException;
 
 /**
- * The service of policy-man.
+ * The service for managing policies.
  */
 @Slf4j
 @Component
 public class PolicyManager {
```
swaroopar, line 59: this class has become too big. Would be good if we can split this into UserPolicyManager and ServicePolicyManager. We can move all common methods to another common class.

modules/policy/src/main/java/org/eclipse/xpanse/modules/policy/policyman/PolicyManager.java
```diff
@@ -69,6 +73,167 @@ public class PolicyManager {
...
+        return conventToServicePolicy(existingEntity);
+    }
+
+    /**
+     * Delete the policy belonging to the registered service template.
+     *
+     * @param id the id of policy.
+     */
+    public void deleteServicePolicy(UUID id) {
+        ServicePolicyEntity existingEntity = getServicePolicyEntity(id, "delete");
```
swaroopar, line 167: this variable is not used anywhere.

modules/api/src/main/java/org/eclipse/xpanse/api/controllers/ServicePolicyManageApi.java
```diff
@@ -0,0 +1,136 @@
...
+public class ServicePolicyManageApi {
+
+    @Resource
+    private PolicyManager policyManager;
+
+    @Resource
+    private IdentityProviderManager identityProviderManager;
```
swaroopar, line 53: looks like this bean is not used in this class.

swaroopar (changes requested):
Looks neat. Just very minor changes requested.

modules/api/src/main/java/org/eclipse/xpanse/api/controllers/ServicePolicyManageApi.java
baixinsui: Updated as suggested.

modules/policy/src/main/java/org/eclipse/xpanse/modules/policy/policyman/PolicyManager.java
baixinsui: Updated as suggested.

modules/api/src/main/java/org/eclipse/xpanse/api/controllers/ServicePolicyManageApi.java
baixinsui: Updated as suggested.

modules/policy/src/main/java/org/eclipse/xpanse/modules/policy/policyman/PolicyManager.java
baixinsui: Removed this variable.

swaroopar (approved):
LGTM

PR #1155, opened 2023-12-15
add service started and service stopped time

### LINKED ISSUE #1133: add service started and service stopped time
As part of ServiceVO and to the corresponding table -> add two new timestamps
1. LastStartedAt -> To be updated when the service is started.
2. LastStoppedAt -> To be updated whenever service is stopped.

WangLiNaruto:
Fixed #1133
image
image
image
image

### DIFF
--- a/modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java
+++ b/modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java
@@ -941,6 +941,7 @@ public class DeployService {
             deployServiceEntity.setServiceState(ServiceState.STARTING);
             deployServiceStorage.storeAndFlush(deployServiceEntity);
             if (start(deployServiceEntity)) {
+                deployServiceEntity.setLastStartedAt(OffsetDateTime.now());
                 deployServiceEntity.setServiceState(ServiceState.RUNNING);
             } else {
                 deployServiceEntity.setServiceState(ServiceState.STARTING_FAILED);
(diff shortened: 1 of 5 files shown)


modules/models/src/main/java/org/eclipse/xpanse/modules/models/service/view/DeployedService.java
```diff
@@ -100,4 +101,12 @@ public class DeployedService {
     @JsonFormat(pattern = "yyyy-MM-dd HH:mm:ss XXX")
     @JsonSerialize(using = OffsetDateTimeSerializer.class)
     private OffsetDateTime lastModifiedTime;
+
+    @Schema(description = "Time of start service.")
+    @JsonFormat(pattern = "yyyy-MM-dd HH:mm:ss XXX", timezone = "GMT+8")
```
swaroopar, line 106: we must use the same format as in the field above `lastModifiedTime`

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java
```diff
@@ -970,6 +972,7 @@ public DeployedService stopService(UUID id) {
         try {
             validateStopDeployServiceEntity(deployServiceEntity);
             deployServiceEntity.setServiceState(ServiceState.STOPPING);
+            deployServiceEntity.setLastStoppedAt(new Timestamp(System.currentTimeMillis()));
```
swaroopar, line 975: this must be inside the if loop in line 977?

swaroopar (changes requested):

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java
```diff
@@ -939,6 +940,7 @@ public DeployedService startService(UUID id) {
         try {
             validateStartDeployServiceEntity(deployServiceEntity);
             deployServiceEntity.setServiceState(ServiceState.STARTING);
+            deployServiceEntity.setLastStartedAt(new Timestamp(System.currentTimeMillis()));
```
swaroopar, line 943: this must be inside loop 945?

swaroopar (approved):
LGTM

PR #1173, opened 2023-12-21
Replace HttpClient with HuaweiCloud SDK in FlexibleEngine Plugin

### LINKED ISSUE #1168: replace http client from java-sdk-core in flexbileEngine plugin
Currently the flexible engine plugin uses the http client from java-sdk-core and this must be replaced and we must use our own http client. 

- [ ] create http client using spring rest client. 
- [ ] write the request signer logic. 
- [ ] remove all dependencies to java-sdk-core from SBOM.

baixinsui:
fixes #1168


plugins/flexibleengine/src/main/java/org/eclipse/xpanse/plugins/flexibleengine/common/FlexibleEngineClient.java
```diff
@@ -0,0 +1,176 @@
...
+            return false;
+        }
+        int statusCode = ((ServiceResponseException) ex).getHttpStatusCode();
+        return statusCode == ERROR_CODE_TOO_MANY_REQUESTS
+                || statusCode == ERROR_CODE_INTERNAL_SERVER_ERROR;
+    }
+
+    private HcClient getHcClient(ICredential credential, String servicePrefix, String regionName) {
+        String endpoint = PROTOCOL_HTTPS + servicePrefix + regionName + ENDPOINT_SUFFIX;
+        Region region = new Region(regionName, endpoint);
```
swaroopar, line 93: this region variable is not used anywhere.

plugins/flexibleengine/src/main/java/org/eclipse/xpanse/plugins/flexibleengine/common/FlexibleEngineClient.java
```diff
@@ -0,0 +1,176 @@
...
+                    .withProjectId(projectId);
+        }
+        return credential;
+    }
+
+    private String getProjectId(ICredential credential, String regionName) {
+        String projectId = null;
+        try {
+            String endpoint = PROTOCOL_HTTPS + IAM_ENDPOINT_PREFIX + regionName + ENDPOINT_SUFFIX;
+            Region region = new Region(regionName, endpoint);
```
swaroopar, line 116: same here. Not used. Please check other methods too in this class.

plugins/huaweicloud/pom.xml
```diff
@@ -18,6 +18,19 @@
         <relativePath>../pom.xml</relativePath>
     </parent>
 
+    <repositories>
+        <repository>
+            <id>HuaweiCloudSDK</id>
+            <url>https://repo.huaweicloud.com/repository/maven/huaweicloudsdk/</url>
```
swaroopar, line 24: as far i see, we do need this internal repository. All huaweicloud SDKs must be available in the maven central already. 
Please check this in the flexibleengine pom as well.

swaroopar (changes requested):

plugins/flexibleengine/src/main/java/org/eclipse/xpanse/plugins/flexibleengine/common/FlexibleEngineClient.java
baixinsui: Removed.

plugins/flexibleengine/src/main/java/org/eclipse/xpanse/plugins/flexibleengine/common/FlexibleEngineClient.java
baixinsui: Removed.

plugins/huaweicloud/pom.xml
baixinsui: Removed the internal repository.

swaroopar (approved):
LGTM

PR #1182, opened 2023-12-22
fix - handle when tfstate file is not found

swaroopar:
fixes https://github.com/eclipse-xpanse/xpanse/issues/1181


iskey (approved):
LGTM

PR #1192, opened 2023-12-24
Feature/ignore dependency

swaroopar:
Activiti latest version is has a bug which causes the server startup to fail due to a faulty create table statement. So this version must be ignored by the dependabot


iskey (approved):
LGTM

PR #1196, opened 2023-12-26
generate terraform plan only if there is any policy to be validated

### LINKED ISSUE #1190: generate terraform plan only if there is any policy to be validated
Currently we are generating terraform plan always and after that checking if there is any policy to be validated - https://github.com/eclipse-xpanse/xpanse/blob/main/modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java#L210

This must be changed. We must generate the JSON plan only if there is any end user policy or service policy to be validated.

baixinsui:
fixes #1190


swaroopar (approved):
LGTM

PR #1198, opened 2023-12-26
remove async threads when terraform-boot is activated

### LINKED ISSUE #1078: remove async threads when terraform-boot is activated
In order to avoid unnecessary handling of services and to have better tracing, we must remove the async thread created when terraform-boot is activated. 

All steps including the deployment and destroy request must be handled with in the sync http request.

WangLiNaruto:
Fixed #1078
related to eclipse-xpanse/terraform-boot#72
image

### DIFF
--- a/modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java
+++ b/modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java
@@ -341,8 +341,7 @@ public class DeployService {
      * @param deployment deployment
      * @param deployTask deployTask
      */
-    @Async("taskExecutor")
-    public void asyncDeployService(Deployment deployment, DeployTask deployTask) {
+    public void deployService(Deployment deployment, DeployTask deployTask) {
         deploy(deployment, deployTask);
     }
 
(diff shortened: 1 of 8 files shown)


modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java
```diff
@@ -657,77 +656,6 @@ public VendorHostedDeployedServiceDetails getVendorHostedServiceDetailsByIdForEn
                 deployServiceEntity);
     }
 
-    /**
-     * Callback method after deployment is complete.
-     */
-    public void deployCallback(String taskId, TerraformResult result) {
```
swaroopar, line 663: we need this one too. There is no change expected here.

modules/api/src/main/java/org/eclipse/xpanse/api/controllers/WebhookApi.java
```diff
@@ -1,70 +0,0 @@
-/*
```
swaroopar, line 1: we need these methods still.

swaroopar (changes requested):

swaroopar (changes requested):
this looks definitely good. 
There are also 3 more places in this class where we have @async annotated method. can we get rid of them too please?

WangLiNaruto:
> this looks definitely good. There are also 3 more places in this class where we have @async annotated method. can we get rid of them too please?


Sure, the other @Async methods I saw are about purge or migrate. Don’t they need to be executed asynchronously? If asynchronous is not needed, I feel like I should submit a new PR.

swaroopar:
> > this looks definitely good. There are also 3 more places in this class where we have @async annotated method. can we get rid of them too please?
> 
> Sure, the other @async methods I saw are about purge or migrate. Don’t they need to be executed asynchronously? If asynchronous is not needed, I feel like I should submit a new PR.

@WangLiNaruto yes, even purge and migrate must work similar to deploy and destroy. We must start async step only in the terraform-local

swaroopar (approved):
LGTM

PR #1207, opened 2023-12-31
fix - remove JVM level environment var OS_AUTH_URL

swaroopar:
fixes https://github.com/eclipse-xpanse/xpanse/issues/1206


iskey (approved):
LGTM

Issue #1221, opened 2024-01-05
openstack plugin - metrics adapter to use deploy variables


swaroopar:
As the openstack auth URL is part of the deploy variable, the metrics adapter must also use the value from the deploy variable to fetch metrics. 

Currently it is expecting the AUTH URL in the environment variable which is not valid anymore.

PR #1222, opened 2024-01-05
use deploy variables in metrics adapter

swaroopar:
fixes https://github.com/eclipse-xpanse/xpanse/issues/1221

@iskey the changes might look big. But just the methods are just split. No big functional changees.


iskey:
> fixes #1221
> 
> @iskey the changes might look big. But just the methods are just split. No big functional changees.

ok, got it, It looks good to me.

iskey (approved):
/LGTM

PR #1241, opened 2024-01-15
generate activiti processInstanceId as UUID and use this as Migration  Request ID

### LINKED ISSUE #1234: generate activiti processInstanceId as UUID and use this as Migration Request ID
In order to have uniformity with other APIs in xpanse, the activiti processInstanceId must be of type UUID instance instead of the default ID.

- [ ] Generate UUID as processInstanceId
- [ ] Return this ID as the migration request ID in the `migrate` method instead of the new ID.

Reference - https://www.activiti.org/5.x/userguide/#advanced.uuid.generator

WangLiNaruto:
Fixed #1234 
image

### DIFF
--- a/modules/workflow/pom.xml
+++ b/modules/workflow/pom.xml
@@ -59,6 +59,12 @@
       <version>${mysql.version}</version>
       <scope>runtime</scope>
     </dependency>
+    <dependency>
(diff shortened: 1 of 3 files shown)


swaroopar (approved):
LGTM

PR #1268, opened 2024-01-22
implement flavor specific service policies

### LINKED ISSUE #1261: implement flavor specific service policies
Currently service policies are at service level. This must be extend to allow CSPs to have common policies at service level and additional policies at flavor level. 

- [ ] Extend the data model to provide an optional "flavor_name" field for service policies. 
- [ ] if flavor name provided during policy creation, validate if the flavor name is valid, if not return the existing `FlavorInvalidException` exception. 
- [ ] If flavor name is not provided, then the policy is always executed during service deployment. 
- [ ] If flavor name is provided, then this policy is validated only if the service was ordered for that specific flavor.

baixinsui:
fixes #1261

### DIFF
--- a/modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/PolicyValidator.java
+++ b/modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/PolicyValidator.java
@@ -7,6 +7,7 @@
 package org.eclipse.xpanse.modules.deployment;
 
 import jakarta.annotation.Resource;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
@@ -47,8 +48,8 @@ public class PolicyValidator {
     private List<ServicePolicy> getServicePolicies(UUID serviceTemplateId) {
         ServiceTemplateEntity existedServiceTemplate =
                 serviceTemplateStorage.getServiceTemplateById(serviceTemplateId);
-        if (Objects.nonNull(existedServiceTemplate)
-                && Objects.nonNull(existedServiceTemplate.getServicePolicyList())) {
+        if (Objects.nonNull(existedServiceTemplate) && Objects.nonNull(
+                existedServiceTemplate.getServicePolicyList())) {
             return existedServiceTemplate.getServicePolicyList().stream()
                     .filter(servicePolicyEntity -> servicePolicyEntity.getEnabled()
                             && StringUtils.isNotBlank(servicePolicyEntity.getPolicy()))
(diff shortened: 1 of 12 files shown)


modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/PolicyValidator.java
```diff
@@ -88,35 +88,46 @@ public void validateDeploymentWithPolicies(DeployTask deployTask) {
...
+
+            List<String> flavorPolicyList = servicePolicies.stream().filter(servicePolicy ->
+                            StringUtils.equals(flavorName, servicePolicy.getFlavorName())
+                                    && StringUtils.isNotBlank(flavorName))
                     .map(ServicePolicy::getPolicy).toList();
+
+            if (!CollectionUtils.isEmpty(flavorPolicyList)) {
+                log.info("Evaluate deployment plan with service policies for flavor: {}",
+                        flavorName);
+                servicePolicyList = flavorPolicyList;
```
swaroopar, line 116: if policies without flavor and  policies with flavor exist, then both the policies are executed. 
Sorry, I think the requirement was not clear in the user story. 
We must execute the policies without a flavor always.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/policy/servicepolicy/ServicePolicy.java
```diff
@@ -45,6 +45,9 @@ public class ServicePolicy {
     @Schema(description = "The id of registered service template which the policy belongs to.")
     private UUID serviceTemplateId;
 
+    @Schema(description = "The flavor name which the policy belongs to.")
```
swaroopar, line 48: please extend the description here to -

The flavor name which the policy belongs to. If a flavor is not provided, then the policy will be executed for during service deployment of all flavors.

swaroopar (changes requested):

modules/models/src/main/java/org/eclipse/xpanse/modules/models/policy/servicepolicy/ServicePolicyCreateRequest.java
```diff
@@ -26,6 +26,10 @@ public class ServicePolicyCreateRequest {
     @Schema(description = "The id of registered service template which the policy belongs to.")
     private UUID serviceTemplateId;
 
+
+    @Schema(description = "The flavor name which the policy belongs to.")
```
swaroopar, line 30: please extend the description here to -

The flavor name which the policy belongs to. If a flavor is not provided, then the policy will be executed for during service deployment of all flavors.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/policy/servicepolicy/ServicePolicy.java
baixinsui: Updated.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/policy/servicepolicy/ServicePolicyCreateRequest.java
baixinsui: Updated.

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/PolicyValidator.java
baixinsui: Updated.

swaroopar (approved):
LGTM

PR #1276, opened 2024-01-23
threads in purge request

### LINKED ISSUE #1213: remove async threads in purge request
similar to deploy and destroy and methods, purge request must not be executed as a separate thread by default. 
only terraform-local must start its down thread to execute any purge tasks.

WangLiNaruto:
Fixed #1213 
image


swaroopar (approved):
LGTM

PR #1279, opened 2024-01-24
fix - condition for huaweicloud async request retry

swaroopar:
fixes https://github.com/eclipse-xpanse/xpanse/issues/1278


iskey (approved):
LGTM

PR #1281, opened 2024-01-24
New API method to get migration order details

### LINKED ISSUE #1236: New API method to get migration order details
- [ ] Create new method getMigrationOrderDetails with input as MigrationRequestId
- [ ] Return all information from the SERVICE_MIGRATION table.

WangLiNaruto:
Fixed #1236 
image
image

### DIFF
--- a/modules/models/src/main/java/org/eclipse/xpanse/modules/models/workflow/migrate/view/ServiceMigrationDetails.java
+++ b/modules/models/src/main/java/org/eclipse/xpanse/modules/models/workflow/migrate/view/ServiceMigrationDetails.java
@@ -1,0 +1,52 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * SPDX-FileCopyrightText: Huawei Inc.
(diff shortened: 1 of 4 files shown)


modules/models/src/main/java/org/eclipse/xpanse/modules/models/workflow/migrate/view/ServiceMigrationDetails.java
```diff
@@ -0,0 +1,58 @@
...
+    private UUID newServiceId;
+
+    @NotNull
+    @Schema(description = "The status of the service migration")
+    private MigrationStatus migrationStatus;
+
+    @NotNull
+    @NotBlank
+    @Schema(description = "The ID of the service migration user")
+    private String userId;
```
swaroopar, line 44: we must not return this.

swaroopar (changes requested):

swaroopar (approved):
LGTM

PR #1294, opened 2024-01-25
migrate workflow does not set "DESTROYING" status

### LINKED ISSUE #1287: migrate workflow does not set "DESTROYING" status
the "StartDestroy" step in the service migration workflow, does not set the state "DESTROYING" intially. 
Only the final states DESTROY_SUCCESSFUL or DESTROY_FAILED are set.

WangLiNaruto:
Fixed #1287


swaroopar (approved):
LGTM

PR #1304, opened 2024-01-26
add new component to show workflow tasks assigned to user

### LINKED ISSUE #1237: update queryTodoTasks and queryDoneTasks methods with authorization
For both  queryTodoTasks and queryDoneTasks  API methods, 

- [ ] throw 403 AccessDeniedException exception when no user ID is found in the request context. 
- [ ] When the userId does not match the user who requested the migration request, also throw AccessDeniedException

WangLiNaruto:
Fixed #1237
image
image
image

### DIFF
--- a/modules/api/src/main/java/org/eclipse/xpanse/api/controllers/WorkFlowApi.java
+++ b/modules/api/src/main/java/org/eclipse/xpanse/api/controllers/WorkFlowApi.java
@@ -19,6 +19,7 @@ import java.util.Map;
 import java.util.Optional;
 import lombok.extern.slf4j.Slf4j;
 import org.eclipse.xpanse.modules.deployment.migration.consts.MigrateConstants;
+import org.eclipse.xpanse.modules.models.workflow.TaskStatus;
 import org.eclipse.xpanse.modules.models.workflow.WorkFlowTask;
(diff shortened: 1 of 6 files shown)


modules/api/src/main/java/org/eclipse/xpanse/api/controllers/WorkFlowApi.java
```diff
@@ -74,6 +74,18 @@ public List<WorkFlowTask> queryDoneTasks() {
         return workflowUtils.doneTasks(userIdOptional.orElse(null));
     }
 
+    /**
+     * Query all tasks of the given user.
+     */
+    @Tag(name = "Workflow", description = "APIs to manage the Workflow")
+    @Operation(description = "Query all tasks of the given user")
+    @GetMapping(value = "/workflow/tasks", produces = MediaType.APPLICATION_JSON_VALUE)
```
swaroopar, line 82: we can have URI as /workflow/tasks/all

also looking at these 3 methods, we can also consider combining them into one. /workflow/tasks/${status} where status is optional.

swaroopar (changes requested):

swaroopar (approved):
LGTM

PR #1326, opened 2024-01-31
fix error state when rollback success

### LINKED ISSUE #1301: when deployment failure and successful rollback set an error state `DESTROY_SUCCESS`
When the deployment failures and then the rollback successes, currently shown state 'DESTROY_SUCCESS' which is misleading, and the details could not be shown.
image
image

baixinsui:
fixes #1301 
related to https://github.com/eclipse-xpanse/terraform-boot/pull/79
image


modules/models/src/main/java/org/eclipse/xpanse/modules/models/service/deploy/enums/DestroyScenario.java
```diff
@@ -0,0 +1,48 @@
...
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonValue;
+import org.apache.commons.lang3.StringUtils;
+import org.eclipse.xpanse.modules.models.common.exceptions.UnsupportedEnumValueException;
+
+/**
+ * Destroy Scenario. The destroy scenario is used to set into TerraformResult and sent back to the
+ * client which sent the destroy request in what scenario.
+ */
+public enum DestroyScenario {
```
swaroopar, line 17: if this enum is not exposed to customer, we must move this to the same place where "DeployTask" class exists.

swaroopar (changes requested):

modules/models/src/main/java/org/eclipse/xpanse/modules/models/service/deploy/enums/DestroyScenario.java
baixinsui: Moved.

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/deployers/opentofu/OpenTofuDeployment.java
```diff
@@ -51,14 +48,13 @@ public class OpenTofuDeployment implements Deployer {
 
     public static final String VERSION_FILE_NAME = "version.tf";
     public static final String SCRIPT_FILE_NAME = "resources.tf";
-    public static final String STATE_FILE_NAME = "terraform.tfstate";
+    public static final String STATE_FILE_NAME = "openTofu.tfstate";
```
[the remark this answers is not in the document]
baixinsui: Restored.

swaroopar (approved):
LGTM

PR #1341, opened 2024-02-04
details option for vendor hosted services to be optionally enabled

### LINKED ISSUE #1171: details option for vendor hosted services to be optionally enabled
For end user role, the details option must be only conditionally enabled. Currently it is always enabled and we just display empty popup as below. 

- [x] It must be enabled when service details are available. 
- [x] It must be disabled if service details are not available and provide the following text on mouse over -> 'Please contact the service provider'

image

WangLiNaruto:
Fixed #1171


swaroopar (changes requested):
@WangLiNaruto why do we need this new service? Is there an issue with the existing service?

swaroopar (approved):
LGTM

PR #1359, opened 2024-02-08
update service migrate ui

### LINKED ISSUE #1283: update service migrate ui
Modify the page problem during migration. Due to the modification of the return value of the migrate interface, when querying under deployment status polling during migration, the process id query is used, and the following error occurs.
image

WangLiNaruto:
Fixed #1283


swaroopar (approved):
LGTM

PR #1373, opened 2024-02-14
use oauth authentication http client for all API calls

swaroopar:
related to https://github.com/eclipse-xpanse/xpanse/issues/1336

@iskey the number of changed files is more due to package movement. Actual code changes are not more.


iskey (approved):
LGTM

PR #1379, opened 2024-02-19
Link service template ID and the DeployedService data.

### LINKED ISSUE #1352: Link service template ID and the DeployedService data
Currently there is no link between the service template ID and the deployed service. Instead we are storing the OCL data again for each deployed service which is duplicated data. 

- [ ] Remove the hidden Ocl from DeployRequestBase -> https://github.com/eclipse-xpanse/xpanse/blob/main/modules/models/src/main/java/org/eclipse/xpanse/modules/models/service/deploy/DeployRequestBase.java#L93
- [ ] add a mandatory SERVICE_TEMPLATE_ID column here -> https://github.com/eclipse-xpanse/xpanse/blob/main/modules/models/src/main/java/org/eclipse/xpanse/modules/models/service/deploy/DeployRequestBase.java#L93
- [ ] Extend DeployedService and DeployedServiceDetails to have the ServiceTemplateId in the response

Alice1319:
fixes #1352

### DIFF
--- a/modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java
+++ b/modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java
@@ -101,11 +101,14 @@ public class DeployService {
         // Create new deploy task by deploy request.
         DeployTask deployTask = new DeployTask();
         deployTask.setId(UUID.randomUUID());
-        deployRequest.setOcl(serviceTemplate.getOcl());
         deployTask.setDeployRequest(deployRequest);
         deployTask.setNamespace(serviceTemplate.getNamespace());
         deployTask.setOcl(serviceTemplate.getOcl());
-        deployTask.setServiceTemplateId(serviceTemplate.getId());
+        if (Objects.nonNull(serviceTemplate.getId())) {
+            deployTask.setServiceTemplateId(serviceTemplate.getId());
+        } else {
+            throw new ServiceTemplateNotRegistered("service template id can't be null.");
+        }
         return deployTask;
     }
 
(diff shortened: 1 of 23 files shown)


modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java
```diff
@@ -101,11 +101,12 @@ public DeployTask createNewDeployTask(DeployRequest deployRequest) {
         // Create new deploy task by deploy request.
         DeployTask deployTask = new DeployTask();
         deployTask.setId(UUID.randomUUID());
-        deployRequest.setOcl(serviceTemplate.getOcl());
         deployTask.setDeployRequest(deployRequest);
         deployTask.setNamespace(serviceTemplate.getNamespace());
         deployTask.setOcl(serviceTemplate.getOcl());
-        deployTask.setServiceTemplateId(serviceTemplate.getId());
+        if (Objects.nonNull(serviceTemplate.getId())) {
```
swaroopar, line 107: if this is null, then we must throw an exception. This cannot be null.

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java
```diff
@@ -134,6 +135,9 @@ private DeployServiceEntity storeNewDeployServiceEntity(DeployTask deployTask) {
         entity.setDeployResourceList(new ArrayList<>());
         entity.setNamespace(deployTask.getNamespace());
         entity.setServiceDeploymentState(ServiceDeploymentState.DEPLOYING);
+        if (Objects.nonNull(deployTask.getServiceTemplateId())) {
```
swaroopar, line 138: same here. it cannot be null

swaroopar (changes requested):

swaroopar (approved):
LGTM

PR #1381, opened 2024-02-19
Add interface for querying all registered service templates for the role CSP

### LINKED ISSUE #1378: create role CSP and add listAllRegisteredServices service
- [ ] Create a new role CSP
- [ ] add role to IAM scripts
- [ ] add role to xpanse
- [ ] create a service GetAllRegisteredServices  -> list all services similar to `listServiceTemplates`. But return without any namespace checks. 
- [ ] Enable the service for CSP and admin roles.

baixinsui:
fixes #1378 
image

related https://github.com/eclipse-xpanse/xpanse-iam/pull/17

### DIFF
--- a/modules/api/src/main/java/org/eclipse/xpanse/api/controllers/CspServiceTemplateApi.java
+++ b/modules/api/src/main/java/org/eclipse/xpanse/api/controllers/CspServiceTemplateApi.java
@@ -1,0 +1,88 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * SPDX-FileCopyrightText: Huawei Inc.
+ *
+ */
+
+package org.eclipse.xpanse.api.controllers;
+
+
+import static org.eclipse.xpanse.modules.security.common.RoleConstants.ROLE_ADMIN;
+import static org.eclipse.xpanse.modules.security.common.RoleConstants.ROLE_CSP;
+
+import io.swagger.v3.oas.annotations.Operation;
+import io.swagger.v3.oas.annotations.Parameter;
+import io.swagger.v3.oas.annotations.tags.Tag;
+import jakarta.annotation.Resource;
+import java.util.Comparator;
+import java.util.List;
+import lombok.extern.slf4j.Slf4j;
+import org.eclipse.xpanse.api.config.ServiceTemplateEntityConverter;
+import org.eclipse.xpanse.modules.database.servicetemplate.ServiceTemplateEntity;
+import org.eclipse.xpanse.modules.models.common.enums.Category;
+import org.eclipse.xpanse.modules.models.common.enums.Csp;
+import org.eclipse.xpanse.modules.models.servicetemplate.enums.ServiceHostingType;
+import org.eclipse.xpanse.modules.models.servicetemplate.view.ServiceTemplateDetailVo;
+import org.eclipse.xpanse.modules.servicetemplate.ServiceTemplateManage;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.MediaType;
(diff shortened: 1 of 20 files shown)


modules/api/src/main/java/org/eclipse/xpanse/api/controllers/ServiceTemplateApi.java
```diff
@@ -210,19 +211,49 @@ public List<ServiceTemplateDetailVo> listServiceTemplates(
...
+    /**
+     * List service templates with query params.
+     *
+     * @param categoryName   name of category.
+     * @param cspName        name of cloud service provider.
+     * @param serviceName    name of service template.
+     * @param serviceVersion version of service template.
+     * @return response
+     */
+    @Tag(name = "Service Vendor",
```
swaroopar, line 233: we need a different tag here -> "Cloud Service Provider"
If possible, we can move this specific method to a new controller also. It would be more cleaner.

swaroopar (changes requested):

modules/api/src/main/java/org/eclipse/xpanse/api/controllers/ServiceTemplateApi.java
baixinsui: Updated as suggestion.

modules/api/src/main/java/org/eclipse/xpanse/api/config/DataModelConverter.java
```diff
@@ -0,0 +1,53 @@
...
+import org.eclipse.xpanse.api.controllers.ServiceCatalogApi;
+import org.eclipse.xpanse.modules.database.servicetemplate.ServiceTemplateEntity;
+import org.eclipse.xpanse.modules.models.servicetemplate.view.ServiceTemplateDetailVo;
+import org.springframework.beans.BeanUtils;
+import org.springframework.hateoas.server.mvc.WebMvcLinkBuilder;
+
+/**
+ * DataModelConverter.
+ */
+public class DataModelConverter {
```
swaroopar, line 18: we need a concrete name for this class otherwise this class might get too big with too many converters. 
We can call it -> `ServiceTemplateEntityConverter` -> then the class call have all converters related to `ServiceTemplateEntity`

modules/api/src/main/java/org/eclipse/xpanse/api/controllers/CspServiceTemplateApi.java
```diff
@@ -0,0 +1,89 @@
...
+     * List service templates with query params.
+     *
+     * @param categoryName   name of category.
+     * @param cspName        name of cloud service provider.
+     * @param serviceName    name of service template.
+     * @param serviceVersion version of service template.
+     * @return response
+     */
+    @Tag(name = "Cloud Service Provider",
+            description = "APIs for cloud service provider managing service templates.")
```
swaroopar, line 62: APIs for cloud service provider to manage service templates.

swaroopar (changes requested):

modules/api/src/main/java/org/eclipse/xpanse/api/config/DataModelConverter.java
baixinsui: Updated.

modules/api/src/main/java/org/eclipse/xpanse/api/controllers/CspServiceTemplateApi.java
baixinsui: Updated.

swaroopar (approved):
LGTM

PR #1398, opened 2024-02-21
introduce approval status to service registration

### LINKED ISSUE #1382: introduce approval status to service registration
- [ ] Introduce new status - APPROVAL_PENDING, APPROVED, REJECTED
- [ ] After the service is registered and updated - set the status to APPROVAL_PENDING
- [ ] in UserOrderableServices - return only APPROVED services
- [ ] In the deployRequest - validate that the request is only for approved service templates. 

We must ensure that the end user does not see the yet to be approved and rejected service templates anywhere.

baixinsui:
fixes #1382 
ISV user view details of registered service template:
image
End user view details of registered service template:
image
End user deploy service when found service template not approved:
image


swaroopar (approved):
LGTM

PR #1399, opened 2024-02-22
Extend OCL to provide auto-fill for deploy variables

### LINKED ISSUE #1337: Extend OCL to provide auto-fill for deploy variables
In order for the cloud provider UIs to provider option to choose existing configurations while deploying services.
`DeployVariable` must allow the service vendor to mention if the variable can simply use the existing data. 

- [ ] Extend `DeployVariable` to have new optional subtype - AutoFill
- [ ] AutoFill to contain fields - resourceType: DeployResourceKind Enum, isAllowCreate: boolean
- [ ] Extend DeployResourceKind to have another optional field which defines the parent resource. Eg., `SUBNET("subnet", DeployResourceKind.VPC);`
- [ ] Validation in service template register - if there is a deploy variable with AutoFill and if  the resourceType has a parent, then validate that there is another deploy variable also with autofill available for the parent resourceType. Otherwise throw exception.

WangLiNaruto:
Fixed #1337
image
image

### DIFF
--- a/modules/models/src/main/java/org/eclipse/xpanse/modules/models/servicetemplate/AutoFill.java
+++ b/modules/models/src/main/java/org/eclipse/xpanse/modules/models/servicetemplate/AutoFill.java
@@ -1,0 +1,27 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * SPDX-FileCopyrightText: Huawei Inc.
+ */
+
+package org.eclipse.xpanse.modules.models.servicetemplate;
+
+import io.swagger.v3.oas.annotations.media.Schema;
+import jakarta.validation.constraints.NotNull;
+import lombok.Data;
+import org.eclipse.xpanse.modules.models.service.deploy.enums.DeployResourceKind;
+
+/**
+ * Define autofill properties for deployment variables.
+ */
+@Data
+public class AutoFill {
+
+    @NotNull
+    @Schema(description = "Type of the cloud resource to be reused.")
+    private DeployResourceKind deployResourceKind;
+
+    @NotNull
+    @Schema(description = " defines if the required cloud resource can be newly created "
+            + "or should the existing resources must only be used.")
+    private Boolean isAllowCreate;
+}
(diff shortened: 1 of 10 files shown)


modules/models/src/main/java/org/eclipse/xpanse/modules/models/servicetemplate/AutoFill.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+@Data
+public class AutoFill {
+
+    @NotNull
+    @Schema(description = "Deployed resource type")
+    private DeployResourceKind deployResourceKind;
+
+    @NotNull
+    @Schema(description = "Not whether the variable is allowed to be created")
```
swaroopar, line 24: schema can be - defines if the required cloud resource can be newly created or should the existing resources must only be used.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/servicetemplate/AutoFill.java
```diff
@@ -0,0 +1,26 @@
...
+import org.eclipse.xpanse.modules.models.service.deploy.enums.DeployResourceKind;
+
+/**
+ * Define autofill properties for deployment variables.
+ */
+@Data
+public class AutoFill {
+
+    @NotNull
+    @Schema(description = "Deployed resource type")
```
swaroopar, line 20: Type of the cloud resource to be reused.

swaroopar (changes requested):
@WangLiNaruto can you also please add a new sample service file with auto fill data?
Sorry i missed to add this in the user story.

modules/servicetemplate/src/main/java/org/eclipse/xpanse/modules/servicetemplate/ServiceTemplateManage.java
```diff
@@ -289,6 +294,43 @@ public String getOpenApiUrl(String id) {
         return openApiUrl;
     }
 
+    private void validateDeployVariable(List<DeployVariable> deployVariables) {
+        if (deployVariables.isEmpty()) {
```
swaroopar, line 298: we do not need this validation. If something is not according to the data model specification, it will be automatically handled by Spring and rejected.

modules/servicetemplate/src/main/java/org/eclipse/xpanse/modules/servicetemplate/ServiceTemplateManage.java
```diff
@@ -289,6 +294,43 @@ public String getOpenApiUrl(String id) {
         return openApiUrl;
     }
 
+    private void validateDeployVariable(List<DeployVariable> deployVariables) {
```
swaroopar, line 297: Its better if we can move this to a separate bean -> DeployVariableAutoFillValidator
Otherwise, this class is becoming too big to handle

swaroopar (changes requested):

swaroopar (approved):
LGTM

PR #1420, opened 2024-02-27
Move terraform provider from code config to deployer script in service template

### LINKED ISSUE #1403: Remove the provider configuration from xpanse
Terraform provider configuration is currently stored in xpanse and this will force all service vendors to use the same version of the TF provider. This is an unnecessary overhead. 

- [x] Remove provider code and properties. 
- [x] Add provider code also to deployer script in sample files.

baixinsui:
fixes #1403 
Approved service template:
image
Service deploy and then destroy:
image


modules/servicetemplate/src/main/java/org/eclipse/xpanse/modules/servicetemplate/ServiceTemplateManage.java
```diff
@@ -209,17 +203,23 @@ public ServiceTemplateEntity registerServiceTemplateByUrl(String oclLocation) th
...
      *
-     * @param id the ID of
+     * @param id             the ID of
+     * @param operationType  the operation type. values:view,update,unregistered.
+     * @param checkNamespace check the namespace of the service template belonging to.
      * @return Returns service template DB newTemplate.
      */
-    public ServiceTemplateEntity getServiceTemplateDetails(String id, boolean checkNamespace) {
+    public ServiceTemplateEntity getServiceTemplateDetails(String id,
+                                                           String operationType,
```
swaroopar, line 212: i'm not sure if we really need this. 
if the purpose is to just get a better log message.. we can just havea generic message saying "No permisions to view or manage service templates ...................................."

baixinsui: Updated as suggested.

swaroopar (approved):
LGTM

PR #1430, opened 2024-03-01
implement oauth profile in xpanse

### LINKED ISSUE #1429: implement oauth profile in xpanse
currently all oauth related configuration is put in the zitadel profile. We must change this and make it similar to `terraform-boot`

- [ ] create new profile `oauth`
- [ ] Move all properties and classes that are generic to any oauth provider to this profile. 
- [ ] Keep only Zitadel specific config and code in zitadel profile. Eg., roles, metadata, etc

WangLiNaruto:
Fixed #1429


swaroopar:
@WangLiNaruto can you please also check all classes which are currently enabled with profile Zitadel only? for example `ZitadelWebSecurityFilter`. This class can be a generic class for any oauth2 provider. If yes, we can also rename the class.

modules/security/src/main/java/org/eclipse/xpanse/modules/security/zitadel/config/Oauth2ResourceWebSecurityFilter.java
```diff
@@ -46,11 +46,11 @@
  * in addition to these global rules.
  */
 @Slf4j
-@Profile("zitadel")
+@Profile({"zitadel", "oauth"})
```
swaroopar, line 49: do we need Zitadel profile here as well?

modules/security/src/main/java/org/eclipse/xpanse/modules/security/zitadel/config/openapi/Oauth2OpenApiConfig.java
```diff
@@ -26,7 +26,7 @@
 /**
  * Configuration springdoc security oauth2.
  */
-@Profile("zitadel")
+@Profile({"zitadel", "oauth"})
```
swaroopar, line 29: same here. we should be able to start this class without "oauth"
We need Zitadel only to set variables.

runtime/src/main/resources/application-zitadel.properties
```diff
@@ -3,28 +3,13 @@
...
-springdoc.show-login-endpoint=true
-springdoc.swagger-ui.oauth.use-pkce-with-authorization-code-grant=true
-springdoc.swagger-ui.oauth.clientId=${authorization.swagger.ui.client.id}
-oauth.authorization.url=${authorization.server.endpoint}/oauth/v2/authorize
-oauth.token.url=${authorization.server.endpoint}/oauth/v2/token
-oauth.protected.api.client.id=
-oauth.protected.api.client.secret=
\ No newline at end of file
+
+# set authorization server endpoint
```
swaroopar, line 7: this can be a bit misleading. 
We can either delete this file and keep only zitadel-testbed
or we can keep this file with empty values.

swaroopar (changes requested):

modules/security/src/main/java/org/eclipse/xpanse/modules/security/zitadel/config/Oauth2ResourceWebSecurityFilter.java
WangLiNaruto: image
These configurations rely on 配置在zitadel中的 “authorization.server.endpoint”，“authorization.api.client.id”， “authorization.api.client.secret”
image

runtime/src/main/resources/application-zitadel.properties
WangLiNaruto: We no longer have a zitadel-testbed configuration file. I have placed all zitadel configuration items in the application-zitadel.properties configuration file.

modules/security/src/main/java/org/eclipse/xpanse/modules/security/zitadel/config/Oauth2ResourceWebSecurityFilter.java
swaroopar: we must make oauth properties file completely independent of zitadel properties file. 
oauth file should have all properties and zitadel properties file should only help to override them.

modules/security/src/main/java/org/eclipse/xpanse/modules/security/zitadel/config/ZitadelOauth2Constant.java
```diff
@@ -11,87 +11,38 @@
...
-
-    /**
-     * Optional scope to request putting the email of the user into the claims.
-     */
-    public static final String PROFILE_SCOPE = "profile";
-
     /**
      * Optional scope to request putting the granted roles of the user into the claims.
      */
     public static final String GRANTED_ROLES_SCOPE = "urn:zitadel:iam:org:project:roles";
```
swaroopar, line 22: this can be removed as it is moved to properties file.

modules/security/src/main/java/org/eclipse/xpanse/modules/security/zitadel/config/ZitadelOauth2Constant.java
```diff
@@ -11,87 +11,38 @@
...
 
-    /**
-     * Optional scope to request putting the metadata of the user into the claims.
-     */
-    public static final String METADATA_SCOPE = "urn:zitadel:iam:user:metadata";
-
     /**
      * All scopes to request putting all required info of the user into the claims.
      */
     public static final String REQUIRED_SCOPES =
```
swaroopar, line 27: this one too can be moved.

modules/security/src/main/java/org/eclipse/xpanse/modules/security/zitadel/config/ZitadelOauth2Constant.java
```diff
@@ -11,87 +11,38 @@
...
 
-    /**
-     * Key in claims for getting the granted roles of the user.
-     */
-    public static final String GRANTED_ROLES_KEY = "urn:zitadel:iam:org:project:roles";
-
     /**
      * Key in claims for getting the metadata of the user.
      */
     public static final String METADATA_KEY = "urn:zitadel:iam:user:metadata";
```
swaroopar, line 44: this one also can be moved?

modules/security/src/main/java/org/eclipse/xpanse/modules/security/zitadel/config/ZitadelOauth2Constant.java
```diff
@@ -11,87 +11,38 @@
...
      * All scopes to request putting all required info of the user into the claims.
      */
     public static final String REQUIRED_SCOPES =
             "openid profile urn:zitadel:iam:org:projects:roles urn:zitadel:iam:user:metadata";
 
+
     /**
      * Key in claims for getting the id of the user.
      */
     public static final String USERID_KEY = "sub";
```
swaroopar, line 34: this one too?

runtime/src/main/resources/application-oauth.properties
```diff
@@ -0,0 +1,32 @@
+#
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Huawei Inc.
+#
+#
+
```
swaroopar, line 6: we can also add add empty values to all properties which we have newly added

swaroopar (changes requested):

modules/security/src/main/java/org/eclipse/xpanse/modules/security/zitadel/config/openapi/ZitadelRequiredRoleDescriptionCustomizer.java
```diff
@@ -19,7 +19,7 @@
  */
 @Profile("zitadel")
 @Component
-public class RequiredRoleDescriptionCustomizer implements OperationCustomizer {
+public class ZitadelRequiredRoleDescriptionCustomizer implements OperationCustomizer {
```
swaroopar, line 22: Even this class does not seem to be specific to Zitadel.

modules/security/src/main/java/org/eclipse/xpanse/modules/security/zitadel/config/openapi/ZitadelOpenApiOauth2Config.java
```diff
@@ -6,11 +6,6 @@
 
 package org.eclipse.xpanse.modules.security.zitadel.config.openapi;
```
swaroopar, line 7: this class does not seem to be specific to Zitadel.

swaroopar (approved):
LGTM

PR #1431, opened 2024-03-01
Filter Csp user managed service templates with csp value in metadata

baixinsui:
fixes https://github.com/eclipse-xpanse/xpanse/issues/1405
related to https://github.com/eclipse-xpanse/xpanse-iam/pull/19
Isv user registered service template with with csp `flexibleEngine`:
image
Csp user with the value `huawei` of the key `csp` in metadata list service templates only return service templates with csp `huawei`:
image
Csp user with the value `huawei` of the key `csp` in metadata get details of service templates with csp `flexibleEngine`:
image
Csp user with the value `huawei` of the key `csp` in metadata review service template with csp `flexibleEngine`:
image


swaroopar (approved):
LGTM

PR #1437, opened 2024-03-05
support running xpanse on one specific cloud provider

baixinsui:
fixes https://github.com/eclipse-xpanse/xpanse/issues/1433
related to https://github.com/eclipse-xpanse/xpanse-website/pull/175
related to https://github.com/eclipse-xpanse/xpanse-ui/pull/582

1. Compile jar using default command 'mvn clean install -DskipTests' without flag '-Pplugin-xxx'. All packages of plugins are packaged into executable jar packages:
image
Start this executable jar using default command 'java -jar runtime/target/xpanse-runtime-1.0.8-SNAPSHOT.jar', all supported plugins are enabled in the running time:
image

2. Compile jar using command 'mvn clean install -DskipTests -Pplugin-huawei' with flag '-Pplugin-xxx'. only the specific 'huawei' plugin package is packaged into executable jar packages:
image
Start this executable jar using command 'java -jar runtime/target/xpanse-runtime-1.0.8-SNAPSHOT.jar' . only the  specific plugin for Huawei Cloud is enabled in the running time:
image

3. When only one specific plugin is enabled in the running time, the following exception will be thrown when  other Csps are in the APIs request parameters. 
Parameter Csp in path or param:
image


runtime/src/main/resources/application-plugin-flexibleengine.properties
```diff
@@ -0,0 +1,6 @@
+#
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Huawei Inc.
+#
+#
+plugins.enabled=flexibleengine
```
swaroopar, line 6: @baixinsui it is better if we can avoid these new profiles. Instead get available CSPs directly from the plugin manager.

baixinsui: @swaroopar Without these configuration files, we can only decide which plugins are supported when packaging the executable Jar with flag '-Pplugin-xxx'.

baixinsui: According to the discussion results, I have deleted these configurations and related codes.

modules/api/src/main/java/org/eclipse/xpanse/api/controllers/AdminServicesApi.java
```diff
@@ -109,17 +110,17 @@ public SystemStatus healthCheck() {
...
     @Tag(name = "Admin", description = "APIs for administrating Xpanse")
-    @Operation(description = "List cloud service provider.")
-    @GetMapping(value = "/csp", produces = MediaType.APPLICATION_JSON_VALUE)
+    @Operation(description = "List cloud service providers.")
+    @GetMapping(value = "/csps", produces = MediaType.APPLICATION_JSON_VALUE)
     @ResponseStatus(HttpStatus.OK)
-    @Secured({ROLE_ADMIN, ROLE_ISV, ROLE_USER})
-    public List<Csp> getCsps(
+    @Secured({ROLE_ADMIN, ROLE_CSP, ROLE_ISV, ROLE_USER})
+    public List<Csp> listCsps(
```
swaroopar, line 117: I think it makes more sense to name it -> getActiveCsps
Probably also URI -> /csps/active

modules/api/src/main/java/org/eclipse/xpanse/api/controllers/ServiceDeployerApi.java
```diff
@@ -155,9 +156,10 @@ public List<DeployedService> listDeployedServicesDetails(
             @RequestParam(name = "serviceVersion", required = false) String serviceVersion,
             @Parameter(name = "serviceState", description = "deployment state of the service")
             @RequestParam(name = "serviceState", required = false)
-                    ServiceDeploymentState serviceState) {
+            ServiceDeploymentState serviceState) {
         // return type is DeployedService but actually returns one of the child types
         // VendorHostedDeployedServiceDetails or DeployedServiceDetails
+        cspPluginValidator.validatePluginEnabled(csp);
```
swaroopar, line 162: Discussing how to get this validation automatically done for each method using AOP.

modules/api/src/main/java/org/eclipse/xpanse/api/controllers/AdminServicesApi.java
baixinsui: Updated the API as suggested

modules/api/src/main/java/org/eclipse/xpanse/api/controllers/ServiceDeployerApi.java
baixinsui: Removed all of this validation for request body  into the AOP and added the validation method into the `CspEnumConverter` for `csp`  in the URI.

swaroopar (approved):
LGTM

PR #1446, opened 2024-03-07
catch all runtime exceptions during deploy and destroy

swaroopar:
fixes https://github.com/eclipse-xpanse/xpanse/issues/1344


iskey (approved):
LGTM

PR #1455, opened 2024-03-12
update OCL to support service specification modifications.

### LINKED ISSUE #1448: update OCL to support service specification modifications
- [ ] Create new type `ModificationImpact`
- [ ] Add `ModificationImpact` to flavors as mandatory variable
- [ ] Add `ModificationImpact` to each DeployVariable as mandatory variable.
- [ ] No logic in the deployer based on these values. These are only for clients. 

```yaml
modificationImpact:
    isDataLost: boolean
   isServiceInterrupted: boolean
```

Alice1319:
fixes #1448

### DIFF
--- a/modules/models/src/main/java/org/eclipse/xpanse/modules/models/servicetemplate/Flavor.java
+++ b/modules/models/src/main/java/org/eclipse/xpanse/modules/models/servicetemplate/Flavor.java
@@ -31,4 +31,8 @@ public class Flavor extends FlavorBasic implements Serializable {
     @Schema(description = "The properties of the flavor")
     private Map<String, String> properties;
 
+    @NotNull
+    @Schema(description = "Impact on service when flavor is changed.")
+    private ModificationImpact modificationImpact;
+
 }
--- a/modules/models/src/main/java/org/eclipse/xpanse/modules/models/servicetemplate/ModificationImpact.java
+++ b/modules/models/src/main/java/org/eclipse/xpanse/modules/models/servicetemplate/ModificationImpact.java
@@ -1,0 +1,28 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * SPDX-FileCopyrightText: Huawei Inc.
+ */
+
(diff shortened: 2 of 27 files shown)


modules/models/src/main/java/org/eclipse/xpanse/modules/models/servicetemplate/Flavor.java
```diff
@@ -31,4 +31,8 @@ public class Flavor extends FlavorBasic implements Serializable {
     @Schema(description = "The properties of the flavor")
     private Map<String, String> properties;
 
+    @NotNull
+    @Schema(description = "The modificationImpact of the flavor")
```
swaroopar, line 35: Impact on service when flavor is changed.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/servicetemplate/ModificationImpact.java
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * Defines for service ModificationImpact.
+ */
+@Data
+public class ModificationImpact implements Serializable {
+
+    @Serial
+    private static final long serialVersionUID = -4160044806152791293L;
+
+    @Schema(description = "The dataLost of ModificationImpact.")
```
swaroopar, line 22: Is data lost when service configuration is modified.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/servicetemplate/ModificationImpact.java
```diff
@@ -0,0 +1,27 @@
...
+    private Boolean isDataLost;
+
+    @Schema(description = "The serviceInterrupted of ModificationImpact.")
```
swaroopar, line 25: Is service availability interrupted when the configuration is interrupted.

swaroopar (changes requested):

swaroopar (approved):
LGTM

PR #1458, opened 2024-03-13
Extend DeployRequestBase class to have area information

### LINKED ISSUE #1435: Extend DeployRequestBase class to have area information
Currently we store only region as string. We must change this to have the `Region` (name and area name) instead. 

- [ ] Extend DeployRequestBase  to have `Region` object instead of just region name. 
- [ ] Update UI - new order and migration

WangLiNaruto:
Fixed #1435
related to eclipse-xpanse/xpanse-ui#611
image
image
image
image


swaroopar (approved):
LGTM

PR #1491, opened 2024-04-01
stabilize handling of service versions

baixinsui:
fixes https://github.com/eclipse-xpanse/xpanse/issues/1474
Register service template with invalid version:
chrome_N758fG0qNc
Register service template with same version:
chrome_c9UoJG30Ka
Register service template with lower version:
chrome_URRd9ZBvb3


modules/servicetemplate/src/main/java/org/eclipse/xpanse/modules/servicetemplate/ServiceTemplateManage.java
```diff
@@ -129,6 +132,37 @@ private ServiceTemplateEntity getNewServiceTemplateEntity(Ocl ocl) {
...
     }
 
+
+    private Semver getSemverVersion(String serviceVersion) {
+        try {
+            return new Semver(serviceVersion);
+        } catch (SemverException e) {
+            String errorMsg = String.format("The service version %s is a invalid semver version.",
+                    serviceVersion);
+            throw new IllegalArgumentException(errorMsg);
```
swaroopar, line 142: we need a new proper exception here - InvalidServiceVersionException

swaroopar (changes requested):

modules/servicetemplate/src/main/java/org/eclipse/xpanse/modules/servicetemplate/ServiceTemplateManage.java
baixinsui: Added new exception as suggestion.

swaroopar (approved):
LGTM

PR #1494, opened 2024-04-01
make all role protection in the API layer optional

### LINKED ISSUE #1493: make all role protection in the API layer optional
In order to prepare the application to be deployed on different setups, it is necessary to enable/disable role protection conditionally.
Currently all our API methods are annotated with `@Secured` annotation. This must be controllable based on a config.

- [ ] use solution https://stackoverflow.com/questions/22176236/disable-enableglobalmethodsecurity-annotation
- [ ] add a new configuration property - 'enable.role.protection' with default value as `true`

baixinsui:
fixes #1493
When the configuration property `enable.role.protection` is true:
Authorize with the user 'csp-huawei' only has role 'csp':
chrome_ccigFeKtfO
Request the API required role admin or isv or user, and the service returns 403.
image

When the configuration property `enable.role.protection` is false:
Authorize with the user 'csp-huawei' only has role 'csp':
image
Request the same API and the service returns 200.
image

### DIFF
--- a/modules/security/src/main/java/org/eclipse/xpanse/modules/security/Oauth2WebSecurityFilter.java
+++ b/modules/security/src/main/java/org/eclipse/xpanse/modules/security/Oauth2WebSecurityFilter.java
@@ -18,6 +18,7 @@ import org.eclipse.xpanse.modules.models.response.Response;
 import org.eclipse.xpanse.modules.models.response.ResultType;
 import org.eclipse.xpanse.modules.security.common.XpanseAuthentication;
 import org.springframework.beans.factory.annotation.Value;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Profile;
@@ -48,9 +49,9 @@ import org.springframework.web.servlet.handler.HandlerMappingIntrospector;
 @Slf4j
 @Profile("oauth")
 @Configuration
-@EnableWebSecurity
-@EnableMethodSecurity(securedEnabled = true)
 public class Oauth2WebSecurityFilter {
+
+    private final ObjectMapper objectMapper = new ObjectMapper();
 
     @Value("${spring.security.oauth2.resourceserver.opaquetoken.introspection-uri}")
     private String introspectionUri;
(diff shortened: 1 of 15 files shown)


modules/security/src/main/java/org/eclipse/xpanse/modules/security/Oauth2WebSecurityFilter.java
```diff
@@ -149,4 +138,58 @@ CorsConfigurationSource corsConfigurationSource() {
     }
 
 
+    /**
+     * Configuration applied when method security is disabled.
+     */
+    @Profile("oauth")
```
swaroopar, line 144: @baixinsui this is a topic that must work with "noauth" as well. 
Basically the idea is to allow xpanse to run completely without any authentication and authorization because in most of cases, the auth process happens outside the xpanse application runtime. So in this case, the role logic must also be removed. 
In case of running xpanse on HuaweiCloud, the API-GW will do the authentication (and authorization -> to be checked) for us. If API-GW does not support authorization (role check) then we will do it in the xpanse application in a custom way -> Read IAM to get roles assigned to user and then decide what methods to allow. This would mean that the default spring MVC roles extractor would not work.  

You can also discuss with WangLi once. He knows the background of this requirement.

swaroopar (changes requested):

modules/security/src/main/java/org/eclipse/xpanse/modules/security/Oauth2WebSecurityFilter.java
baixinsui: @swaroopar What you mean is when the property `enable.role.protection` is false, the  authentication should be disable also? In this case, maybe we need to rename the property to `enable.oauth.security` is more better.

swaroopar: if authentication is enabled -> then authorization can be enabled or disabled
if authentication is disabled -> then authorization is by default disabled

baixinsui: @swaroopar Disabling authentication and authorization is completed with the PR https://github.com/eclipse-xpanse/xpanse/pull/1499. This PR is not need to modify.

baixinsui: Updated the code based on the PR https://github.com/eclipse-xpanse/xpanse/pull/1499

swaroopar (approved):
LGTM

PR #1499, opened 2024-04-02
make authentication in API layer optional

### LINKED ISSUE #1498: make authentication in API layer optional
In order to run xpanse behind API-GWs, we must make authentication on all the APIs optional since the authentication will be automatically done by API-GWs. 

- [ ] Write a new filter which makes all endpoints accessible without authentication. 
- [ ] example - https://stackoverflow.com/questions/40228036/how-to-turn-off-spring-security-in-spring-boot-application/64662887#64662887
- [ ] add a new profile - `noauth` and enable this bean when the profile is activated.

baixinsui:
fixes #1498 
When  the application  is running with the profile `noauth`,  all the APIs can be called successfully without authentication.
image
chrome_e9ZJc0MuOw
chrome_AKpax2ZHQG


swaroopar (approved):
LGTM

PR #1519, opened 2024-04-10
refactor methods about check permissions and user info without auth

baixinsui:
fixes https://github.com/eclipse-xpanse/xpanse/issues/1495
When Xpanse is running without authentication and authorization, the services shown as below:
System health without identity service:
noauth_system_health_without_identity
Registered service templates with null namespace:
service-templates-with-null-namespace
All service templates with different csp can be reviewed:
service-templates-with-null-namespace
Manage user credential with `no-auth-user-id`:
manage-user-credentails-with-noauth-user-id
Deployed services with `no-auth-user-id`:
deploy-services-with-noauth-user-id
Migrate services with `no-auth-user-id` and deal failed task manually:
deploy-services-with-noauth-user-id


swaroopar (approved):
@baixinsui Looks great. Neat work.

PR #1523, opened 2024-04-11
Return empty list even if the SDK method throw exception

baixinsui:
fixes https://github.com/eclipse-xpanse/xpanse/issues/1520

### DIFF
--- a/plugins/flexibleengine/src/main/java/org/eclipse/xpanse/plugins/flexibleengine/manage/FlexibleEngineResourceManager.java
+++ b/plugins/flexibleengine/src/main/java/org/eclipse/xpanse/plugins/flexibleengine/manage/FlexibleEngineResourceManager.java
@@ -5,8 +5,9 @@
 
 package org.eclipse.xpanse.plugins.flexibleengine.manage;
 
+import static org.eclipse.xpanse.plugins.flexibleengine.common.FlexibleEngineRetryStrategy.DEFAULT_DELAY_MILLIS;
+
 import com.huaweicloud.sdk.core.auth.ICredential;
-import com.huaweicloud.sdk.core.exception.SdkException;
 import com.huaweicloud.sdk.ecs.v2.EcsClient;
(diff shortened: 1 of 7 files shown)


plugins/flexibleengine/src/main/java/org/eclipse/xpanse/plugins/flexibleengine/manage/FlexibleEngineResourceManager.java
```diff
@@ -98,130 +99,159 @@ public List<String> getExistingResourceNamesWithKind(String userId,
...
             if (response.getHttpStatusCode() == 200) {
-                availabilityZones = response.getAvailabilityZoneInfo().stream()
+                availabilityZoneNames = response.getAvailabilityZoneInfo().stream()
                         .map(NovaAvailabilityZone::getZoneName).toList();
             }
-        } catch (SdkException e) {
-            log.error("Get FlexibleEngine availability zones of region:{} failed, error:{}",
-                    region, e.getMessage());
+        } catch (Exception e) {
+            log.error("FlexibleEngineClient listAvailabilityZones with region:{} error.", region,
```
swaroopar, line 116: @baixinsui i'm a little confused. here the problem is still the same know? we still just catch all exceptions and return an empty list. My expectation is that we throw an exception to client if this call fails.

swaroopar (changes requested):

plugins/flexibleengine/src/main/java/org/eclipse/xpanse/plugins/flexibleengine/manage/FlexibleEngineResourceManager.java
baixinsui: Ok.I misunderstood. I will throw the ClientApiCallFailedException(errorMsg)

swaroopar (approved):
LGTM

PR #1524, opened 2024-04-11
implement AOP based solution for tracking all API requests

### LINKED ISSUE #1461: implement AOP based solution for tracking all API requests
We need a way to track/audit all API requests coming into xpanse. The reason behind such a requirement is, most cloud providers write all service modification requests to specific systems. In case of Huawei cloud, the data is written to CTS service.

- [ ] Find a way to figure out how to extract for which CSP is the request intended to. Some available options -> make CSP a mandatory URI field in all POST, PUT and DELETE request, or add CSP as header, etc. 
- [ ] Use Spring AOP to `pointCut` all controller POST, PUT and DELETE methods.
- [ ] OrchestratorPlugin interface to have a new method -> auditApiRequest. This method must be to write a simple log line as  default implementation. 
- [ ] Audit only POST, PUT and DELETE requests. 

Please also discuss with @baixinsui about the

WangLiNaruto:
Fixed #1461

image
image
image

image
image

### DIFF
--- a/modules/api/src/main/java/org/eclipse/xpanse/api/config/AuditLogWriter.java
+++ b/modules/api/src/main/java/org/eclipse/xpanse/api/config/AuditLogWriter.java
@@ -1,0 +1,156 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * SPDX-FileCopyrightText: Huawei Inc.
+ */
+
+package org.eclipse.xpanse.api.config;
+
+import jakarta.annotation.Resource;
+import jakarta.servlet.http.HttpServletRequest;
+import java.lang.reflect.Method;
+import java.time.OffsetDateTime;
+import java.util.Objects;
+import lombok.extern.slf4j.Slf4j;
+import org.aspectj.lang.ProceedingJoinPoint;
+import org.aspectj.lang.annotation.Around;
+import org.aspectj.lang.annotation.Aspect;
+import org.aspectj.lang.annotation.Pointcut;
+import org.aspectj.lang.reflect.MethodSignature;
+import org.eclipse.xpanse.modules.models.common.enums.Csp;
+import org.eclipse.xpanse.modules.models.credential.CreateCredential;
+import org.eclipse.xpanse.modules.models.policy.userpolicy.UserPolicyCreateRequest;
+import org.eclipse.xpanse.modules.models.policy.userpolicy.UserPolicyUpdateRequest;
+import org.eclipse.xpanse.modules.models.service.deploy.DeployRequest;
+import org.eclipse.xpanse.modules.models.servicetemplate.Ocl;
+import org.eclipse.xpanse.modules.models.workflow.migrate.MigrateRequest;
+import org.eclipse.xpanse.modules.orchestrator.OrchestratorPlugin;
+import org.eclipse.xpanse.modules.orchestrator.PluginManager;
+import org.eclipse.xpanse.modules.orchestrator.audit.AuditLog;
+import org.eclipse.xpanse.modules.security.UserServiceHelper;
+import org.springframework.stereotype.Component;
+import org.springframework.web.context.request.RequestContextHolder;
+import org.springframework.web.context.request.ServletRequestAttributes;
+
+/**
+ * Writer for audit logs.
+ */
+@Slf4j
+@Aspect
+@Component
+public class AuditLogWriter {
+
+    private static final String DEFAULT_GET_CSP_METHOD_NAME = "getCspFromRequestUri";
+
+    @Resource
+    private PluginManager pluginManager;
+    @Resource
+    private UserServiceHelper userServiceHelper;
+    @Resource
+    private GetCspInfoFromRequest getCspInfoFromRequest;
+
+    /**
+     * Pointcut for all controller methods.
+     */
+    @Pointcut("execution(* org.eclipse.xpanse.api.controllers.*.*(..))")
+    public void controllerMethods() {
+    }
+
+    /**
+     * Validate request parameters.
+     *
+     * @param joinPoint The join point.
+     */
+    @Around("controllerMethods()")
+    public Object auditRequest(ProceedingJoinPoint joinPoint) throws Throwable {
+        Csp csp = getCsp(joinPoint);
+        log.info("Get csp from request attributes: {}", csp);
+        Object result = joinPoint.proceed();
+        if (Objects.nonNull(csp)) {
+            HttpServletRequest request =
+                    ((ServletRequestAttributes) RequestContextHolder.currentRequestAttributes())
+                            .getRequest();
+
+            OrchestratorPlugin orchestratorPlugin = pluginManager.getOrchestratorPlugin(csp);
+            orchestratorPlugin.auditApiRequest(
+                    getAuditLog(csp, joinPoint.getSignature().getName(), request.getMethod(),
+                            joinPoint.getArgs(), result));
+        }
+        return result;
+    }
+
+    private Csp getCsp(ProceedingJoinPoint joinPoint) {
+        if (pluginManager.getPluginsMap().size() == 1) {
+            return pluginManager.getPluginsMap().keySet().iterator().next();
+        }
+        Method requestMethod = ((MethodSignature) joinPoint.getSignature()).getMethod();
+        AuditApiRequest auditApiRequest = requestMethod.getAnnotation(AuditApiRequest.class);
+        if (Objects.isNull(auditApiRequest) || !auditApiRequest.enabled()) {
+            return null;
+        }
+        return getCspFromMethodInfo(auditApiRequest, joinPoint.getArgs());
+    }
+
+    private Csp getCspFromMethodInfo(AuditApiRequest auditApiRequest, Object[] args) {
+        String methodName = auditApiRequest.methodName();
+        if (DEFAULT_GET_CSP_METHOD_NAME.equals(methodName)) {
+            return getCspFromRequestParams(args);
+        }
+        try {
+            Class<?> clazz = auditApiRequest.clazz();
+            Class<?>[] paramTypes = auditApiRequest.paramTypes();
+            Method method = clazz.getMethod(methodName, paramTypes);
+            int[] paramIndexes = auditApiRequest.paramIndexes();
+            Object[] paramValues = new Object[paramIndexes.length];
+            for (int i = 0; i < paramIndexes.length; i++) {
+                paramValues[i] = args[paramIndexes[i]];
+            }
+            Object result = method.invoke(getCspInfoFromRequest, paramValues);
+            if (result instanceof Csp csp) {
+                return csp;
+            }
+        } catch (Exception e) {
+            log.error("Get csp with annotation @AuditApiRequest error.", e);
+        }
+        return null;
+    }
+
+    private Csp getCspFromRequestParams(Object[] args) {
+        for (Object arg : args) {
+            if (arg instanceof Ocl ocl) {
+                return ocl.getCloudServiceProvider().getName();
+            } else if (arg instanceof DeployRequest deployRequest) {
+                return deployRequest.getCsp();
+            } else if (arg instanceof MigrateRequest migrateRequest) {
+                return migrateRequest.getCsp();
+            } else if (arg instanceof CreateCredential createCredential) {
+                return createCredential.getCsp();
+            } else if (arg instanceof UserPolicyCreateRequest userPolicyCreateRequest) {
+                return userPolicyCreateRequest.getCsp();
+            } else if (arg instanceof UserPolicyUpdateRequest userPolicyUpdateRequest) {
+                return userPolicyUpdateRequest.getCsp();
+            } else if (arg instanceof Csp csp) {
+                return csp;
+            } else {
+                return null;
+            }
+        }
+        return null;
+    }
+
+    private AuditLog getAuditLog(Csp csp, String methodName, String methodType, Object[] args,
+            Object result) {
+        AuditLog auditLog = new AuditLog();
+        auditLog.setMethodName(methodName);
+        auditLog.setMethodType(methodType);
+        auditLog.setParams(args);
+        auditLog.setResult(result);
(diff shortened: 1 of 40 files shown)


baixinsui:
@WangLiNaruto  We need to add a test to check if all API methods have annotation @GetCspMethodInfo added. If any API method has not been annotated, the test case will fail and an error message which API method in which class has not been annotated will be thrown.

modules/common/src/main/java/org/eclipse/xpanse/common/aop/HttpRequestInfoManager.java
```diff
@@ -0,0 +1,55 @@
...
+import org.eclipse.xpanse.modules.models.common.enums.Csp;
+import org.springframework.web.context.request.RequestAttributes;
+import org.springframework.web.context.request.RequestContextHolder;
+import org.springframework.web.context.request.ServletRequestAttributes;
+
+/**
+ * Bean to provide methods to manage request attributes.
+ */
+@Slf4j
+public class HttpRequestInfoManager {
```
swaroopar, line 21: Is this class really required? We do not seem to use the CSP from http request attribute.

modules/api/src/main/java/org/eclipse/xpanse/api/config/GetCspMethodInfo.java
```diff
@@ -0,0 +1,53 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Get custom annotations for csp information.
+ */
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface GetCspMethodInfo {
```
swaroopar, line 19: This annotation can be renamed as `AuditApiRequest`

modules/api/src/main/java/org/eclipse/xpanse/api/config/AuditLogWriter.java
```diff
@@ -0,0 +1,164 @@
...
+    public void controllerMethods() {
+    }
+
+    /**
+     * Validate request parameters.
+     *
+     * @param joinPoint The join point.
+     */
+    @Around("controllerMethods()")
+    public Object setCspIntoHttpRequestInfo(ProceedingJoinPoint joinPoint) throws Throwable {
```
swaroopar, line 63: this method should be called "auditRequest" or something like that.

modules/api/src/main/java/org/eclipse/xpanse/api/config/AuditLogWriter.java
```diff
@@ -0,0 +1,164 @@
...
+        String methodName = joinPoint.getSignature().getName();
+        log.info("method: {} start execute.", methodName);
+        Csp csp = getCsp(joinPoint);
+        log.info("Get csp from request attributes: {}", csp);
+        Object result = joinPoint.proceed();
+        if (Objects.isNull(csp)) {
+            return result;
+        }
+        ServletRequestAttributes attributes =
```
swaroopar, line 72: we do not need this extra header.

modules/api/src/main/java/org/eclipse/xpanse/api/config/AuditLogWriter.java
```diff
@@ -0,0 +1,164 @@
...
+                csp = getCspFromRequestParams(args);
+            } else {
+                csp = getCspWithMethodInfo(getCspMethodInfo, args);
+            }
+            HttpRequestInfoManager.setCspIntoRequestAttributes(csp);
+        }
+        return csp;
+    }
+
+    private Csp getCspWithMethodInfo(GetCspMethodInfo getCspMethodInfo, Object[] args) {
```
swaroopar, line 108: method name getCspFromMethodInfo

modules/models/src/main/java/org/eclipse/xpanse/modules/models/policy/userpolicy/UserPolicyUpdateRequest.java
```diff
@@ -21,7 +20,6 @@ public class UserPolicyUpdateRequest {
     /**
      * The id of the policy.
      */
-    @NotNull
```
swaroopar, line 24: why is this removed?

modules/api/src/main/java/org/eclipse/xpanse/api/config/GetCspMethodInfo.java
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * The index of params can be used to get csp info with the method.
+     */
+    int[] paramIndexes() default {0};
+
+    /**
+     * Identifies whether csp can be obtained directly.
+     */
+    boolean enabled() default true;
```
swaroopar, line 45: two booleans fields. It is confusing. 
My understanding is that the same annotation can be used for all types of methods. So the only boolean flag is necessary to disable audit completely.

modules/api/src/main/java/org/eclipse/xpanse/api/controllers/IsvServiceDeployApi.java
```diff
@@ -81,6 +83,7 @@ public List<DeployedService> listDeployedServicesOfIsv(
     @GetMapping(value = "/services/isv/details/vendor_hosted/{id}",
             produces = MediaType.APPLICATION_JSON_VALUE)
     @ResponseStatus(HttpStatus.OK)
+    @GetCspMethodInfo(isAudit = false)
```
swaroopar, line 86: This must be audited.

modules/api/src/main/java/org/eclipse/xpanse/api/controllers/OpenTofuMakerWebhookApi.java
```diff
@@ -49,6 +50,7 @@ public class OpenTofuMakerWebhookApi {
     @PostMapping(value = "${webhook.tofu-maker.deployCallbackUri}/{task_id}", produces =
             MediaType.APPLICATION_JSON_VALUE)
     @ResponseStatus(HttpStatus.OK)
+    @GetCspMethodInfo(enabled = false, methodName = "getCspWithServiceId")
```
swaroopar, line 53: if it is not enabled, why do we need the methodName? Comment applies to few other methods as well.  Please check all.

modules/api/src/main/java/org/eclipse/xpanse/api/controllers/ServiceCatalogApi.java
```diff
@@ -109,6 +111,7 @@ public List<UserOrderableServiceVo> listOrderableServices(
     @GetMapping(value = "/catalog/services/{id}",
             produces = MediaType.APPLICATION_JSON_VALUE)
     @ResponseStatus(HttpStatus.OK)
+    @GetCspMethodInfo(isAudit = false)
```
swaroopar, line 114: we can audit this.

swaroopar (changes requested):

modules/models/src/main/java/org/eclipse/xpanse/modules/models/policy/userpolicy/UserPolicyUpdateRequest.java
WangLiNaruto: image
The API interface has changed and the id has been moved to the path, so this field can be null.

modules/api/src/main/java/org/eclipse/xpanse/api/controllers/OpenTofuMakerWebhookApi.java
```diff
@@ -50,6 +51,7 @@ public class OpenTofuMakerWebhookApi {
     @PostMapping(value = "${webhook.tofu-maker.deployCallbackUri}/{task_id}", produces =
             MediaType.APPLICATION_JSON_VALUE)
     @ResponseStatus(HttpStatus.OK)
+    @AuditApiRequest(enabled = true, methodName = "getCspWithServiceId")
```
swaroopar, line 54: we can disable the audit for all callback methods.

swaroopar (changes requested):

modules/api/src/test/java/org/eclipse/xpanse/api/config/XpanseAuditApiRequestTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+    public static final String ANNOTATED_METHOD_PACKAGE_NAME = "org.eclipse.xpanse.api.controllers";
+
+    @Test
+    void testAllAnnotatedMethodsHaveAuditApiRequest() {
+        List<Method> annotatedMethods = getAnnotatedMethods(ANNOTATED_METHOD_PACKAGE_NAME);
+
+        for (Method method : annotatedMethods) {
+            assertTrue(method.isAnnotationPresent(AuditApiRequest.class),
+                    "Method " + method.getName() + "  does not have @AuditApiRequest annotation.");
```
baixinsui, line 25: It is better to add class name in the error message.

modules/api/src/test/java/org/eclipse/xpanse/api/config/XpanseAuditApiRequestTest.java
```diff
@@ -0,0 +1,64 @@
...
+        List<Method> annotatedMethods = new ArrayList<>();
+        List<Class<?>> classes = getClasses(packageName);
+        for (Class<?> clazz : classes) {
+            Method[] methods = clazz.getDeclaredMethods();
+            for (Method method : methods) {
+                if (method.isAnnotationPresent(PostMapping.class) ||
+                        method.isAnnotationPresent(GetMapping.class) ||
+                        method.isAnnotationPresent(PutMapping.class) ||
+                        method.isAnnotationPresent(DeleteMapping.class)) {
+                    annotatedMethods.add(method);
```
baixinsui, line 39: @RequestMapping and @PatchMapping are missing.

modules/api/src/main/java/org/eclipse/xpanse/api/config/GetCspInfoFromRequest.java
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+
+    /**
+     * Get Csp with id of service template.
+     *
+     * @param id id of service template.
+     * @return csp.
+     */
+    public Csp getCspWithTemplateId(String id) {
```
swaroopar, line 78: getCspWithTemplateId -> getCspFromServiceTemplateId

modules/api/src/main/java/org/eclipse/xpanse/api/config/GetCspInfoFromRequest.java
```diff
@@ -0,0 +1,190 @@
...
+        return null;
+    }
+
+    /**
+     * Get Csp with id of service.
+     *
+     * @param id id of service.
+     * @return csp.
+     */
+    public Csp getCspWithServiceId(String id) {
```
swaroopar, line 95: getCspFromServiceId

modules/api/src/main/java/org/eclipse/xpanse/api/config/GetCspInfoFromRequest.java
```diff
@@ -0,0 +1,190 @@
...
+        return null;
+    }
+
+    /**
+     * Get Csp with id of migrate.
+     *
+     * @param id id of service.
+     * @return csp.
+     */
+    public Csp getCspWithMigrateId(String id) {
```
swaroopar, line 112: getCspFromServiceMigrationId

modules/api/src/main/java/org/eclipse/xpanse/api/config/GetCspInfoFromRequest.java
```diff
@@ -0,0 +1,190 @@
...
+        return null;
+    }
+
+    /**
+     * Get Csp with id of user policy.
+     *
+     * @param id id of user policy.
+     * @return csp.
+     */
+    public Csp getCspWithUserPolicyId(String id) {
```
swaroopar, line 134: getCspFromUserPolicyId

modules/api/src/main/java/org/eclipse/xpanse/api/config/GetCspInfoFromRequest.java
```diff
@@ -0,0 +1,190 @@
...
+        return null;
+    }
+
+    /**
+     * Get Csp with taskId of migrate task.
+     *
+     * @param taskId id of migrate task.
+     * @return csp.
+     */
+    public Csp getCspWithWorkflowTaskId(String taskId) {
```
swaroopar, line 168: getCsFromWorkflowTaskId

modules/api/src/main/java/org/eclipse/xpanse/api/config/GetCspInfoFromRequest.java
```diff
@@ -0,0 +1,190 @@
...
+    @Resource
+    private DatabaseServiceMigrationStorage serviceMigrationStorage;
+
+    /**
+     * Get Csp with the URL of Ocl.
+     *
+     * @param url url string of Ocl.
+     * @return csp.
+     */
+    public Csp getCspWithOclLocation(String url) {
```
swaroopar, line 60: getCspFromOclLocation

modules/api/src/main/java/org/eclipse/xpanse/api/config/AuditLogWriter.java
```diff
@@ -0,0 +1,160 @@
...
+
+/**
+ * Writer for audit logs.
+ */
+@Slf4j
+@Aspect
+@Component
+public class AuditLogWriter {
+
+    private static final String DEFAULT_GET_CSP_METHOD_NAME = "getCspFromRequestUri";
```
swaroopar, line 42: what i actually meant is.. we will put this name also in the all the controller methods instead of leaving it blank there. Then it is more clear. We can remove this static parameter from here and simply add this method name to all controller methods where it is empty now.

swaroopar (changes requested):

swaroopar (approved):
LGTM

PR #1550, opened 2024-04-17
add service lock method for deployed services

baixinsui:
fixes https://github.com/eclipse-xpanse/xpanse/issues/1533


runtime/src/test/java/org/eclipse/xpanse/runtime/modules/common/OpenApiGeneratorJarManageTest.java
```diff
@@ -43,6 +44,7 @@ void testGetOpenApiWorkdir() {
     }
 
     @Test
+    @Disabled
```
swaroopar, line 47: @baixinsui  we must enable this test again.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/service/config/ServiceLockConfig.java
```diff
@@ -0,0 +1,33 @@
...
+ * Service lock config model.
+ */
+@Data
+public class ServiceLockConfig implements Serializable {
+
+    @Serial
+    private static final long serialVersionUID = 5655216229043227729L;
+
+    /**
+     * Whether the service can not be modified.
```
swaroopar, line 23: comment and description -> Whether the service is locked from deletion

modules/models/src/main/java/org/eclipse/xpanse/modules/models/service/config/ServiceLockConfig.java
```diff
@@ -0,0 +1,33 @@
...
+     */
+    @Schema(description = "Whether the service can not be destroyed.")
+    private boolean isDestroyLocked;
+    /**
+     * Whether the service can not be modified.
+     */
```
swaroopar, line 29: comment and description -> Whether the service is locked from modifications

swaroopar (changes requested):

runtime/src/test/java/org/eclipse/xpanse/runtime/modules/common/OpenApiGeneratorJarManageTest.java
baixinsui: Enabled

modules/models/src/main/java/org/eclipse/xpanse/modules/models/service/config/ServiceLockConfig.java
baixinsui: Updated.

baixinsui: Updated.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/service/config/ServiceLockConfig.java
baixinsui: updated.

swaroopar (approved):
LGTM

PR #1572, opened 2024-04-19
Store modify info into DB

### LINKED ISSUE #1563: flavor not updated after service scaling is completed
After the service scaling request is completed, the flavor value in DB is not updated.

baixinsui:
fixes #1563 
Service data before modification:
image
Modify service:
image
Service data after modification failed:
image
Service data after modification success:
image


swaroopar (approved):
LGTM

PR #1573, opened 2024-04-19
clean previous error when modification successful

### LINKED ISSUE #1564: modification previous error not cleared.
When modification deployment fails, we store the error as expected. 
But when we try the modification again and it is successful, then we are not clearing the previous error. The previous error must be deleted from the DB.

image

baixinsui:
fixes #1564 
First modification failure:
chrome_prDbpPMC4G
Then modification success:
chrome_ugInZ0DmQU


swaroopar (approved):
LGTM

Issue #1541, opened 2024-04-15
add eula status to deploymentRequest and service templates


swaroopar:
- [ ] eula information to be returned in ISV, CSP and User service template data models. 
- [ ] add a new optional `isEulaAccepted` field to deployRequest. 
- [ ] If the service template has `eula` defined, then this field must be true. 
- [ ] if the service template has `eula` defined but the field is false, then throw new exception -> EulaNotAccepted

Depends on #1540

WangLiNaruto:
let me do it

PR #1582, opened 2024-04-23
add eula status to deploymentRequest and service templates

### LINKED ISSUE #1540: add new field eula to OCL
- [ ] add new field string `eula` to OCL
- [ ] field to be optional
- [ ] description of the field ->  end user license agreement content of the service.

WangLiNaruto:
Fixed #1540  and #1541
related to eclipse-xpanse/xpanse-ui#730

### DIFF
--- a/modules/models/src/main/java/org/eclipse/xpanse/modules/models/response/ResultType.java
+++ b/modules/models/src/main/java/org/eclipse/xpanse/modules/models/response/ResultType.java
@@ -62,6 +62,7 @@ public enum ResultType {
     SERVICE_MIGRATION_FAILED_EXCEPTION("Service Migration Failed Exception"),
     SERVICE_MIGRATION_NOT_FOUND("Service Migration Not Found"),
     SERVICE_LOCKED("Service Locked"),
+    EULA_NOT_ACCEPTED("Eula not accepted"),
     INVALID_GIT_REPO_DETAILS("Invalid Git Repo Details");
 
     private final String value;
--- a/modules/models/src/main/java/org/eclipse/xpanse/modules/models/service/deploy/DeployRequestBase.java
+++ b/modules/models/src/main/java/org/eclipse/xpanse/modules/models/service/deploy/DeployRequestBase.java
@@ -95,4 +95,7 @@ public class DeployRequestBase implements Serializable {
 
     @Schema(description = "The availability zones to deploy the service instance.")
     private Map<String, String> availabilityZones;
+
+    @Schema(description = "The flag to determine if the user has accepted the EULA terms.")
+    private boolean isEulaAccepted;
 }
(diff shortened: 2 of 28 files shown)


samples/flexibleEngine-Kafka.yml
```diff
@@ -61,6 +61,7 @@ flavors:
 # The contact details of the service.
 serviceProviderContactDetails:
   email: ["test@test.com"]
+eula:
```
swaroopar, line 64: it need not be empty string always. It must work without the "eula" field completely in the OCL file. So we can remove files with these empty entries. Instead have couple of samples with a eula field and also text added to it.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/response/ResultType.java
```diff
@@ -62,6 +62,7 @@ public enum ResultType {
     SERVICE_MIGRATION_FAILED_EXCEPTION("Service Migration Failed Exception"),
     SERVICE_MIGRATION_NOT_FOUND("Service Migration Not Found"),
     SERVICE_LOCKED("Service Locked"),
+    EULA_NOT_ACCEPT("Eula not accepted"),
```
swaroopar, line 65: should be EULA_NOT_ACCEPTED

modules/models/src/main/java/org/eclipse/xpanse/modules/models/service/deploy/DeployRequestBase.java
```diff
@@ -95,4 +95,7 @@ public class DeployRequestBase implements Serializable {
 
     @Schema(description = "The availability zones to deploy the service instance.")
     private Map<String, String> availabilityZones;
+
+    @Schema(description = "The mark of whether to accept Eula’s services.")
```
swaroopar, line 99: description can be - The flag to determine if the user has accepted the EULA terms.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/service/deploy/exceptions/EulaNotAccepted.java
```diff
@@ -0,0 +1,17 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * SPDX-FileCopyrightText: Huawei Inc.
+ */
+
+package org.eclipse.xpanse.modules.models.service.deploy.exceptions;
+
+/**
+ * Exception thrown when deployment service does not accept Eula.
```
swaroopar, line 9: Exception thrown when the service template defines an EULA but the end user has requested for the service without accepting it.

swaroopar (changes requested):

swaroopar (approved):
LGTM

PR #1588, opened 2024-04-25
Add pricing information to each flavor

baixinsui:
fixes https://github.com/eclipse-xpanse/xpanse/issues/1544

ISV users query service templates with field `pricing`:
image
CSP users review service templates with filed `pricing` :
image
End-users query orderable service templates without field `pricing` :
image


modules/models/src/main/java/org/eclipse/xpanse/modules/models/servicetemplate/ServiceFlavor.java
```diff
@@ -30,8 +31,8 @@ public class ServiceFlavor implements Serializable {
     private String name;
 
     @NotNull
-    @Schema(description = "The price of the flavor")
-    private Integer fixedPrice;
+    @Schema(description = "The pricing of the flavor")
+    private RatingMode pricing;
```
swaroopar, line 35: I feel now that it is better we make a different type for ISV with the `pricing` and the existing ServiceFlavor without `pricing` for end users. What do you think about it?

modules/models/src/main/java/org/eclipse/xpanse/modules/models/billing/ResourceUsage.java
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Defines the data model for the resource usage.
+ */
+@Data
+public class ResourceUsage implements Serializable {
+
+    @Serial
+    private static final long serialVersionUID = 240913796673011260L;
+
+    @Schema(description = "The license price of the flavor of the manged service.")
```
swaroopar, line 24: typo in the world `managed`. To be fixed in all fields

modules/models/src/main/java/org/eclipse/xpanse/modules/models/billing/ResourceUsage.java
```diff
@@ -0,0 +1,33 @@
...
+    private Price licensePrice;
+
+    @Schema(description = "The listed price of the flavor of the manged service.")
```
swaroopar, line 27: The additional price to be added on top of the calculated price.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/billing/RatingMode.java
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Defines the rating mode of the managed service.
+ */
+@Data
+public class RatingMode implements Serializable {
+
+    @Serial
+    private static final long serialVersionUID = 240913796673011260L;
+
+    @Schema(description = "The fixed price of the flavor in the managed service.")
```
swaroopar, line 23: Extend the comment as - The fixed price of the flavor in the managed service. This price includes all prices and this is the price shown to the customer.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/billing/RatingMode.java
```diff
@@ -0,0 +1,32 @@
...
+    private Price fixedPrice;
+
+    @Schema(description = "The resource usage of the flavor in the managed service.")
+    private ResourceUsage resourceUsage;
+
+    @NotNull
+    @Schema(description = "Whether the price is only for management layer.")
```
swaroopar, line 30: Whether the price is only for management layer. Consumption of the workload resources will be billed additionally.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/servicetemplate/utils/OclLoader.java
```diff
@@ -8,19 +8,92 @@
...
+        List<ServiceFlavor> serviceFlavors = ocl.getFlavors().getServiceFlavors();
+        Map<String, Long> nameCountMap = serviceFlavors.stream()
+                .collect(Collectors.groupingBy(ServiceFlavor::getName, Collectors.counting()));
+        nameCountMap.entrySet().stream().filter(entry -> entry.getValue() > 1).forEach(entry -> {
+            String message =
+                    String.format("Service flavor with name %s is duplicated.", entry.getKey());
+            log.error(message);
+            errors.add(message);
+        });
+        List<BillingMode> billingModes = ocl.getBilling().getBillingModes();
```
swaroopar, line 59: I think this is not a good place for this logic. 
We can add a new component for `BillingConfigValidator` in `servicetemplate` module and use it.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/servicetemplate/utils/OclLoader.java
```diff
@@ -8,19 +8,92 @@
...
+            RatingMode ratingMode = serviceFlavor.getPricing();
+            String name = serviceFlavor.getName();
+            if (Objects.isNull(ratingMode.getFixedPrice()) && Objects.isNull(
+                    ratingMode.getResourceUsage())) {
+                String message = String.format("Service flavor %s has no 'fixedPrice' or "
+                        + "'resourceUsage' defined in 'pricing'.", name);
+                log.error(message);
+                errors.add(message);
+            } else {
+                if (isSupportedPayPerUse) {
```
swaroopar, line 74: do we really need these two checks just for payPerUse?
The first check in line 67 should be enough for all types of billing modes?

swaroopar (changes requested):

modules/models/src/main/java/org/eclipse/xpanse/modules/models/servicetemplate/ServiceFlavor.java
baixinsui: If there are some changes to `ServiceFlavor` in featrue, it may not be good for us to maintain two models. If you don't want to return null, we can set the objectMapper configuration.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/servicetemplate/utils/OclLoader.java
baixinsui: Updated with new billingModes

modules/models/src/main/java/org/eclipse/xpanse/modules/models/servicetemplate/ServiceFlavor.java
swaroopar: @baixinsui My only concern is, if the end user sees the data model, he sees that there is a "price" field there. so returning null or removing the field completely does not matter.

swaroopar (approved):
LGTM

PR #1593, opened 2024-04-26
add eula status to deploymentRequest and service templates

### LINKED ISSUE #1541: add eula status to deploymentRequest and service templates

WangLiNaruto:
Fixed #1541


swaroopar (approved):
LGTM

PR #1601, opened 2024-05-06
Extend flavor with priority to control service downgrade

baixinsui:
fixes https://github.com/eclipse-xpanse/xpanse/issues/1577
Modify service flavor from lower priority to higher priority successfully:
msedge_KMVyD9sqXa
Modify service flavor from higher priority to lower priority failed:
msedge_CofYZ8ApMV


modules/models/src/main/java/org/eclipse/xpanse/modules/models/servicetemplate/ServiceFlavor.java
```diff
@@ -34,4 +34,7 @@ public class ServiceFlavor implements Serializable {
     @Schema(description = "The properties of the flavor")
     private Map<String, String> properties;
 
+    @NotNull
+    @Schema(description = "The priority of the flavor. The larger value means higher priority.")
```
swaroopar, line 38: it must be other way round. "The larger the value means lower priority"

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java
```diff
@@ -312,6 +325,20 @@ public DeployTask getModifyTask(ModifyRequest modifyRequest,
         return modifyTask;
     }
 
+    private void validateFlavorDowngradedIsAllowed(String originalFlavor, String newFlavor,
+                                                   FlavorsWithPrice flavors) {
+        if (!flavors.isDowngradeAllowed()) {
+            ServiceFlavor newServiceFlavor = getServiceFlavorWithName(newFlavor, flavors);
+            ServiceFlavor originalServiceFlavor = getServiceFlavorWithName(originalFlavor, flavors);
+            if (newServiceFlavor.getPriority() < originalServiceFlavor.getPriority()) {
```
swaroopar, line 333: should be opposite.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/servicetemplate/ServiceFlavor.java
```diff
@@ -34,4 +34,7 @@ public class ServiceFlavor implements Serializable {
...
+    private int priority;
```
swaroopar, line 39: would be good if we can have a check that priority minimum allowed value is "1"

samples/huawei-Kafka-autofill.yml
```diff
@@ -103,6 +105,7 @@ flavors:
   modificationImpact:
     isDataLost: true
     isServiceInterrupted: true
+  isDowngradeAllowed: false
```
swaroopar, line 108: default value would be true. So it is better to have maximum examples with "true" value.

swaroopar (changes requested):

modules/models/src/main/java/org/eclipse/xpanse/modules/models/servicetemplate/ServiceFlavor.java
baixinsui: Updated.

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java
baixinsui: Updated.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/servicetemplate/ServiceFlavor.java
baixinsui: Updated.

swaroopar (approved):
LGTM

PR #1618, opened 2024-05-15
set default lock config for new deploy service

### LINKED ISSUE #1617: Error in clicking the locks option when lockConfig is null
When the `lockConfig` of the service is null, an error occurs on UI page when clicking on the `locks` option of the service.
image
We can fix it the following way:
Set a default value `{"isDestroyLocked": "false", "isModifyLocked": "false"}` for the field `lockConfig` of the new `DeployServiceEntity` in the database.

baixinsui:
fixes #1617 

image


swaroopar (approved):
LGTM

PR #1647, opened 2024-05-27
Add API to obtain the prices of all flavors of the service

baixinsui:
fixes https://github.com/eclipse-xpanse/xpanse/issues/1645

Get the price of one specific flavor of the service.
image

Get the prices of all flavors of the service.
image


Alice1319:
LGTM

swaroopar (approved):

PR #1657, opened 2024-05-30
add auto-approve feature configurable per CSP

baixinsui:
fixes https://github.com/eclipse-xpanse/xpanse/issues/1652

Huawei auto approve service templates is enable:
Register huawei service template with state  `approved`:
msedge_sQLsNNjDOm
Update registered huawei service template with state `approved`:
ShareX_cJpHQrRAxR

FlexibleEngine auto approve servie templates is disable:
Register flexibleEngine service template with state  `approval pending`:
msedge_HJE2LpDfiP
Update registered flexibleEngine service template with state `approval pending`:
chrome_D8yDKSMaEV


modules/servicetemplate/src/main/java/org/eclipse/xpanse/modules/servicetemplate/config/CspAutoApproveConfig.java
```diff
@@ -0,0 +1,46 @@
...
+
+import org.eclipse.xpanse.modules.models.common.enums.Csp;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+
+/**
+ * Configuration for CSP auto approve service templates.
+ */
+@Component
+public class CspAutoApproveConfig {
```
swaroopar, line 17: @baixinsui i would avoid such common class for all CSPs. 
can we instead make this as a plugin interface method -> the method returns the config parameter.

swaroopar (changes requested):

Alice1319:
LGTM

swaroopar (approved):
LGTM

PR #1658, opened 2024-05-30
add caching for getAvailabilityZones

### LINKED ISSUE #1634: add caching for getAvailabilityZones
Add caching for getAvailabilityZones request as this data will not change frequently. 

- [ ] Implement caching using caffeine cache. 
- [ ] The cache should be kept for 60mins. This value should be configurable. 
- [ ] Return HTTP cache control headers from spring controller

WangLiNaruto:
Fixed #1634 
image
image
image
image

### DIFF
--- a/modules/api/src/main/java/org/eclipse/xpanse/api/controllers/ServiceDeployerApi.java
+++ b/modules/api/src/main/java/org/eclipse/xpanse/api/controllers/ServiceDeployerApi.java
@@ -21,6 +21,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 import java.util.UUID;
+import java.util.concurrent.TimeUnit;
 import lombok.extern.slf4j.Slf4j;
 import org.eclipse.xpanse.api.config.AuditApiRequest;
 import org.eclipse.xpanse.modules.database.service.DeployServiceEntity;
@@ -41,8 +42,11 @@ import org.eclipse.xpanse.modules.orchestrator.OrchestratorPlugin;
 import org.eclipse.xpanse.modules.orchestrator.PluginManager;
 import org.eclipse.xpanse.modules.orchestrator.deployment.DeployTask;
 import org.eclipse.xpanse.modules.security.UserServiceHelper;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.http.CacheControl;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.MediaType;
+import org.springframework.http.ResponseEntity;
 import org.springframework.security.access.AccessDeniedException;
 import org.springframework.security.access.annotation.Secured;
 import org.springframework.web.bind.annotation.CrossOrigin;
@@ -67,6 +71,9 @@ import org.springframework.web.bind.annotation.RestController;
 @CrossOrigin
 @Secured({ROLE_ADMIN, ROLE_USER})
 public class ServiceDeployerApi {
+
+    @Value("${region.azs.cache.expire.time.in.minutes:60}")
+    private long duration;
 
     @Resource
     private DeployService deployService;
@@ -308,7 +315,7 @@ public class ServiceDeployerApi {
     @ResponseStatus(HttpStatus.OK)
     @Secured({ROLE_ADMIN, ROLE_CSP, ROLE_ISV, ROLE_USER})
     @AuditApiRequest(methodName = "getCspFromRequestUri")
-    public List<String> getAvailabilityZones(
+    public ResponseEntity<List<String>> getAvailabilityZones(
             @Parameter(name = "cspName", description = "name of the cloud service provider")
             @RequestParam(name = "cspName") Csp csp,
             @Parameter(name = "regionName", description = "name of the region")
@@ -316,8 +323,11 @@ public class ServiceDeployerApi {
 
         String currentUserId = this.userServiceHelper.getCurrentUserId();
         OrchestratorPlugin orchestratorPlugin = pluginManager.getOrchestratorPlugin(csp);
-        return orchestratorPlugin.getAvailabilityZonesOfRegion(
+        List<String> availabilityZones = orchestratorPlugin.getAvailabilityZonesOfRegion(
                 currentUserId, regionName);
+        CacheControl cacheControl =
+                CacheControl.maxAge(duration, TimeUnit.MINUTES).mustRevalidate();
+        return ResponseEntity.ok().cacheControl(cacheControl).body(availabilityZones);
     }
 
 }
--- a/modules/cache/pom.xml
+++ b/modules/cache/pom.xml
@@ -1,0 +1,37 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ SPDX-License-Identifier: Apache-2.0
+  ~ SPDX-FileCopyrightText: Huawei Inc.
+  ~
+  -->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.eclipse.xpanse</groupId>
+    <artifactId>modules</artifactId>
+    <version>1.0.15-SNAPSHOT</version>
+    <relativePath>../pom.xml</relativePath>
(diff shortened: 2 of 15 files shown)


modules/api/src/main/java/org/eclipse/xpanse/api/config/CaffeineCacheConfig.java
```diff
@@ -0,0 +1,40 @@
...
+import org.springframework.context.annotation.Configuration;
+
+/**
+ * The configuration class of Caffeine Cache.
+ */
+@Configuration
+@EnableCaching
+public class CaffeineCacheConfig {
+
+    @Value("${caffeine.cache.expire.time}")
```
baixinsui, line 24: It is best to add default values to avoid the configuration item lost

modules/api/src/main/java/org/eclipse/xpanse/api/config/CaffeineCacheConfig.java
```diff
@@ -0,0 +1,40 @@
...
+    private long duration;
+
+    /**
+     * Create the configured CaffeineCacheManager.
+     *
+     * @return CaffeineCacheManager
+     */
+    @Bean
+    public CaffeineCacheManager caffeineCacheManager() {
+        CaffeineCacheManager cacheManager = new CaffeineCacheManager("caffeine_cache");
```
baixinsui, line 34: It's better to replace `caffeine_cache` with a meaningful name, like `region_azs_cache`.

baixinsui (changes requested):

runtime/src/main/java/org/eclipse/xpanse/runtime/XpanseApplication.java
```diff
@@ -21,6 +22,7 @@
 /**
  * Main entry class to xpanse runtime. This class can be directly executed to start the server.
  */
+@EnableCaching
```
baixinsui, line 25: The annotation here is duplicated with the annotation above class `CaffeineCacheConfig`. Is it necessary?

Alice1319:
LGTM

runtime/src/main/resources/application.properties
```diff
@@ -45,4 +45,5 @@ otel.exporter.otlp.enabled=false
 spring.activiti.history-level=full
 git.command.timeout.seconds=10
 http.request.retry.max.attempts=5
-http.request.retry.delay.milliseconds=1000
\ No newline at end of file
+http.request.retry.delay.milliseconds=1000
+caffeine.cache.expire.time=60
```
baixinsui, line 49: The `caffeine.cache.expire.time` is used in two different class, one used time unit is second but another is minute.If the time of these two configurations is different, it is better to use two different configuration items.

modules/api/src/main/java/org/eclipse/xpanse/api/controllers/ServiceDeployerApi.java
```diff
@@ -68,6 +72,9 @@
 @Secured({ROLE_ADMIN, ROLE_USER})
 public class ServiceDeployerApi {
 
+    @Value("${caffeine.cache.expire.time:60}")
```
baixinsui, line 75: It is better to add time uint in the configuration item name, like 'caffeine.cache.expire.seconds'.

modules/api/src/main/java/org/eclipse/xpanse/api/config/CaffeineCacheConfig.java
```diff
@@ -0,0 +1,38 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+/**
+ * The configuration class of Caffeine Cache.
+ */
+@Configuration
+public class CaffeineCacheConfig {
+
+    @Value("${caffeine.cache.expire.time:60}")
```
baixinsui, line 22: It is better to add time uint in the configuration item name, like 'caffeine.cache.expire.minutes'.

modules/api/src/main/java/org/eclipse/xpanse/api/controllers/ServiceDeployerApi.java
swaroopar: Please fix this comment. 
something like this "region.azs.cache.expire.time.in.minutes"

modules/api/src/main/java/org/eclipse/xpanse/api/config/cache/CaffeineCacheConfig.java
```diff
@@ -0,0 +1,38 @@
...
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.cache.caffeine.CaffeineCacheManager;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+/**
+ * The configuration class of Caffeine Cache.
+ */
+@Configuration
+public class CaffeineCacheConfig {
```
swaroopar, line 20: since this cache manager is just for AZ cache. We need a specific naming. 
Also it is better we move this class to "cache" maven module.

modules/api/src/main/java/org/eclipse/xpanse/api/controllers/ServiceDeployerApi.java
```diff
@@ -308,16 +315,19 @@ public Response purge(@PathVariable("id") String id) {
     @ResponseStatus(HttpStatus.OK)
     @Secured({ROLE_ADMIN, ROLE_CSP, ROLE_ISV, ROLE_USER})
     @AuditApiRequest(methodName = "getCspFromRequestUri")
-    public List<String> getAvailabilityZones(
+    public ResponseEntity<List<String>> getAvailabilityZones(
```
swaroopar, line 318: can you please confirm if this change does not impact the response model on swagger UI?

WangLiNaruto: @swaroopar Hi,I have tested and currently, it will not have any impact on the front end

modules/cache/pom.xml
```diff
@@ -0,0 +1,41 @@
...
+    <dependency>
+      <groupId>com.github.ben-manes.caffeine</groupId>
+      <artifactId>caffeine</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-starter-cache</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.springframework</groupId>
```
swaroopar, line 37: this is needed for which purpose?

modules/cache/src/main/java/org/eclipse/xpanse/modules/cache/RegionAzCacheConfig.java
```diff
@@ -0,0 +1,38 @@
...
+    private long duration;
+
+    /**
+     * Create the configured CaffeineCacheManager.
+     *
+     * @return CaffeineCacheManager
+     */
+    @Bean
+    public CaffeineCacheManager caffeineCacheManager() {
+        CaffeineCacheManager cacheManager = new CaffeineCacheManager("region_azs_cache");
```
swaroopar, line 32: can we move this cache manager name to a constants file in cache module? then it can be referred to from other modules instead of rewriting this string in multiple places.

swaroopar (changes requested):

modules/cache/src/main/java/org/eclipse/xpanse/modules/cache/consts/RegionAzCacheConstant.java
```diff
@@ -0,0 +1,16 @@
...
+ * SPDX-FileCopyrightText: Huawei Inc.
+ *
+ */
+
+package org.eclipse.xpanse.modules.cache.consts;
+
+/**
+ * Region az information cache constant class.
+ */
+public class RegionAzCacheConstant {
```
swaroopar, line 12: we can call this as 'CacheNames' and it can hold names of all caches which we will have in xpanse

modules/cache/src/main/java/org/eclipse/xpanse/modules/cache/consts/RegionAzCacheConstant.java
```diff
@@ -0,0 +1,16 @@
...
+
+    public static final String REGION_AZ_CACHE_MANAGER_NAME = "region_azs_cache";
```
swaroopar, line 14: REGION_AZ_CACHE_MANAGER_NAME - REGION_AZ_CACHE_NAME

swaroopar (changes requested):

swaroopar (approved):

PR #1675, opened 2024-06-03
SDK client retry strategy use default max retry attempts

baixinsui:
fixes https://github.com/eclipse-xpanse/xpanse/issues/1672

### DIFF
--- a/plugins/huaweicloud/src/main/java/org/eclipse/xpanse/plugins/huaweicloud/manage/HuaweiCloudVmStateManager.java
+++ b/plugins/huaweicloud/src/main/java/org/eclipse/xpanse/plugins/huaweicloud/manage/HuaweiCloudVmStateManager.java
@@ -6,6 +6,9 @@
 
 package org.eclipse.xpanse.plugins.huaweicloud.manage;
 
+
+import static org.eclipse.xpanse.plugins.huaweicloud.common.HuaweiCloudRetryStrategy.WAITING_JOB_SUCCESS_RETRY_TIMES;
+
 import com.huaweicloud.sdk.core.auth.ICredential;
 import com.huaweicloud.sdk.ecs.v2.EcsClient;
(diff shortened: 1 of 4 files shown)


Alice1319:
LGTM

plugins/huaweicloud/src/main/java/org/eclipse/xpanse/plugins/huaweicloud/manage/HuaweiCloudVmStateManager.java
```diff
@@ -142,7 +142,7 @@ private EcsClient getEcsClient(ServiceStateManageRequest serviceStateManageReque
 
     private boolean checkEcsExecResultByJobId(EcsClient ecsClient, String jobId) {
         ShowJobResponse response = ecsClient.showJobInvoker(new ShowJobRequest().withJobId(jobId))
-                .retryTimes(huaweiCloudRetryStrategy.getRetryMaxAttempts())
+                .retryTimes(huaweiCloudRetryStrategy.getSdkRetryMaxAttempts())
```
swaroopar, line 145: hi @baixinsui I think we are still mixing two things. 
In the SDK there are two types of retry 
1. http.retry - which we do when the http request fails. 
2. batch job status - retry  - here we only check for the status of the job by polling the response status. 

We must separate these two retries.

swaroopar (changes requested):

plugins/huaweicloud/src/main/java/org/eclipse/xpanse/plugins/huaweicloud/manage/HuaweiCloudVmStateManager.java
baixinsui: I set a fixed max value to the retry times for waiting job success.

swaroopar (approved):
LGTM

PR #1676, opened 2024-06-03
add debug logging to HuaweiCloud SDK conditionally

baixinsui:
fixes https://github.com/eclipse-xpanse/xpanse/issues/1673

By default, the Huawei Cloud SDK only outputs access logs:
image

When configuring `huaweicloud.sdk.debug.http.listener.enabled=true`, the Huawei Cloud SDK will also output debug logs:
image

### DIFF
--- a/plugins/flexibleengine/src/main/java/org/eclipse/xpanse/plugins/flexibleengine/common/FlexibleEngineClient.java
+++ b/plugins/flexibleengine/src/main/java/org/eclipse/xpanse/plugins/flexibleengine/common/FlexibleEngineClient.java
@@ -36,6 +36,7 @@ import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.eclipse.xpanse.modules.models.monitor.exceptions.ClientApiCallFailedException;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Component;
 
 /**
@@ -44,6 +45,9 @@ import org.springframework.stereotype.Component;
 @Slf4j
 @Component
 public class FlexibleEngineClient extends FlexibleEngineCredentials {
+
+    @Value("${flexibleengine.sdk.enable.http.debug.logs:false}")
+    private boolean sdkHttpDebugLogsEnabled;
 
     @Resource
     private FlexibleEngineRetryStrategy retryStrategy;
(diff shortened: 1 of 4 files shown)


plugins/flexibleengine/pom.xml
```diff
@@ -93,10 +93,10 @@
             <version>${huaweicloud.sdk.version}</version>
         </dependency>
         <dependency>
-            <groupId>org.slf4j</groupId>
-            <artifactId>slf4j-api</artifactId>
+            <groupId>ch.qos.logback</groupId>
```
swaroopar, line 96: @baixinsui do we really need to have logback here as dependency? i thought having the SL4J as bridge should be sufficient.

runtime/src/main/resources/application.properties
```diff
@@ -50,4 +50,5 @@ huaweicloud.auto.approve.service.template.enabled=false
 flexibleengine.auto.approve.service.template.enabled=false
 openstack.auto.approve.service.template.enabled=false
 scs.auto.approve.service.template.enabled=false
-region.azs.cache.expire.time.in.minutes=60
\ No newline at end of file
+region.azs.cache.expire.time.in.minutes=60
+huaweicloud.sdk.debug.http.listener.enabled=false
```
swaroopar, line 54: i think we can rename it to - huaweicloud.sdk.enable.http.debug.logs

swaroopar (changes requested):

plugins/flexibleengine/pom.xml
baixinsui: Restore.

runtime/src/main/resources/application.properties
baixinsui: Updated.

plugins/flexibleengine/src/main/java/org/eclipse/xpanse/plugins/flexibleengine/common/FlexibleEngineClient.java
```diff
@@ -45,6 +46,9 @@
 @Component
 public class FlexibleEngineClient extends FlexibleEngineCredentials {
 
+    @Value("${huaweicloud.sdk.enable.http.debug.logs:false}")
```
swaroopar, line 49: @baixinsui its better to have two different configs for different plugins.

swaroopar (changes requested):

swaroopar (approved):
LGTM

PR #1680, opened 2024-06-04
Get the price from international website when the specific exception is caught

baixinsui:
fixes https://github.com/eclipse-xpanse/xpanse/issues/1677
Users in the Chinese website get the price with currency CNY:
chrome_tGDYwaW9QY
Users in the international website get the price with currency USD:
image

### DIFF
--- a/plugins/huaweicloud/src/main/java/org/eclipse/xpanse/plugins/huaweicloud/common/HuaweiCloudClient.java
+++ b/plugins/huaweicloud/src/main/java/org/eclipse/xpanse/plugins/huaweicloud/common/HuaweiCloudClient.java
@@ -8,6 +8,8 @@ package org.eclipse.xpanse.plugins.huaweicloud.common;
 
 import com.huaweicloud.sdk.bss.v2.BssClient;
 import com.huaweicloud.sdk.bss.v2.region.BssRegion;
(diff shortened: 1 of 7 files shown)


Alice1319:
LGTM

plugins/huaweicloud/src/main/java/org/eclipse/xpanse/plugins/huaweicloud/common/HuaweiCloudClient.java
```diff
@@ -125,6 +134,45 @@ public BssClient getBssClient(ICredential globalCredential) {
...
+    /**
+     * Get HuaweiCloud Bss intl Client.
+     *
+     * @param globalCredential ICredential
+     */
+    public BssintlClient getBssintlClient(ICredential globalCredential) {
+        return BssintlClient.newBuilder()
+                .withHttpConfig(getHttpConfig())
+                .withCredential(globalCredential)
+                .withRegion(BssintlRegion.AP_SOUTHEAST_1)
```
swaroopar, line 146: @baixinsui is this hardcoding of region necessary? Because prices are region dependent. 

If this is really necessary, then we must add a comment on why we do it.

swaroopar (changes requested):

swaroopar (approved):
LGTM

PR #1692, opened 2024-06-06
Add API to redeploy services with failed state

### LINKED ISSUE #1687: introduce retryFailedDeployment method to retry the failed order
In case of new deployment and migration, if the deployment fails then the user can simply send the failed serviceId again and then this can retry the same deployment again. 

- [ ] create new API method - 'retryFailedDeployment' - /deploy/retry/${serviceId}
- [ ] Request must be accepted only if the service is in "failed" state. 
- [ ] The service must do the deployment with the same service parameters which the user had sent in the original request.

baixinsui:
fixes #1687 
New API to redeploy service with id:
image

Redeploy the service not with failed state return 400:
msedge_tw2HVQ8JKy
Redeploy the service with failed state return 202:
msedge_6lDGUsKGKP


swaroopar (approved):
LGTM

PR #1702, opened 2024-06-11
Add unique values in list validation for OCL

### LINKED ISSUE #1696: Add unique values in list validation for OCL
In OCL, the regions list currently allows us to provide  name multiple times. Such list entries must be checked to ensure that the list contains unique values. 
Similar to `region`, we must also check other list fields in OCL where the values must be unique - Eg., deploy variables, AZs, ServiceProviderContactDetails, etc

- [ ] Add validation - consider using using https://stackoverflow.com/questions/47293514/bean-validation-jsr-303-duplicateitemcheck-for-a-list-in-java
- [ ] Throw exception - ListElementsNotUniqueException - with message containing exact field name and which value is duplicate.
- [ ] update description of the API fields to mention that the list elements must be unique.

baixinsui:
fixes #1696


modules/models/src/main/java/org/eclipse/xpanse/modules/models/servicetemplate/Deployment.java
```diff
@@ -32,14 +35,21 @@ public class Deployment implements Serializable {
...
     private CredentialType credentialType = CredentialType.VARIABLES;
 
-    @Schema(description = "The list of availability zones of the service.")
-    private List<AvailabilityZoneConfig> serviceAvailability;
+    @Valid
+    @Size(min = 1)
+    @UniqueElements
+    @Schema(description = "The list of availability zones of the service."
+            + "The list elements must be unique.")
+    private List<AvailabilityZoneConfig> serviceAvailabilities;
```
swaroopar, line 52: the word service availabilities is not so correct in this context. Even though it is a list, it is more of a configuration that works together to ensure availability of the complete service
We can rename it to serviceAvailbilityConfig or keep it like before.

swaroopar (changes requested):

modules/models/src/main/java/org/eclipse/xpanse/modules/models/servicetemplate/Deployment.java
baixinsui: updated.

swaroopar (approved):

PR #1707, opened 2024-06-12
add serviceId to getAvailabilityZones method

### LINKED ISSUE #1704: add serviceId to getAvailabilityZones method
In some plugins (Eg., SCS), the cloud endpoints are part of the service template configuration. So it is necessary to add optional "serviceId" parameter to the getAvailabilityZones method.

- [ ] add optional parameter `serviceId` to `getAvailabilityZones` method.
- [ ] In case of SCS and Openstack plugins, use the service Id 
https://github.com/eclipse-xpanse/xpanse/blob/69f7966b267026d58669a4144fd26c0e04a0e685/plugins/scs/src/main/java/org/eclipse/xpanse/plugins/scs/manage/ScsResourceManager.java#L230

baixinsui:
fixes #1704


swaroopar (approved):
LGTM

PR #1712, opened 2024-06-14
add caching for getServicePriceByFlavor service

### LINKED ISSUE #1699: add caching for getServicePriceByFlavor service
- [ ] Add new cache manager for getServicePriceByFlavor
- [ ] Add @cacheable annotation to getServicePriceByFlavor
- [ ] Change getPricesByService to call getServicePriceByFlavor instead of the  method calling the plugins directly.

baixinsui:
fixes #1699 
Called service `getServicePriceByFlavor` firstly without cache:
Insomnia_AmDe32TmFJ
Called service `getServicePriceByFlavor` again with cache:
msedge_XJmnV8r8Uj

Called service `getPricesByService` firstly without cache:
chrome_kjpqZYBApG
Called service `getPricesByService` again with cache:
chrome_xMEkqI9v3r


WangLiNaruto:
LGTM

modules/cache/src/main/java/org/eclipse/xpanse/modules/cache/RegionAzCacheConfig.java
```diff
@@ -28,13 +33,16 @@ public class RegionAzCacheConfig {
      *
      * @return CaffeineCacheManager
      */
-    @Bean
-    public CaffeineCacheManager caffeineCacheManager() {
-        CaffeineCacheManager cacheManager = new CaffeineCacheManager(
-                CacheNames.REGION_AZ_CACHE_NAME);
+    @Primary
```
swaroopar, line 36: can we remove the `@primary` annotation here? this can have some unexpected issues.

swaroopar (changes requested):

modules/cache/src/main/java/org/eclipse/xpanse/modules/cache/RegionAzCacheConfig.java
baixinsui: Removing the `@primary`  will cause the following exception to be thrown when the application is starting:
idea64_zgxiNz1lIh

baixinsui: No. Removing the @primary will cause exception as following:

baixinsui: No. Removing the @primary will cause the exception as below:
image
I tried to add a special name to credential cache management, but it still failed until the @primary was used.

swaroopar (approved):
LGTM

PR #1725, opened 2024-06-20
handle multiple cloud providers based on same cloud implementation

### LINKED ISSUE #1709: handle multiple cloud providers based on same cloud implementation
In order to handle multiple clouds based on same cloud implementation - Eg., Regio Cloud and Plus Server are both SCS based clouds,
Also multiple clouds can have Openstack based clouds. In fact, SCS itself is based on Openstack. 
To leads to new requirements - 
1. Plugins should be based on CSPs and not implementations
2. Duplicate code must be removed. 

- [ ] Create new package in plugins -> common and under common one more package - openstack-common
- [ ] Use openstack-common code in openstack and SCS plugins .
- [ ] update CSP enums to have new values - RegioCloud, Plus Server, Openstack-Testlab
- [ ] Remove CSP enums -> SCS, Openstack. 
- [ ] Add plugin code  for  RegioCloud, Plus Server, openstack-testlab (rename openstack)
- [ ] For each of the openstack and SCS based clouds, we need a mandatory required property for the keystone URL 
   https://github.com/swaroopar/osc/blob/0f53ec967c5e24189748045520a168839a3ff945/plugins/openstack/src/main/java/org/eclipse/xpanse/plugins/openstack/OpenstackOrchestratorPlugin.java#L96

- [ ] This required property must be unique. No plugin must have a mandatory field `OS_AUTH_URL`.  Because this is a common name and when an env var is set with this name, then all plugins (even Huawei Cloud) use this var and then causes unexpected issues. 
 example - openstack_testlab_url, regiocloud_url, etc 
This must be mapped to OS_AUTH_URL in the terraform deployer. 
- [ ] add service Id to getExistingResourceNamesWithKind method as well.

baixinsui:
fixes #1709 
related to https://github.com/eclipse-xpanse/xpanse-ui/pull/915
Deploy service of the provider `OpenstackTestlab`:
image
image

Get metrics of the service of the provider `OpenstackTestlab`
image
`


baixinsui:
@swaroopar 
I have only completed code refactoring and have not tested it yet. Could you please take a look at what needs to be modified first.

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/utils/DeployEnvironments.java
```diff
@@ -113,10 +117,25 @@ private Map<String, String> getEnv(Map<String, Object> serviceRequestProperties,
                                 ? aesUtil.decode(variable.getValue()) : variable.getValue());
             }
         }
-
+        setProviderOsAuthUrl(csp, variables);
         return variables;
     }
 
+    private void setProviderOsAuthUrl(Csp csp, Map<String, String> variables) {
```
swaroopar, line 124: this logic must be part of plugin of each cloud provider.
the Orchestrator must only ask the corresponding plugin for these variables and the translation must come from it.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/common/enums/Csp.java
```diff
@@ -14,14 +14,15 @@
...
-    AWS("aws"),
-    AZURE("azure"),
-    GOOGLE("google");
+    HUAWEI_CLOUD("HuaweiCloud"),
+    FLEXIBLE_ENGINE("FlexibleEngine"),
+    OPENSTACK_TESTLAB("OpenstackTestlab"),
+    PLUS_SERVER("PlusServer"),
+    REGIO_CLOUD("RegioCloud"),
+    ALIBABA_CLOUD("AlibabaCloud"),
+    AMAZON_CLOUD("AmazonCloud"),
```
swaroopar, line 23: this should be just `aws`

modules/models/src/main/java/org/eclipse/xpanse/modules/models/common/enums/Csp.java
```diff
@@ -14,14 +14,15 @@
...
+    AZURE_CLOUD("AzureCloud"),
+    GOOGLE_CLOUD("GoogleCloud");
```
swaroopar, line 25: this should be `GCP("GoogleCloudPlatform")`

modules/models/src/main/java/org/eclipse/xpanse/modules/models/common/enums/Csp.java
```diff
@@ -14,14 +14,15 @@
...
-    AZURE("azure"),
-    GOOGLE("google");
+    HUAWEI_CLOUD("HuaweiCloud"),
+    FLEXIBLE_ENGINE("FlexibleEngine"),
+    OPENSTACK_TESTLAB("OpenstackTestlab"),
+    PLUS_SERVER("PlusServer"),
+    REGIO_CLOUD("RegioCloud"),
+    ALIBABA_CLOUD("AlibabaCloud"),
+    AMAZON_CLOUD("AmazonCloud"),
+    AZURE_CLOUD("AzureCloud"),
```
swaroopar, line 24: this should be just `azure`

modules/orchestrator/src/main/java/org/eclipse/xpanse/modules/orchestrator/PluginManager.java
```diff
@@ -46,17 +46,20 @@ public void loadPlugins() {
...
         });
         checkCredentialDefinitions();
-        log.info("Cloud service providers:{} with activated plugins.", pluginsMap.keySet());
+        log.info("Cloud service providers {} with activated plugins.", pluginsMap.keySet());
         if (pluginsMap.isEmpty()) {
             throw new PluginNotFoundException("No plugin is activated.");
         }
+        if (pluginsMap.size() > 1) {
+            String osAuthUrl = applicationContext.getEnvironment().getProperty("OS_AUTH_URL");
+            if (Objects.nonNull(osAuthUrl)) {
```
swaroopar, line 58: The logic here is fine. But we need another way to get the list of of "black listed" environment variables. 
We can have this is a list in properties file.

plugins/common/src/main/java/org/eclipse/xpanse/plugins/common/opensatckcommon/auth/ProviderAuthInfoResolver.java
```diff
@@ -0,0 +1,159 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * SPDX-FileCopyrightText: Huawei Inc.
+ */
+
+package org.eclipse.xpanse.plugins.common.opensatckcommon.auth;
```
swaroopar, line 6: package name can be just - openstack

plugins/huawei-cloud/pom.xml
```diff
@@ -19,7 +19,7 @@
     </parent>
 
     <groupId>org.eclipse.xpanse.plugins</groupId>
-    <artifactId>huaweicloud</artifactId>
+    <artifactId>huawei-cloud</artifactId>
```
swaroopar, line 22: we must keep the old name itself.  same applies for flexibleengine, regiocloud, plusserver.. they are all just one single word

swaroopar (changes requested):

swaroopar (changes requested):

swaroopar (approved):
LGTM

PR #1744, opened 2024-06-26
refactor all occurrences of findAll() JPA repository method

### LINKED ISSUE #1724: refactor all occurrences of findAll() JPA repository method which is an anti pattern
Currently there are some occurecues of JPA's `findAll()` method which is an anti-pattern since it loads the full table from DB into the java memory. 

More details - https://vladmihalcea.com/spring-data-findall-anti-pattern/

Preferred solution -

- [ ] As mentioned the above blog, create a custom repository with only the methods are necessary for xpanse. 
- [ ] Create an architecture test to ensure that no repository is created with inbuilt repository methods.  (Check existing architecture tests)

baixinsui:
fixes #1724 
image

### DIFF
--- a/modules/database/src/main/java/org/eclipse/xpanse/modules/database/CustomJpaRepository.java
+++ b/modules/database/src/main/java/org/eclipse/xpanse/modules/database/CustomJpaRepository.java
@@ -1,0 +1,31 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * SPDX-FileCopyrightText: Huawei Inc.
+ *
+ */
+
+package org.eclipse.xpanse.modules.database;
+
+import java.util.Optional;
+import org.springframework.data.repository.NoRepositoryBean;
+import org.springframework.data.repository.Repository;
+
+/**
+ * Custom repository interface for database operations.
+ */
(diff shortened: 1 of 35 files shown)


modules/database/src/main/java/org/eclipse/xpanse/modules/database/CustomJpaRepository.java
```diff
@@ -0,0 +1,32 @@
...
+
+import java.util.Optional;
+import org.springframework.data.repository.NoRepositoryBean;
+import org.springframework.data.repository.Repository;
+
+/**
+ * Custom repository interface for database operations.
+ */
+@NoRepositoryBean
+//CHECKSTYLE OFF: InterfaceTypeParameterName
```
swaroopar, line 17: @baixinsui instead we can change this rule to allow any uppercase multi-character strings as parameter name. 
https://checkstyle.sourceforge.io/checks/naming/interfacetypeparametername.html
Then the rule will be good for the entire project.

swaroopar (changes requested):

swaroopar (approved):
LGTM

PR #1756, opened 2024-06-28
deployment Apis return all exceptions

### LINKED ISSUE #1721: deploy/redeploy always returns success even if there is a failure.
Currently we are catching all exceptions that are returned as part of the "synchronous" process and return an "accepeted 202" always. 

https://github.com/swaroopar/osc/blob/38bef5dbc922cbc991d4b387ae9b2cb43f3253a8/modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java#L244

This must be changed as follows

- [ ] All exceptions generated as part of the synchronous request must be caught and returned as it is to client with error HTTP codes and correct exceptions. 
- [ ] The exception must be stored to DB as 'DeployResult' as before. 
- [ ] Implement change to all services that involve async methods - deploy, redeploy, purge, destroy, etc

baixinsui:
fixes #1721 
Deploy service throws exception:
chrome_S8s5vqTYui
Deployment failed service in list:
image


swaroopar (approved):
LGTM

PR #1763, opened 2024-07-01
Delete controller cache to prevent return errors

### LINKED ISSUE #1760: issue with pricing and AZ api caching
Currently we are setting caching control even when the request is not cached. Due to this, the client is caching even error responses.

baixinsui:
fixes #1760 

Return errors with error credential:
msedge_GPRrEjmMMU
Return correct data with correct credential not cached errors:
msedge_7QvafA4eiK

### DIFF
--- a/modules/api/src/main/java/org/eclipse/xpanse/api/controllers/ServiceDeployerApi.java
+++ b/modules/api/src/main/java/org/eclipse/xpanse/api/controllers/ServiceDeployerApi.java
@@ -350,13 +350,18 @@ public class ServiceDeployerApi {
             @RequestParam(name = "regionName") String regionName,
             @Parameter(name = "serviceId", description = "id of the deployed service")
             @RequestParam(name = "serviceId", required = false) String serviceId) {
-
-        UUID uuid = StringUtils.isBlank(serviceId) ? null : UUID.fromString(serviceId);
-        String currentUserId = this.userServiceHelper.getCurrentUserId();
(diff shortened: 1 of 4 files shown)


modules/api/src/main/java/org/eclipse/xpanse/api/controllers/ServiceDeployerApi.java
```diff
@@ -354,14 +347,8 @@ public ResponseEntity<List<String>> getAvailabilityZones(
         UUID uuid = StringUtils.isBlank(serviceId) ? null : UUID.fromString(serviceId);
         String currentUserId = this.userServiceHelper.getCurrentUserId();
         OrchestratorPlugin orchestratorPlugin = pluginManager.getOrchestratorPlugin(csp);
-        List<String> availabilityZones = orchestratorPlugin.getAvailabilityZonesOfRegion(
+        return orchestratorPlugin.getAvailabilityZonesOfRegion(
```
swaroopar, line 350: @baixinsui instead of completely returning CacheControl, can we still return it when we know that the response has no errors?
Advantage is, if we add cacheControl then the browser will send these requests again. (Response will be additionally cached in browser)

swaroopar (changes requested):

modules/api/src/main/java/org/eclipse/xpanse/api/controllers/ServiceDeployerApi.java
baixinsui: Updated. Remove the cache control when catch exceptions

swaroopar (approved):
LGTM

PR #1768, opened 2024-07-01
return string instead of enum in exception message

swaroopar:


modules/credential/src/main/java/org/eclipse/xpanse/modules/credential/CredentialCenter.java
```diff
@@ -221,7 +221,8 @@ public AbstractCredentialInfo getCredential(Csp csp,
                 csp, credentialType, userId);
         if (credentialInfos.isEmpty()) {
             throw new CredentialsNotFoundException(
-                    String.format("No credential information found for the given Csp:%s.", csp));
+                    String.format("No credential information found for the given Csp:%s.",
+                            csp.toValue()));
```
baixinsui, line 225: There are other uses that need to be modified in the class `PluginManager`.

swaroopar: @baixinsui thanks for the hint. I have now corrected PluginManager class as well

iskey (approved):
LGTM

PR #1784, opened 2024-07-08
store initial configuration of the service during service deployment

### LINKED ISSUE #1774: store initial configuration of the service during service deployment
As part of service configuration update, we must start with storing initial service configuration properties during the service deployment. 

- [ ] Add new table SERVICE_CONFIGURATION with foreign Key as the ID of the DEPLOY_SERVICE table. and CONFIGURATION column 
- [ ] CONFIGURATION column stored configuration as JSON. 
- [ ] Add created and updated time columns 
- [ ] Extract initial configuration as below - 
    1. JSON Key is `name` of ServiceConfigurationParameter
    2. Value is the `initialValue` of ServiceConfigurationParameter
    3. If the same `name` is present in the flavor properties, then that value from the deployed flavor must be used as the initial value. 
- [ ] Use  kafka service properties from bitnami kafka docker image and update the service template
https://github.com/bitnami/containers/tree/main/bitnami/kafka#111-debian-9-r205-220-debian-9-r40-111-ol-7-r286-and-220-ol-7-r53
To start with, we can only provide the following configuration parameters in the template - KAFKA_CFG_MESSAGE_MAX_BYTES, KAFKA_CFG_LOG_DIRS, KAFKA_CFG_NUM_IO_THREADS, KAFKA_LOG_FLUSH_INTERVAL_MESSAGES, KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR

WangLiNaruto:
Fixed #1774
image
image
image

image
image
image


modules/database/src/main/java/org/eclipse/xpanse/modules/database/serviceconfiguration/ServiceConfigurationEntity.java
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * ServiceConfigurationEntity for persistence.
+ */
+@Table(name = "SERVICE_CONFIGURATION")
+@Entity
+@Data
+public class ServiceConfigurationEntity {
+
+    @Id
+    private UUID id;
```
baixinsui, line 33: It is better to rename `id` to `service_id` if the configs map is related to the deployed service.

modules/database/src/main/java/org/eclipse/xpanse/modules/database/serviceconfiguration/DatabaseServiceConfigurationStorage.java
```diff
@@ -0,0 +1,73 @@
...
+
+
+    @Override
+    public ServiceConfigurationEntity storeAndFlush(
+            ServiceConfigurationEntity serviceConfigurationEntity) {
+        return serviceConfigurationRepository.saveAndFlush(serviceConfigurationEntity);
+    }
+
+    @Override
+    public List<ServiceConfigurationEntity> listServiceConfigurations(
```
baixinsui, line 44: If only the `id` is used as the query criterion, we can remove this method using the method findById(UUID id);

modules/database/src/main/java/org/eclipse/xpanse/modules/database/serviceconfiguration/ServiceConfigurationQueryModel.java
```diff
@@ -0,0 +1,19 @@
...
+import java.util.UUID;
+import lombok.Data;
+
+/**
+ * The query model for search ServiceConfigurationEntity.
+ */
+@Data
+public class ServiceConfigurationQueryModel {
+
+    private UUID id;
```
baixinsui, line 18: Removed it same time as deleting the medthod listServiceConfigurations(ServiceConfigurationQueryModel configQuery)

modules/database/src/main/java/org/eclipse/xpanse/modules/database/serviceconfiguration/ServiceConfigurationStorage.java
```diff
@@ -0,0 +1,49 @@
...
+     */
+    ServiceConfigurationEntity storeAndFlush(ServiceConfigurationEntity serviceConfigurationEntity);
+
+    /**
+     * Method to get stored database entries by query model.
+     *
+     * @param query service configuration query model.
+     * @return Returns all rows from the service configuration database table.
+     */
+    List<ServiceConfigurationEntity> listServiceConfigurations(
```
baixinsui, line 31: We can remove this method using  findServiceConfigurationById(UUID id);

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java
```diff
@@ -474,6 +499,7 @@ private void destroy(DeployTask destroyTask, DeployServiceEntity deployServiceEn
             destroyResult.setState(getDeployerTaskFailedState(destroyTask.getDeploymentScenario()));
             destroyResult.setMessage(errorMsg);
         }
+        deleteServiceConfiguration(destroyTask);
```
baixinsui, line 502: I think this deletion should be placed after the destory is successful

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java
```diff
@@ -310,17 +321,31 @@ public void redeployService(DeployServiceEntity serviceToRedeploy) {
...
-                                            DeployServiceEntity deployServiceEntity) {
+            DeployServiceEntity deployServiceEntity,
+            ServiceConfigurationEntity serviceConfigurationEntity) {
         log.info("Performing rollback of already provisioned resources.");
         if (Objects.nonNull(deployServiceEntity.getDeployResourceList())
                 && !deployServiceEntity.getDeployResourceList().isEmpty()) {
             log.info("Rollback to destroy created resources for service with ID: {}",
                     deployTask.getId());
             deployTask.setDeploymentScenario(DeploymentScenario.ROLLBACK);
+            if (Objects.nonNull(serviceConfigurationEntity)) {
```
baixinsui, line 332: This calling is duplicate with the calling `deleteServiceConfiguration(destroyTask);` in the method `destroy` (Line 502); because this method will call the method `destroy`.

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/utils/DeployTaskToServiceConfigurationEntityConverter.java
```diff
@@ -0,0 +1,57 @@
...
+import org.eclipse.xpanse.modules.database.serviceconfiguration.ServiceConfigurationEntity;
+import org.eclipse.xpanse.modules.models.servicetemplate.ServiceConfigurationParameter;
+import org.eclipse.xpanse.modules.models.servicetemplate.ServiceFlavorWithPrice;
+import org.eclipse.xpanse.modules.orchestrator.deployment.DeployTask;
+import org.springframework.util.CollectionUtils;
+
+/**
+ * Bean to create ServiceConfigurationEntity object from DeployTask.
+ */
+public class DeployTaskToServiceConfigurationEntityConverter {
```
swaroopar, line 21: would prefer to keep this also as a spring bean.

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/utils/DeployTaskToServiceConfigurationEntityConverter.java
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     * Method to create a ServiceConfigurationEntity from DeployTask.
+     *
+     * @param deployTask DeployTask.
+     * @return ServiceConfigurationEntity.
+     */
+    public static ServiceConfigurationEntity getServiceConfigurationEntity(DeployTask deployTask) {
```
swaroopar, line 29: the method must be - getInitialServiceConfiguration

modules/database/src/main/java/org/eclipse/xpanse/modules/database/serviceconfiguration/ServiceConfigurationEntity.java
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * ServiceConfigurationEntity for persistence.
+ */
+@Table(name = "SERVICE_CONFIGURATION")
+@Entity
+@Data
+public class ServiceConfigurationEntity {
+
+    @Id
+    @Column(name = "service_id", nullable = false)
```
swaroopar, line 33: This table is created as a separate table. Instead, we need this as a child table of DEPLOY_SERVICE with foreign key as SERVICE_ID. (The first requirement in the user story)

something like this https://www.baeldung.com/jpa-one-to-one#spk-strategy

swaroopar (changes requested):

modules/database/src/main/java/org/eclipse/xpanse/modules/database/serviceconfiguration/ServiceConfigurationEntity.java
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * ServiceConfigurationEntity for persistence.
+ */
+@Table(name = "SERVICE_CONFIGURATION")
+@Entity
+@Data
+public class ServiceConfigurationEntity {
+
+    @Id
+    private UUID id;
```
swaroopar, line 37: Do we really need an additional primary key here? Since there is only one-to-one mapping to the parent table, we can make the service_id column as the primary key which is anyways unique.

swaroopar (changes requested):

swaroopar (approved):
LGTM

PR #1820, opened 2024-07-24
Support Redis Cache and Refactor Credential Cache Using Cache Annotations

baixinsui:
Fixed https://github.com/eclipse-xpanse/xpanse/issues/1810

**Enable redis cache:**
Cache of  flavor price in redis:
chrome_NTGW3LJEs3
Cache credentials in redis:
After created:
chrome_B9dsLdYf5Z
After updated:
powershell_R5V1LxoSxL
After expired：
chrome_Auba8v5GSE
Cache monitor metrics in redis:
image

**Disable redis cache using caffeine cache:**
image


modules/cache/src/main/java/org/eclipse/xpanse/modules/cache/CaffeineCacheConfig.java
```diff
@@ -56,4 +69,32 @@ private Cache<Object, Object> getServiceFlavorPriceCache() {
...
+                    if (Objects.nonNull(key)) {
+                        log.info("Credential cache removed, key: {}, cause: {}", key, cause);
+                    }
+                })
+                .build();
+        return cache;
+    }
+
+    private Cache<Object, Object> getMonitorMetricsCache() {
+        Cache cache = Caffeine.newBuilder()
```
swaroopar, line 89: we can directly return the cache object here.

modules/cache/src/main/java/org/eclipse/xpanse/modules/cache/CaffeineCacheConfig.java
```diff
@@ -56,4 +69,32 @@ private Cache<Object, Object> getServiceFlavorPriceCache() {
                 : DEFAULT_CACHE_EXPIRE_TIME_IN_MINUTES;
         return Caffeine.newBuilder().expireAfterWrite(duration, TimeUnit.MINUTES).build();
     }
+
+
+    private Cache<Object, Object> getCredentialsCache() {
+        Cache cache = Caffeine.newBuilder()
```
swaroopar, line 75: we can directly return the cache object here.

modules/cache/src/main/java/org/eclipse/xpanse/modules/cache/CaffeineCacheConfig.java
```diff
@@ -56,4 +69,32 @@ private Cache<Object, Object> getServiceFlavorPriceCache() {
...
+                        log.info("Credential cache removed, key: {}, cause: {}", key, cause);
+                    }
+                })
+                .build();
+        return cache;
+    }
+
+    private Cache<Object, Object> getMonitorMetricsCache() {
+        Cache cache = Caffeine.newBuilder()
+                .expireAfterWrite(10, TimeUnit.MINUTES)
```
swaroopar, line 90: this expiry does not match the value in the below code. 
I think it is better to move this as a config value in properties file. 
infact, would be better to have this configuration in properties file.. similar to other caches 

https://github.com/eclipse-xpanse/xpanse/blob/abc2b1c8b62b13cdee5103ac230153818039b882/modules/cache/src/main/java/org/eclipse/xpanse/modules/cache/monitor/MonitorMetricsStore.java#L92

swaroopar (changes requested):

modules/cache/src/main/java/org/eclipse/xpanse/modules/cache/CaffeineCacheConfig.java
baixinsui: Updated.

modules/cache/src/main/java/org/eclipse/xpanse/modules/cache/CaffeineCacheConfig.java
baixinsui: Updated.

modules/cache/src/main/java/org/eclipse/xpanse/modules/cache/CaffeineCacheConfig.java
baixinsui: Updated as suggestion.

swaroopar (approved):
LGTM

PR #1843, opened 2024-08-01
add orderID to MDC context and log lines

### LINKED ISSUE #1840: add orderID to MDC context and log lines.
- [ ] Add new `ORDER_ID` field to log lines. 
- [ ] Rename TASK_ID to `SERVICE_ID` in the current configuration and code.

baixinsui:
Fixed #1840

### DIFF
--- a/modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/deployers/terraform/terraformboot/TerraformBootScriptValidator.java
+++ b/modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/deployers/terraform/terraformboot/TerraformBootScriptValidator.java
@@ -5,15 +5,16 @@
 
 package org.eclipse.xpanse.modules.deployment.deployers.terraform.terraformboot;
 
-import static org.eclipse.xpanse.modules.logging.CustomRequestIdGenerator.TASK_ID;
+import static org.eclipse.xpanse.modules.logging.LoggingKeyConstant.SERVICE_ID;
+import static org.eclipse.xpanse.modules.logging.LoggingKeyConstant.TRACKING_ID;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import java.util.Collections;
 import java.util.List;
-import java.util.Objects;
 import java.util.UUID;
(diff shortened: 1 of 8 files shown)


modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/deployers/terraform/terraformboot/TerraformBootScriptValidator.java
```diff
@@ -128,4 +123,15 @@ private List<String> getFilesByOcl(Deployment deployment) {
         String deployer = deployment.getDeployer();
         return Collections.singletonList(deployer);
     }
+
+    private UUID getRequestId() {
```
swaroopar, line 127: i think when we send requests to terraform-boot or tofu-maker the order ID is present. It is better to use that here.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/constant/LoggingKeyConstant.java
```diff
@@ -0,0 +1,27 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ * SPDX-FileCopyrightText: Huawei Inc.
+ */
+
+package org.eclipse.xpanse.modules.models.constant;
+
+/**
+ * Constant keys for logging.
+ */
+public class LoggingKeyConstant {
```
swaroopar, line 11: would prefer to keep this class in the logging module.

swaroopar (changes requested):

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/deployers/terraform/terraformboot/TerraformBootScriptValidator.java
baixinsui: This method is used in register service template, the order id always is null.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/constant/LoggingKeyConstant.java
baixinsui: Moved

swaroopar (approved):
LGTM

PR #1854, opened 2024-08-06
Return new Response object to store order and service id

baixinsui:
Fixed https://github.com/eclipse-xpanse/xpanse/issues/1839
image


modules/models/src/main/java/org/eclipse/xpanse/modules/models/response/Response.java
```diff
@@ -33,6 +35,9 @@ public class Response {
     @Schema(description = "Describes if the request is successful")
     private Boolean success;
 
+    @Schema(description = "Describes if the request is successful")
```
swaroopar, line 38: description to be updated.
unique ID to track the request on server.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/response/Response.java
```diff
@@ -33,6 +35,9 @@ public class Response {
...
+    private String trackingId;
```
swaroopar, line 39: I think it is better to keep this as response header "X-TRACKING-ID' then we can send it even in positive responses as well. That's how a lot of other API systems have implemented it. 

If this a bigger change, we can keep this as it is in this PR and change it to a HTTP response header in another user story.

swaroopar (changes requested):

modules/models/src/main/java/org/eclipse/xpanse/modules/models/response/ResultType.java
```diff
@@ -42,7 +42,7 @@ public enum ResultType {
     SERVICE_DEPLOYMENT_NOT_FOUND("Service Deployment Not Found"),
     RESOURCE_NOT_FOUND("Resource Not Found"),
     DEPLOYMENT_VARIABLE_INVALID("Deployment Variable Invalid"),
-    DEPLOYMENT_FAILED("Deployment Failed"),
+    SERVICE_ORDER_INIT_FAILED("Service Order Init Failed"),
```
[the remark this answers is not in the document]
baixinsui: Removed this unused enumeration.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/response/Response.java
baixinsui: Move the `trackingId`  into Headers

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java
```diff
@@ -93,13 +107,160 @@ public class DeployService {
...
+
+    /**
+     * Get user managed service.
+     *
+     * @param serviceId deployed service id.
+     * @param errorMsg  the error message.
+     * @return DeployServiceEntity.
+     * @throws AccessDeniedException if the current user is not the owner of the service.
+     */
+    private DeployServiceEntity getUserManagedService(UUID serviceId, String errorMsg) {
```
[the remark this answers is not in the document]
baixinsui: Updated.

modules/models/src/main/java/org/eclipse/xpanse/modules/models/response/Response.java
baixinsui: The tracking id moved into headers.
Error Response:
idea64_x2wNMwZD7K
Successful Response:
idea64_0QozROUWCs

swaroopar (approved):
LGTM

PR #1862, opened 2024-08-08
avoid retries for auth errors in cloud API requests

### LINKED ISSUE #1848: avoid retries for 401 and 403 HTTP errors in cloud API requests
Currently the plugins calling the cloud APIs are retrying all failures. But the permanent errors must not be retried. 

- [ ] Disable retries for 401 and 403 errors
- [ ] Log request and responses sent and received from Huaweicloud APIs by the SDK.

baixinsui:
Fixed #1848 
401 error in  plugin of  HuaweiCloud:
image

401 error in plugin of FlexibleEngine:
image

### DIFF
--- a/plugins/flexibleengine/src/main/java/org/eclipse/xpanse/plugins/flexibleengine/common/FlexibleEngineRetryStrategy.java
+++ b/plugins/flexibleengine/src/main/java/org/eclipse/xpanse/plugins/flexibleengine/common/FlexibleEngineRetryStrategy.java
@@ -7,12 +7,16 @@
 package org.eclipse.xpanse.plugins.flexibleengine.common;
 
 import com.huaweicloud.sdk.core.SdkResponse;
+import com.huaweicloud.sdk.core.exception.ClientRequestException;
 import com.huaweicloud.sdk.core.exception.ServiceResponseException;
 import com.huaweicloud.sdk.core.retry.RetryContext;
 import com.huaweicloud.sdk.core.retry.backoff.BackoffStrategy;
 import java.util.Objects;
 import lombok.extern.slf4j.Slf4j;
+import org.eclipse.xpanse.modules.models.common.exceptions.ClientAuthenticationFailedException;
 import org.springframework.beans.factory.annotation.Value;
+import org.springframework.http.HttpStatus;
+import org.springframework.retry.support.RetrySynchronizationManager;
 import org.springframework.stereotype.Component;
 
 /**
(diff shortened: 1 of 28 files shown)


plugins/flexibleengine/src/main/java/org/eclipse/xpanse/plugins/flexibleengine/common/FlexibleEngineRetryStrategy.java
```diff
@@ -93,4 +97,39 @@ public boolean matchRetryCondition(SdkResponse response, Exception ex) {
...
+    /**
+     * Handle auth exception for spring retry.
+     *
+     * @param ex Exception
+     */
+    public void handleAuthExceptionForSpringRetry(Exception ex) {
+        int retryCount = Objects.isNull(RetrySynchronizationManager.getContext())
+                ? 0 : RetrySynchronizationManager.getContext().getRetryCount();
+        log.error(ex.getMessage() + System.lineSeparator() + "Retry count:" + retryCount);
+        if (ex instanceof ClientAuthenticationFailedException) {
```
swaroopar, line 110: do we really need this check? this exception is only thrown by the code below it? 

if this is not needed, we can remove from other plugins also

swaroopar (changes requested):

plugins/flexibleengine/src/main/java/org/eclipse/xpanse/plugins/flexibleengine/common/FlexibleEngineRetryStrategy.java
baixinsui: The method getPrice called the method getProjectId which will throw ClientAuthenticationFailedException , it is needed
image

modules/api/src/main/java/org/eclipse/xpanse/api/controllers/ServiceDeployerApi.java
```diff
@@ -304,7 +304,7 @@ public ResponseEntity<List<String>> getAvailabilityZones(
                     this.deployService.getAvailabilityZonesOfRegion(csp, regionName, serviceUuid);
             return ResponseEntity.ok().cacheControl(getCacheControl()).body(availabilityZones);
         } catch (Exception ex) {
-            log.error("Error fetching availability zones", ex);
+            log.error("Error fetching availability zones. {}", ex.getMessage());
```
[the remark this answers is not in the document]
baixinsui: Restored

PR #1875, opened 2024-08-13
handle BSS APIs not available in all regions

### LINKED ISSUE #1850: handle BSS APIs not available in all regions
Currently the BSS APIs are not available in all regions. For example on eu-west-101 region, there is no API available. 
So if the the user is using his eu-west-101 credentials, then call to bss and bss-intl will fail since the eu-west-101 IAM is different to other regions.  New proposed solution to be tried is

- [ ] Try bss-intl -> if fails, try bss
- [ ] if even that fails, then try the public API which is also used by the page - https://www.huaweicloud.com/intl/en-us/pricing/calculator.html#/ecs
   this uses the API and does not need any credentials.  -> https://portal-intl.huaweicloud.com/api/cbc/global/rest/BSS/billing/ratingservice/v2/inquiry/resource
- [ ] Remove the extra check for text "CBC". https://github.com/eclipse-xpanse/xpanse/blob/main/plugins/huaweicloud/src/main/java/org/eclipse/xpanse/plugins/huaweicloud/price/HuaweiCloudPriceCalculator.java#L110
- [ ] Currently when we get errors, we ignore it and directly set only the license price and markup price as recurring price. This must be avoided - https://github.com/eclipse-xpanse/xpanse/blob/main/plugins/huaweicloud/src/main/java/org/eclipse/xpanse/plugins/huaweicloud/price/HuaweiCloudPriceCalculator.java#L110
sample request 
```json
{
    "regionId": "eu-west-101",
    "chargingMode": 1,
    "periodType": 4,
    "periodNum": 1,
    "subscriptionNum": 1,
    "productInfos": [
        {
            "id": "1722586728180-0-OFFI796331475118358528",
            "cloudServiceType": "hws.service.type.ec2",
            "resourceType": "hws.resource.type.vm",
            "resourceSpecCode": "c6.2xlarge.2.linux",
            "productNum": 1,
            "usageFactor": "Duration",
            "usageMeasureId": 4,
            "usageValue": 1
        },
        {
            "id": "1722586728180-1-OFFI794174495392378882",
            "cloudServiceType": "hws.service.type.ebs",
            "resourceType": "hws.resource.type.volume",
            "resourceSpecCode": "GPSSD",
            "productNum": 1,
            "resourceSize": 40,
            "resouceSizeMeasureId": 17,
            "usageFactor": "Duration",
            "usageMeasureId": 4,
            "usageValue": 1
        },
        {
            "id": "1722586728180-2-OFFI794477835484422145",
            "cloudServiceType": "hws.service.type.vpc",
            "resourceType": "hws.resource.type.bandwidth",
            "resourceSpecCode": "19_bgp",
            "productNum": 1,
            "resourceSize": 5,
            "resouceSizeMeasureId": 15,
            "usageFactor": "Duration",
            "usageMeasureId": 4,
            "usageValue": 1
        }
    ]
}
```

baixinsui:
Fixed #1850 

Get prices with `BssintlClient` for International website:
idea64_S4wQd6ByCt

Get prices with `BssClient` for Chinese website:
idea64_OXsFpzoKSP

Get prices with global rest API:
image

### DIFF
--- a/plugins/huaweicloud/src/main/java/org/eclipse/xpanse/plugins/huaweicloud/price/HuaweiCloudGlobalPriceCalculator.java
+++ b/plugins/huaweicloud/src/main/java/org/eclipse/xpanse/plugins/huaweicloud/price/HuaweiCloudGlobalPriceCalculator.java
@@ -1,0 +1,159 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * SPDX-FileCopyrightText: Huawei Inc.
+ *
+ */
+
+package org.eclipse.xpanse.plugins.huaweicloud.price;
+
+import com.huaweicloud.sdk.bssintl.v2.model.ListOnDemandResourceRatingsResponse;
+import java.math.BigDecimal;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.UUID;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang3.StringUtils;
+import org.eclipse.xpanse.modules.models.billing.Price;
+import org.eclipse.xpanse.modules.models.billing.Resource;
+import org.eclipse.xpanse.modules.models.billing.enums.Currency;
+import org.eclipse.xpanse.modules.models.billing.enums.PricingPeriod;
+import org.eclipse.xpanse.modules.models.common.exceptions.ClientApiCallFailedException;
+import org.eclipse.xpanse.modules.orchestrator.price.ServiceFlavorPriceRequest;
+import org.eclipse.xpanse.plugins.huaweicloud.price.model.ProductInfo;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.http.HttpEntity;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpMethod;
+import org.springframework.http.HttpStatusCode;
+import org.springframework.http.ResponseEntity;
+import org.springframework.stereotype.Component;
+import org.springframework.web.client.HttpClientErrorException;
+import org.springframework.web.client.HttpServerErrorException;
+import org.springframework.web.client.RestTemplate;
+
+/**
+ * Class that implements the price calculation for the global rest apis of HuaweiCloud.
+ */
+@Slf4j
+@Component
+public class HuaweiCloudGlobalPriceCalculator {
+
+    private final RestTemplate restTemplate = new RestTemplate();
+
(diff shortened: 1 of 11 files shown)


plugins/huaweicloud/src/main/java/org/eclipse/xpanse/plugins/huaweicloud/price/HuaweiCloudGlobalPriceCalculator.java
```diff
@@ -0,0 +1,155 @@
...
+
+    /**
+     * Get the price of the service with global rest api.
+     *
+     * @param request ServiceFlavorPriceRequest
+     * @return Price in the international website.
+     */
+    public Price getPriceWithResourcesUsageByGlobalRestApi(ServiceFlavorPriceRequest request) {
+        try {
+            String globalRestApiUrl = "https://portal-intl.huaweicloud.com"
```
swaroopar, line 52: would be good to keep this in application properties file.

swaroopar (changes requested):

plugins/huaweicloud/src/main/java/org/eclipse/xpanse/plugins/huaweicloud/price/HuaweiCloudGlobalPriceCalculator.java
baixinsui: Updated.

plugins/huaweicloud/src/main/java/org/eclipse/xpanse/plugins/huaweicloud/price/HuaweiCloudGlobalPriceCalculator.java
```diff
@@ -0,0 +1,159 @@
...
+/**
+ * Class that implements the price calculation for the global rest apis of HuaweiCloud.
+ */
+@Slf4j
+@Component
+public class HuaweiCloudGlobalPriceCalculator {
+
+    private final RestTemplate restTemplate = new RestTemplate();
+
+    @Value("${huaweicloud.price.calculator.global.restapi.url:https://portal-intl.huaweicloud.com"
```
swaroopar, line 45: @baixinsui this is missing in the application.properties file?

baixinsui: Submitted

swaroopar (approved):
LGTM

PR #1880, opened 2024-08-15
make all fixed prices region specific

baixinsui:
**Fixed https://github.com/eclipse-xpanse/xpanse/issues/1851**

**Service template with one specific price for region 'eu-west-101' and  one common price for 'any' region:**
image
**When select region 'eu-west-101', fixed prices of service as below:**
chrome_9VDNJDqXUt
**When select other region except 'eu-west-101', fixed prices of service as below:**
     Fixed prices for region 'cn-southwest-2':
chrome_JU2kE0JEh9
     Fixed prices for region 'eu-west-0':
image


swaroopar (approved):
LGTM

PR #1883, opened 2024-08-16
validate regions of csp in service template

baixinsui:
Fixed https://github.com/eclipse-xpanse/xpanse/issues/1847

Unavailable region in HuaweiCloud:
chrome_WTTLik8adL
Unavailable region in FlexibleEngine:
chrome_wW48EWOI2Z


swaroopar (approved):
LGTM

PR #1896, opened 2024-08-23
ServiceConfigurationParameter - Extend the parameter with resource information

### LINKED ISSUE #1788: ServiceConfigurationParameter - Extend the parameter with resource information
For each ServiceConfigurationParameter, it is necessary to add the resource information to which the configuration parameter belongs to.  Using this, we can decide on which VM we must run the configuration change 

- [ ] add a new string field called `managedBy` to `ServiceConfigurationParameter` class
- [ ] Add validation to check if the resource name provided in the `managedBy` field is correct. This must be validated as part of the service registration and update. 
- [ ] Add a new method to `OrchestratorPlugin` interface - GetResourcesInServiceDeployment.
- [ ] Return  map of resource name and resource type. 
 We can consider using [hcl4j](https://github.com/bertramdev/hcl4j) to parse the HCL file. 
- [ ] Use this method to validate the service configuration parameters.

for the example below, resource name is "kafka-broker" and type is "huaweicloud_compute_instance" and resource name is 
```hcl
resource "huaweicloud_compute_instance" "kafka-broker" {}
```

WangLiNaruto:
Fixed #1788 

image
image

image
image

### DIFF
--- a/modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/utils/ScriptsGitRepoManage.java
+++ b/modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/utils/ScriptsGitRepoManage.java
@@ -39,7 +39,7 @@ public class ScriptsGitRepoManage {
     @Retryable(retryFor = GitRepoCloneException.class,
             maxAttemptsExpression = "${http.request.retry.max.attempts}",
             backoff = @Backoff(delayExpression = "${http.request.retry.delay.milliseconds}"))
-    public void checkoutScripts(String workspace, ScriptsRepo scriptsRepo) {
+    public File[] checkoutScripts(String workspace, ScriptsRepo scriptsRepo) {
         File workspaceDirectory = new File(workspace);
         FileRepositoryBuilder repositoryBuilder = new FileRepositoryBuilder();
         repositoryBuilder.findGitDir(workspaceDirectory);
@@ -64,10 +64,10 @@ public class ScriptsGitRepoManage {
         } else {
             log.info("Scripts repo is already cloned in the workspace.");
         }
-        folderContainsScripts(workspace, scriptsRepo);
+        return folderContainsScripts(workspace, scriptsRepo);
     }
 
-    private void folderContainsScripts(String workspace, ScriptsRepo scriptsRepo) {
+    private File[] folderContainsScripts(String workspace, ScriptsRepo scriptsRepo) {
         File directory = new File(workspace
                 + (Objects.nonNull(scriptsRepo.getScriptsPath())
                 ? File.separator + scriptsRepo.getScriptsPath()
@@ -82,5 +82,6 @@ public class ScriptsGitRepoManage {
                             ? scriptsRepo.getScriptsPath() : "root")
                             + "' folder.");
         }
+        return files;
     }
 }
--- a/modules/orchestrator/src/main/java/org/eclipse/xpanse/modules/orchestrator/OrchestratorPlugin.java
+++ b/modules/orchestrator/src/main/java/org/eclipse/xpanse/modules/orchestrator/OrchestratorPlugin.java
@@ -6,7 +6,9 @@
 
 package org.eclipse.xpanse.modules.orchestrator;
 
+import java.io.File;
 import java.util.List;
+import java.util.Map;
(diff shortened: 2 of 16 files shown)


modules/servicetemplate/src/main/java/org/eclipse/xpanse/modules/servicetemplate/ServiceTemplateManage.java
```diff
@@ -225,6 +241,57 @@ public ServiceTemplateEntity registerServiceTemplate(Ocl ocl) {
...
+                log.error(errorMsg);
+                errors.add(errorMsg);
+            }
+        });
+        if (!CollectionUtils.isEmpty(errors)) {
+            throw new ServiceConfigurationInvalidException(errors.toString());
+        }
+    }
+
+    private File getScriptFile(String script) {
```
swaroopar, line 272: should we additionally handle if the scripts are hosted on a GIT repo?

modules/servicetemplate/src/main/java/org/eclipse/xpanse/modules/servicetemplate/ServiceTemplateManage.java
```diff
@@ -225,6 +241,57 @@ public ServiceTemplateEntity registerServiceTemplate(Ocl ocl) {
         return storedServiceTemplate;
     }
 
+    private void validateServiceConfigurationParameters(Ocl ocl) {
```
swaroopar, line 244: its better we move this complete logic to a separate component.

modules/orchestrator/src/main/java/org/eclipse/xpanse/modules/orchestrator/OrchestratorPlugin.java
```diff
@@ -54,4 +56,6 @@ public interface OrchestratorPlugin extends ServiceResourceHandler, Authenticati
      * @return true if all regions are valid.
      */
     boolean validateRegionsOfService(Ocl ocl);
+
+    Map<String, String> getResourcesInServiceDeployment(File scriptFile);
```
swaroopar, line 60: since we are returning only the compute resources, it is better we rename this to `getComputeResourcesInServiceDeployment`

modules/servicetemplate/src/main/java/org/eclipse/xpanse/modules/servicetemplate/utils/ServiceConfigurationParameterValidator.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    private File prepareDeployWorkspaceWithScripts(Ocl ocl) {
+        String workspace = getWorkspacePath(UUID.randomUUID());
+        if (Objects.nonNull(ocl.getDeployment().getScriptsRepo())) {
+            File[] files = scriptsGitRepoManage
+                    .checkoutScripts(workspace, ocl.getDeployment().getScriptsRepo());
+            return Arrays.stream(files)
+                    .filter(file -> file.getName().endsWith(".tf"))
+                    .filter(file -> !"provider.tf".equals(file.getName()))
```
swaroopar, line 85: move the provider.tf value also to static variable.

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/utils/ScriptsGitRepoManage.java
```diff
@@ -39,7 +39,7 @@ public class ScriptsGitRepoManage {
     @Retryable(retryFor = GitRepoCloneException.class,
             maxAttemptsExpression = "${http.request.retry.max.attempts}",
             backoff = @Backoff(delayExpression = "${http.request.retry.delay.milliseconds}"))
-    public void checkoutScripts(String workspace, ScriptsRepo scriptsRepo) {
+    public File[] checkoutScripts(String workspace, ScriptsRepo scriptsRepo) {
```
swaroopar, line 42: this seems to be duplicate work. 
The file is already available in the GIT checkout. We are loading it into memory and writing new files again. 
We must try to directly run the HCLParser from the GIT repo location.

swaroopar (changes requested):

swaroopar (approved):
LGTM

PR #1909, opened 2024-08-29
Adapt changeServiceConfiguration API

### LINKED ISSUE #1867: Adapt changeServiceConfiguration API
- [x] Split the config request into multiple ones based on the config manager which is responsible for the configuration paramters. 
- [x] All config requests must have same order ID. 
- [x] If the service component has multiple nodes and if runOnEach is true, then add multiple entries for each resource. 
- [x] Use the getComputeResourceInventoryOfService to get the names of the resources. 
- [x] Store all requests to DB. 

Depends on https://github.com/eclipse-xpanse/xpanse/issues/1868

### LINKED ISSUE #1868: Change Table SERVICE_CONFIGURATION_UPDATE_REQUEST
- [x] add SERVICE_ID - make it a foreign key of DEPLOY_SERVICE table. 
- [x] ORDER_ID should be foreign key of order table. 
- [x] Add RESOURCE_NAME column. 
- [x] Add CONFIG_MANAGER column. 
- [x] RESULT_MESSAGE column

WangLiNaruto:
Fixed #1867 
Fixed #1868 

image
image


Alice1319:
LGTM

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/ServiceConfigurationManager.java
```diff
@@ -85,42 +112,118 @@ public ServiceOrder changeServiceConfiguration(String serviceId,
...
+        });
+        updateRequestMap.forEach((k, v) -> {
+            if (existsServiceConfig.containsKey(k)) {
+                existsServiceConfig.put(k, v);
+            }
+        });
+        return existsServiceConfig;
+    }
+
+    private ServiceConfigurationUpdateRequest getInitServiceConfigurationUpdateRequest(UUID orderId,
```
swaroopar, line 185: we can remove the 'init' word from this method name as it is not the initial request only.

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/ServiceConfigurationManager.java
```diff
@@ -85,42 +112,118 @@ public ServiceOrder changeServiceConfiguration(String serviceId,
         }
     }
 
-    private ServiceConfigurationUpdateRequest addServiceConfigurationUpdateRequest(
-            Map<String, Object> existingConfiguration, Map<String, Object> updateRequestMap) {
+    private void addServiceConfigurationUpdateRequests(UUID orderId, String serviceId,
```
swaroopar, line 115: we need atleast one unittest for this. Can you you please add one?

swaroopar (changes requested):

runtime/src/test/java/org/eclipse/xpanse/runtime/ServiceConfigurationApiTest.java
```diff
@@ -0,0 +1,82 @@
...
+
+    @BeforeEach
+    void setUp() {
+        MockitoAnnotations.openMocks(this);
+        this.mockMvc = MockMvcBuilders.standaloneSetup(serviceConfigurationApi).build();
+    }
+
+    @Test
+    @WithMockUser(roles = {"USER"})
+    void testChangeServiceConfiguration_Success() throws Exception {
```
swaroopar, line 66: would it be possible to add a full end to end test for this?
1. Register a kafka template. 
2. Request config change. 
3. Validate entries in config table - check that exact number of expected requests are created. 
4. validate the request properties are as expected. 

Then we know that if anything changes in the logic, this test will fail.

swaroopar (changes requested):

modules/database/src/main/java/org/eclipse/xpanse/modules/database/serviceconfiguration/update/DatabaseServiceConfigurationUpdateStorage.java
```diff
@@ -31,4 +34,11 @@ public ServiceConfigurationUpdateRequest storeAndFlush(
             ServiceConfigurationUpdateRequest serviceConfigurationUpdateRequest) {
         return serviceConfigurationUpdateRepository.saveAndFlush(serviceConfigurationUpdateRequest);
     }
+
+    @Override
+    public void batchInsert(List<ServiceConfigurationUpdateRequest> requests) {
+        if (!CollectionUtils.isEmpty(requests)) {
+            requests.forEach(serviceConfigurationUpdateRepository::saveAndFlush);
```
baixinsui, line 41: We can add  a new method `<S extends T> List<S> saveAll(Iterable<S> entities);`  in the `CustomJpaRepository`, and then use the new method `saveAll` to improve efficiency by reducing database connections

modules/database/src/main/java/org/eclipse/xpanse/modules/database/serviceconfiguration/update/ServiceConfigurationUpdateRequest.java
```diff
@@ -35,8 +41,23 @@ public class ServiceConfigurationUpdateRequest implements Serializable {
...
+    @Column(name = "ID", nullable = false)
+    private UUID id;
+
+    @ManyToOne
+    @JoinColumn(name = "ORDER_ID", nullable = false)
+    private ServiceOrderEntity serviceOrderEntity;
+
+    @ManyToOne
+    @JoinColumn(name = "SERVICE_ID", nullable = false)
+    private DeployServiceEntity deployServiceEntity;
```
baixinsui, line 54: Will these records be deleted by cascading when associated services are destroyed and purged from the database

modules/database/src/main/java/org/eclipse/xpanse/modules/database/serviceconfiguration/update/DatabaseServiceConfigurationUpdateStorage.java
swaroopar: agreed. I Missed this.

modules/database/src/main/java/org/eclipse/xpanse/modules/database/serviceconfiguration/update/ServiceConfigurationUpdateRequest.java
swaroopar: yes, please check that all child tables are also deleted automatically when the service is purged.

swaroopar (approved):
LGTM

PR #1918, opened 2024-09-05
Fix error when create service policies with empty flavors list 

baixinsui:
Fixed https://github.com/eclipse-xpanse/xpanse/issues/1919


swaroopar (approved):
LGTM

Issue #1900, opened 2024-08-26
add billingmode and region column to myserivces and reports


swaroopar:
- [x] Add billingmode and region column to myservices and reports page. 
- [x] Use antd component similar to serviceHostedBy Column for billing mode. 
- [x] Region can be plain text.

Alice1319:
I‘ll take this.

PR #1928, opened 2024-09-09
add region and billingmode for deployedService.

Alice1319:
releated #1900


modules/models/src/main/java/org/eclipse/xpanse/modules/models/service/view/DeployedService.java
```diff
@@ -40,6 +42,20 @@ public class DeployedService {
     @Schema(description = "The catalog of the service")
     private Category category;
 
+    /**
+     * The region of the Service.
+     */
+    @NotNull
+    @Schema(description = "The region of the provider.")
```
baixinsui, line 49: It better to update the description `The region of the service.`

Alice1319: fixed

swaroopar (approved):
LGTM

Issue #1936, opened 2024-09-11
End User Scale Service error


baixinsui:
When the user has no role(default granted user)  select options scale or modify parameters of a deployed service, the UI client will call the API `/xpanse/service_templates/{serviceTemplateId}` which need users have role `ISV` and return 403 error, this cause the page error.
Modify parameters of the service error:
![chrome_Qxvm9Me4cj](https://github.com/user-attachments/assets/e36f36b2-8518-4650-847d-9e040b09365d)
Scale the service error:
![image](https://github.com/user-attachments/assets/0108fb48-a615-4c8b-83da-c3b3d7efeca1)

This error must be fixed. There are two solutions as below:
1.  The UI Client change to call the API `/xpanse/catalog/services/{serviceTemplateId}` which is for the end users in all pages for the end users.
2.  In the back-end, add role `User` into the needed roles of the API `/xpanse/service_templates/{serviceTemplateId}`.

swaroopar:
@baixinsui we must use the API which is specifically for end user. This must be fixed on the UI.

baixinsui:
@swaroopar I tried to replace the API in the UI client but there are other errors. I could not find out the caused. Can you fix it in the UI Client or assign it to other partners.

swaroopar:
@Alice1319 could you please check this bug? We are using a wrong API to fetch service template details.

Alice1319:
Sure,I'll check this

PR #1948, opened 2024-09-14
update flavor of orderable service details api for end user.

Alice1319:
releated #1936

### DIFF
--- a/modules/api/src/main/java/org/eclipse/xpanse/api/config/ServiceTemplateEntityConverter.java
+++ b/modules/api/src/main/java/org/eclipse/xpanse/api/config/ServiceTemplateEntityConverter.java
@@ -10,6 +10,7 @@ import java.util.Objects;
 import org.apache.commons.lang3.StringUtils;
 import org.eclipse.xpanse.api.controllers.ServiceCatalogApi;
 import org.eclipse.xpanse.modules.database.servicetemplate.ServiceTemplateEntity;
+import org.eclipse.xpanse.modules.models.servicetemplate.EndUserFlavors;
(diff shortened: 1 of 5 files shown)


modules/models/src/main/java/org/eclipse/xpanse/modules/models/servicetemplate/UserOrderableServiceFlavor.java
```diff
@@ -0,0 +1,45 @@
...
+import lombok.EqualsAndHashCode;
+import lombok.ToString;
+
+/**
+ * Defines user orderable service flavor.
+ */
+@Data
+@ToString(callSuper = true)
+@EqualsAndHashCode(callSuper = true)
+public class UserOrderableServiceFlavor extends ServiceFlavor {
```
swaroopar, line 22: we can rename this to ServiceFlavorEndUserInfo

swaroopar (changes requested):

modules/models/src/main/java/org/eclipse/xpanse/modules/models/servicetemplate/UserOrderableServiceFlavor.java
swaroopar, line 22: can you also please add a comment here to this class for future purpose?
`we need a separate data model to end user since the original flavor data model is used by ISV and CSP also have the pricing configuration info which should not be returned to the end user`

swaroopar (changes requested):

baixinsui (commented):
LGTM

swaroopar (approved):
LGTM

PR #1952, opened 2024-09-18
extend OCL with deployer tool version

baixinsui:
**Fixed https://github.com/eclipse-xpanse/xpanse/issues/1940**

**Register service successfully:**
image

**Register services failed by invalid version of `deployerTool`:**
version of `deployerTool` is null:
image
invalid version without comparing symbols:
image
version with invalid comparing symbols:
image
version with invalid version number:
image


swaroopar (approved):
LGTM

PR #1953, opened 2024-09-19
add distributed locking mechanism for requests from agents

### LINKED ISSUE #1913: add distributed locking mechanism for requests from agents
- [ ] if `runOnlyOnce` is set to `true` then the `resourceName` column must be set to blank.  This will be flexible that any resource in the deployment can fetch the pending request. 
- [ ] When getPendingConfigurationChangeRequest is received, return the pending request with either resource_name specific for the resource_name in the request URI or with the resource_name  is empty. If the resource_name is empty, check if the resource_name in the request belongs to the same `config_manager` group in the service inventory. 
- [ ] To enable distributed locking - the read from DB must be executed using JPA `LockModeType.PESSIMISTIC_WRITE`. And the update of the status to IN-PROGRESS. Both must be wrapped under one `@transactional`
- [ ] Set `javax.persistence.lock.timeout` to 2seconds for this specific transaction and return when LockTimeoutException` occurs then return HTTP 204. 
- [ ] Test that when multiple requests are received, only one gets processed and other one waits until the distributed lock is released. 
- [ ] In the pollingApi, if the config update order with resource_name empty is returned to an agent, then update that agent's resource_name in the entry. So that we then know that which resource has actually got the request and is processing it. 

depends on #1869

WangLiNaruto:
Fixed #1913

### DIFF
--- a/modules/database/src/main/java/org/eclipse/xpanse/modules/database/serviceconfiguration/update/ServiceConfigurationUpdateRepository.java
+++ b/modules/database/src/main/java/org/eclipse/xpanse/modules/database/serviceconfiguration/update/ServiceConfigurationUpdateRepository.java
@@ -6,9 +6,16 @@
 
 package org.eclipse.xpanse.modules.database.serviceconfiguration.update;
 
+import jakarta.persistence.LockModeType;
+import jakarta.persistence.QueryHint;
+import java.util.List;
 import java.util.UUID;
 import org.eclipse.xpanse.modules.database.CustomJpaRepository;
+import org.springframework.data.jpa.domain.Specification;
 import org.springframework.data.jpa.repository.JpaSpecificationExecutor;
+import org.springframework.data.jpa.repository.Lock;
+import org.springframework.data.jpa.repository.QueryHints;
+import org.springframework.lang.Nullable;
 
 /**
  * Interface to access default JPA methods.
@@ -17,4 +24,10 @@ public interface ServiceConfigurationUpdateRepository extends
         CustomJpaRepository<ServiceConfigurationUpdateRequest, UUID>,
         JpaSpecificationExecutor<ServiceConfigurationUpdateRequest> {
 
+    @Lock(LockModeType.PESSIMISTIC_WRITE)
+    // it is not possible to read this config from spring properties.
+    @QueryHints({@QueryHint(name = "javax.persistence.lock.timeout", value = "2000")})
+    @Override
+    List<ServiceConfigurationUpdateRequest> findAll(
+            @Nullable Specification<ServiceConfigurationUpdateRequest> spec);
 }
--- a/modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/ServiceConfigurationManager.java
+++ b/modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/ServiceConfigurationManager.java
@@ -7,6 +7,7 @@
 package org.eclipse.xpanse.modules.deployment;
 
 import jakarta.annotation.Resource;
+import jakarta.transaction.Transactional;
 import java.time.OffsetDateTime;
 import java.util.ArrayList;
 import java.util.Comparator;
@@ -49,6 +50,7 @@ import org.eclipse.xpanse.modules.models.servicetemplate.ServiceConfigurationPar
 import org.eclipse.xpanse.modules.models.servicetemplate.exceptions.ServiceTemplateNotRegistered;
 import org.eclipse.xpanse.modules.models.servicetemplate.utils.JsonObjectSchema;
 import org.eclipse.xpanse.modules.security.UserServiceHelper;
+import org.hibernate.exception.LockTimeoutException;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.stereotype.Component;
@@ -167,8 +169,6 @@ public class ServiceConfigurationManager {
                             ServiceConfigurationUpdateRequest request =
                                     getServiceConfigurationUpdateRequest(orderId, groupName,
                                             deployServiceEntity, properties);
-                            request.setResourceName(
-                                    deployResourceList.getFirst().getResourceName());
                             requests.add(request);
                         } else {
                             deployResourceList.forEach(deployResource -> {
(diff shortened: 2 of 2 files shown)


modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/ServiceConfigurationUpdateHandler.java
```diff
@@ -0,0 +1,65 @@
...
+    @Transactional
+    public ServiceConfigurationUpdateRequest
+            updateServiceConfigurationUpdateRequest(UUID id, String resourceName) {
+        try {
+            Map<String, Object> properties = new HashMap<>();
+            properties.put(JAVAX_PERSISTENCE_LOCK_TIMEOUT, lockTimeout);
+            ServiceConfigurationUpdateRequest request = entityManager
+                    .find(ServiceConfigurationUpdateRequest.class, id,
+                            LockModeType.PESSIMISTIC_WRITE, properties);
+            if (Objects.isNull(request.getResourceName())) {
```
baixinsui, line 55: When the request is null, the code `request.getResourceName()` will throw `NullPointerException`.  Please check it and output the null log.

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/ServiceConfigurationUpdateHandler.java
```diff
@@ -0,0 +1,65 @@
...
+                request.setResourceName(resourceName);
+            }
+            request.setStatus(ServiceConfigurationStatus.PROCESSING);
+            return serviceConfigurationUpdateStorage.storeAndFlush(request);
+        } catch (LockTimeoutException e) {
+            log.error("Update service configuration update request object timed out", e);
+            return null;
```
baixinsui, line 62: If the method is important to the service, please throw the `LockTimeoutException` and use  `@Retryable(value = LockTimeoutException.class, maxAttempts = 3)` on this method.

WangLiNaruto: @baixinsui 
image
There is no need for a retry mechanism here. For parallel requests, if the lock times out when obtaining data, 204 needs to be returned.

runtime/src/main/resources/application.properties
```diff
@@ -68,4 +68,5 @@ multiple.providers.black.properties=OS_AUTH_URL
 huaweicloud.international.price.calculator.url=https://portal-intl.huaweicloud.com/api/cbc/global/rest/BSS/billing/ratingservice/v2/inquiry/resource
 huaweicloud.european.price.calculator.url=https://portal.eu.huaweicloud.com/api/cbc/global/rest/BSS/billing/ratingservice/v2/inquiry/resource
 huaweicloud.chinese.price.calculator.url=https://portal.huaweicloud.com/api/cbc/global/rest/BSS/billing/ratingservice/v2/inquiry/resource
-enable.agent.api.only=false
\ No newline at end of file
+enable.agent.api.only=false
+javax.persistence.lock.timeout=2000
```
swaroopar, line 72: we need a better name here which is specific just for the agent wait time out. 
pending.config.change.query.wait.time

swaroopar, line 72: we need a better name here which is specific just for the agent wait time out.
pending.config.change.query.wait.time

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/ServiceConfigurationUpdateHandler.java
```diff
@@ -0,0 +1,65 @@
...
+     */
+    @Transactional
+    public ServiceConfigurationUpdateRequest
+            updateServiceConfigurationUpdateRequest(UUID id, String resourceName) {
+        try {
+            Map<String, Object> properties = new HashMap<>();
+            properties.put(JAVAX_PERSISTENCE_LOCK_TIMEOUT, lockTimeout);
+            ServiceConfigurationUpdateRequest request = entityManager
+                    .find(ServiceConfigurationUpdateRequest.class, id,
+                            LockModeType.PESSIMISTIC_WRITE, properties);
```
swaroopar, line 54: this is a wrong place. The lock must be at the read query. The name is a bit misleading. Even though the lockType is PESSIMISTIC_WRITE, it actually blocks both read and write. 
So the first query that is executed must lock the rows for read and write both. 
So any other query which comes after that, must wait until the locks are released. 

You can see some examples here. 
https://beratyesbek.medium.com/pessimistic-and-optimistic-locking-in-jpa-spring-boot-aa59f2b906fa

swaroopar (changes requested):

modules/database/src/main/java/org/eclipse/xpanse/modules/database/serviceconfiguration/update/ServiceConfigurationUpdateRepository.java
```diff
@@ -6,15 +6,32 @@
...
  */
-public interface ServiceConfigurationUpdateRepository extends
-        CustomJpaRepository<ServiceConfigurationUpdateRequest, UUID>,
+@Repository
+public interface ServiceConfigurationUpdateRepository
+        extends CustomJpaRepository<ServiceConfigurationUpdateRequest, UUID>,
         JpaSpecificationExecutor<ServiceConfigurationUpdateRequest> {
 
+    /**
+     * it is possible to read this config from spring properties.
```
swaroopar, line 30: sorry my mistake. It must be 'It is not possible ..'
Also better to add this comment above line 33

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/ServiceConfigurationManager.java
```diff
@@ -167,15 +169,19 @@ private void addServiceConfigurationUpdateRequests(UUID orderId, String serviceI
                             ServiceConfigurationUpdateRequest request =
                                     getServiceConfigurationUpdateRequest(orderId, groupName,
                                             deployServiceEntity, properties);
-                            request.setResourceName(
-                                    deployResourceList.getFirst().getResourceName());
+                            if (configManageScript.getRunOnlyOnce()) {
+                                request.setResourceName(
```
swaroopar, line 173: this must be set to blank if runOnlyOnce is true. Can you please check again?

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/ServiceConfigurationManager.java
```diff
@@ -167,15 +169,19 @@ private void addServiceConfigurationUpdateRequests(UUID orderId, String serviceI
...
+                                        deployResourceList.getFirst().getResourceName());
+                            }
                             requests.add(request);
                         } else {
                             deployResourceList.forEach(deployResource -> {
                                 ServiceConfigurationUpdateRequest request =
                                         getServiceConfigurationUpdateRequest(orderId, groupName,
                                                 deployServiceEntity, properties);
-                                request.setResourceName(deployResource.getResourceName());
+                                if (configManageScript.getRunOnlyOnce()) {
```
swaroopar, line 182: confused why do we have same check in line 168, 172 and 182

swaroopar (changes requested):

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/ServiceConfigurationManager.java
```diff
@@ -257,40 +258,111 @@ private void validate(ServiceTemplateEntity serviceTemplateEntity,
...
-        request.setStatus(ServiceConfigurationStatus.PROCESSING);
-        serviceConfigurationUpdateStorage.storeAndFlush(request);
-        ServiceConfigurationChangeRequest serviceConfigurationChangeRequest
-                = getChangeRequestForAgent(request);
-        return ResponseEntity.status(HttpStatus.OK).body(serviceConfigurationChangeRequest);
+        if (!resourceNameMap.containsKey(resourceName)) {
+            String errorMsg = String.format("The service with serviceId %s does not contain "
+                    + "a resource with resource_name %s", serviceId, resourceName);
+            log.error(errorMsg);
+            throw new IllegalArgumentException(errorMsg);
```
swaroopar, line 344: would prefer to send any of our custom exception. If we do not have any thing that fits, we must create a new one.

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/ServiceConfigurationManager.java
```diff
@@ -257,40 +258,111 @@ private void validate(ServiceTemplateEntity serviceTemplateEntity,
...
         List<ServiceConfigurationUpdateRequest> requests =
                 serviceConfigurationUpdateStorage.listServiceConfigurationUpdateRequests(model);
         if (CollectionUtils.isEmpty(requests)) {
-            return ResponseEntity.status(HttpStatus.NO_CONTENT)
-                    .body(new ServiceConfigurationChangeRequest());
+            return null;
         }
-        Optional<ServiceConfigurationUpdateRequest> latestRequest = requests.stream()
+        Optional<ServiceConfigurationUpdateRequest> oldestRequestOptional = requests.stream()
+                .filter(request -> Objects.isNull(request.getResourceName())
```
swaroopar, line 316: here we check if its null, but we must also check if the configManager is same  as the group to which the resource_name in the request belongs to.

swaroopar (changes requested):

swaroopar (approved):
LGTM

PR #1974, opened 2024-09-26
change ServiceConfigurationChangeRequest to use changeId instead of orderId

### LINKED ISSUE #1963: change ServiceConfigurationChangeRequest  to use changeId instead of orderId
Since one orderId can have multiple changeIds, it is better to use changeId in the request so that it can be uniquely correlated. 

- [x] ServiceConfigurationChangeRequest -> change orderId to changeId in the data model and pass the ID of the request row.

### LINKED ISSUE #1968: getPendingConfigurationChangeRequest returns HTTP 204 with wrong response body
When getPendingConfigurationChangeRequest returns 204, it must not return any response body. 

20240925 12:22:35.713 [http-nio-8080-exec-6] INFO  o.e.x.m.logging.CustomHttpLogWriter [e09c3a0b-112d-4d06-ba41-58caf16a0c7c] [] []- Response: 204 No Content Duration: 3ms  {"orderId":null,"configParameters":null,"ansibleScriptConfig":null,"ansibleInventory":null}

WangLiNaruto:
Fixed #1963
Fixed #1968

image
image

image
image


swaroopar (approved):
LGTM

PR #1979, opened 2024-09-26
create API method to get current configuration of the service

### LINKED ISSUE #1967: create API method to get current configuration of the service
- [ ] Create method - getCurrentConfigurationOfAService
- [ ] URI - GET /xpanse/service/current/config/{serviceId}

WangLiNaruto:
Fixed #1967
image

### DIFF
--- a/modules/api/src/main/java/org/eclipse/xpanse/api/controllers/ServiceConfigurationApi.java
+++ b/modules/api/src/main/java/org/eclipse/xpanse/api/controllers/ServiceConfigurationApi.java
@@ -20,6 +20,7 @@ import org.eclipse.xpanse.api.config.AuditApiRequest;
 import org.eclipse.xpanse.modules.database.serviceconfiguration.update.ServiceConfigurationUpdateRequest;
 import org.eclipse.xpanse.modules.deployment.ServiceConfigurationManager;
 import org.eclipse.xpanse.modules.models.service.order.ServiceOrder;
+import org.eclipse.xpanse.modules.models.serviceconfiguration.ServiceConfigurationDetails;
 import org.eclipse.xpanse.modules.models.serviceconfiguration.ServiceConfigurationUpdate;
 import org.eclipse.xpanse.modules.models.serviceconfiguration.enums.ServiceConfigurationStatus;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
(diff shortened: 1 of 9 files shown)


modules/api/src/main/java/org/eclipse/xpanse/api/controllers/ServiceConfigurationApi.java
```diff
@@ -51,6 +52,20 @@ public class ServiceConfigurationApi {
     @Resource
     private ServiceConfigurationManager serviceConfigurationManager;
 
+    @Tag(name = "ServiceConfiguration",
+            description = "APIs for managing service's configuration.")
```
Alice1319, line 56: LGTM

Alice1319:
LGTM

modules/api/src/main/java/org/eclipse/xpanse/api/controllers/ServiceConfigurationApi.java
```diff
@@ -51,6 +52,20 @@ public class ServiceConfigurationApi {
...
+    @GetMapping(value = "/service/current/config/{serviceId}",
+            produces = MediaType.APPLICATION_JSON_VALUE)
+    @ResponseStatus(HttpStatus.OK)
+    @Operation(description = "Query the service's configuration by serviceId.")
```
baixinsui, line 60: Itis better to update the description with 'Query the service's  current configuration by id of the deployed service'

modules/api/src/main/java/org/eclipse/xpanse/api/controllers/ServiceConfigurationApi.java
```diff
@@ -51,6 +52,33 @@ public class ServiceConfigurationApi {
...
+     */
+    @Tag(name = "ServiceConfiguration",
+            description = "APIs for managing service's configuration.")
+    @GetMapping(value = "/service/current/config/{serviceId}",
+            produces = MediaType.APPLICATION_JSON_VALUE)
+    @ResponseStatus(HttpStatus.OK)
+    @Operation(description = "Query the service's current configuration by"
+            + " id of the deployed service.")
+    @AuditApiRequest(methodName = "getCspFromServiceId")
+    public ServiceConfigurationEntity getCurrentConfigurationOfService(
```
swaroopar, line 69: we must not return the entity as it is. If we do so, the deploy service entity is also returned.  This must be suppressed.

swaroopar (changes requested):

swaroopar (approved):
LGTM

PR #1980, opened 2024-09-26
extend terraform and opentofu local deployers to support multiple versions

baixinsui:
Fixed https://github.com/eclipse-xpanse/xpanse/issues/1941

OpenTofu with multiple versions:
image

Terraform with multiple versions:
image


runtime/src/test/java/org/eclipse/xpanse/runtime/modules/deployment/OpenTofuInstallerTest.java
```diff
@@ -0,0 +1,61 @@
+package org.eclipse.xpanse.runtime.modules.deployment;
+
+import static org.eclipse.xpanse.modules.deployment.deployers.opentofu.opentofulocal.OpenTofuInstaller.OPEN_TOFU_VERSION_OUTPUT_PATTERN;
```
Alice1319, line 3: this line of code can be moved below line 11.

runtime/src/test/java/org/eclipse/xpanse/runtime/modules/deployment/TerraformInstallerTest.java
```diff
@@ -0,0 +1,62 @@
+package org.eclipse.xpanse.runtime.modules.deployment;
+
+import static org.eclipse.xpanse.modules.deployment.deployers.terraform.terraformlocal.TerraformInstaller.TERRAFORM_VERSION_OUTPUT_PATTERN;
```
Alice1319, line 3: same to this line.

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/deployers/DeployerToolVersionsFetcher.java
```diff
@@ -0,0 +1,106 @@
...
+            try {
+                List<String> versionsInPage =
+                        fetchVersionsFromApi(fetchVersionsApiUrl, page, perPage);
+                if (CollectionUtils.isEmpty(versionsInPage)) {
+                    break;
+                } else {
+                    allVersions.addAll(versionsInPage);
+                    page++;
+                }
+            } catch (IOException e) {
```
WangLiNaruto, line 67: The exception should be handled and the loop terminated

Alice1319:
LGTM

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/deployers/DeployerToolVersionsFetcher.java
```diff
@@ -0,0 +1,94 @@
...
+            case OPEN_TOFU -> OPEN_TOFU_VERSIONS_GITHUB_URL;
+        };
+    }
+
+    private List<String> fetchVersionsFromWebsite(String releasesUrlWithPage) throws IOException {
+        List<String> versions = new ArrayList<>();
+        Document doc = Jsoup.connect(releasesUrlWithPage)
+                .ignoreHttpErrors(true).timeout(5000).get();
+        // version elements with css like <span class="ml-1 wb-break-all">v1.8.0</span>
+        Elements versionElements = doc.select("span.ml-1.wb-break-all");
```
swaroopar, line 83: @baixinsui this looks a bit hacky. We did not use such solution in terraform-boot know?

How about using 'https://api.github.com/repos/hashicorp/terraform/releases' instead? May be we can even use the github Java SDK - https://github-api.kohsuke.org/. 
The releases and tags API of Github works without authentication for public repositories. 
Just an idea.

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/deployers/DeployerToolVersionsFetcher.java
```diff
@@ -0,0 +1,94 @@
+/*
```
swaroopar, line 1: we can move these two classes to another package - deployertools

swaroopar (changes requested):

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/deployers/DeployerToolVersionsFetcher.java
baixinsui: I used request the API without the authentication. Due to the limit of 60 times per hour, some tests of the APIs failed.

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java
```diff
@@ -104,7 +104,7 @@ public class DeployService {
     private ServiceOrderManager serviceOrderManager;
     @Resource
     private ServiceDeploymentStatusChangePolling serviceDeploymentStatusChangePolling;
-    @Resource
+    @Resource(name = "xpanseAsyncTaskExecutor")
```
swaroopar, line 107: can we use the name from here - https://github.com/swaroopar/osc/blob/191b44b81a1113b5675a64260e2b046aac918c36/modules/async/src/main/java/org/eclipse/xpanse/modules/async/TaskConfiguration.java#L21

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/ServiceStateManager.java
```diff
@@ -57,8 +56,7 @@ public class ServiceStateManager {
     private DatabaseServiceStateManagementTaskStorage taskStorage;
     @Resource
     private DatabaseDeployServiceStorage deployServiceStorage;
-    @Qualifier("xpanseAsyncTaskExecutor")
-    @Resource
+    @Resource(name = "xpanseAsyncTaskExecutor")
```
swaroopar, line 59: can we use the name from here - https://github.com/swaroopar/osc/blob/191b44b81a1113b5675a64260e2b046aac918c36/modules/async/src/main/java/org/eclipse/xpanse/modules/async/TaskConfiguration.java#L21

swaroopar (changes requested):

swaroopar (approved):
LGTM

PR #1996, opened 2024-10-01
pass service id as an environment variable to deployers

swaroopar:
fixes https://github.com/eclipse-xpanse/xpanse/issues/1991


iskey (approved):
LGTM

PR #2021, opened 2024-10-10
add agent version and url as variables to deployer

swaroopar:
fixes https://github.com/eclipse-xpanse/xpanse/issues/2006


iskey (approved):
LGTM

PR #2045, opened 2024-10-17
return service configuration information for different roles

### LINKED ISSUE #2009: return service configuration information for different roles
- [ ] Return `ServiceConfigurationManage` object completely in ServiceTemplateDetailVo. This will be used by CSP and ISV roles. 
- [ ] Return only `configurationParameters` from `ServiceConfigurationManage` object  in UserOrderableServiceVo. This will be used by USER role.

WangLiNaruto:
Fixed #2009

image
image
image

### DIFF
--- a/modules/api/src/main/java/org/eclipse/xpanse/api/config/ServiceTemplateEntityConverter.java
+++ b/modules/api/src/main/java/org/eclipse/xpanse/api/config/ServiceTemplateEntityConverter.java
@@ -12,6 +12,7 @@ import org.eclipse.xpanse.api.controllers.ServiceCatalogApi;
 import org.eclipse.xpanse.modules.database.servicetemplate.ServiceTemplateEntity;
 import org.eclipse.xpanse.modules.models.servicetemplate.EndUserFlavors;
 import org.eclipse.xpanse.modules.models.servicetemplate.FlavorsWithPrice;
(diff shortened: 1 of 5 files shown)


modules/api/src/main/java/org/eclipse/xpanse/api/config/ServiceTemplateEntityConverter.java
```diff
@@ -100,6 +102,8 @@ public static UserOrderableServiceVo convertToUserOrderableServiceVo(
             userOrderableServiceVo.setServiceAvailabilityConfig(
                     serviceTemplateEntity.getOcl().getDeployment().getServiceAvailabilityConfig());
             userOrderableServiceVo.setEula(serviceTemplateEntity.getOcl().getEula());
+            userOrderableServiceVo.setConfigurationParameters(serviceTemplateEntity.getOcl()
+                    .getServiceConfigurationManage().getConfigurationParameters());
```
baixinsui, line 106: The filed `serviceConfigurationManage`  is not a NotNull field in `Ocl` object. Need to check null to prevent null pointer exception

WangLiNaruto: Fixed.

Alice1319:
LGTM

swaroopar (approved):
LGTM

PR #2060, opened 2024-10-23
use service_order table for lock/unlock requests

baixinsui:
Fixed https://github.com/eclipse-xpanse/xpanse/issues/2058

chrome_a80eqeQlgT

### DIFF
--- a/modules/database/src/main/java/org/eclipse/xpanse/modules/database/serviceorder/ServiceOrderEntity.java
+++ b/modules/database/src/main/java/org/eclipse/xpanse/modules/database/serviceorder/ServiceOrderEntity.java
@@ -118,4 +118,9 @@ public class ServiceOrderEntity implements Serializable {
     @Convert(converter = ObjectJsonConverter.class)
     private Map<String, Object> newConfigRequest;
 
+    @Column(name = "REQUEST_BODY", columnDefinition = "json")
+    @Type(value = JsonType.class)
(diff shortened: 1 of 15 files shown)


WangLiNaruto:
LGTM

Alice1319:
LGTM

modules/database/src/main/java/org/eclipse/xpanse/modules/database/serviceorder/ServiceOrderEntity.java
```diff
@@ -118,4 +119,14 @@ public class ServiceOrderEntity implements Serializable {
     @Convert(converter = ObjectJsonConverter.class)
     private Map<String, Object> newConfigRequest;
 
+    @Column(name = "PREVIOUS_LOCK_CONFIG", columnDefinition = "json")
```
swaroopar, line 122: i would avoid this. Because the columns on this table will be then growing with every new feature added. Instead, can we simply add a new plain JSON  'REQUEST_BODY' and the request body to it. Do you think this would work?

If yes, then we must enter the data for deploy, modifiy, migrate, lock, etc

swaroopar (changes requested):

wuzhuanhong:
LGTM

luoping-12345:
LGTM

swaroopar (approved):
LGTM

PR #2072, opened 2024-10-28
remove cascade from child to parent table

### LINKED ISSUE #2049: remove cascade from child to parent table
In the PR which made a foreign key relation between DEPLOY_SERVICE and SERVICE_ORDER tables, it was requested to have the cascade only from parent to child entity. But somehow this was missed during the subsequent reviews. 

Please check and remove the cascade from child to parent tables. 
And on the parent table, we just need cascade 'cascade = CascadeType.REMOVE'. Please check if we really need all types of transactions to be cascaded to child entities. 

https://github.com/eclipse-xpanse/xpanse/blob/bbb0bcc4c57f72ff257c97f4561a9e7cc740a937/modules/database/src/main/java/org/eclipse/xpanse/modules/database/serviceorder/ServiceOrderEntity.java#L58

WangLiNaruto:
Fixed #2049


baixinsui:
LGTM

swaroopar (approved):
LGTM

PR #2095, opened 2024-11-05
When there is no Service Configuration, the Myservice page Current Config still shows the problem modification

### LINKED ISSUE #2073: When there is no Service Configuration, the Myservice page Current Config still shows the problem modification
When there is no Service Configuration, the Myservice page Current Config still shows the problem modification
image

WangLiNaruto:
Fixed #2073

image
image
image

image
image


Alice1319:
LGTM

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/ServiceDetailsViewManager.java
```diff
@@ -175,7 +191,20 @@ public VendorHostedDeployedServiceDetails getVendorHostedServiceDetailsByIdForEn
             log.error(errorMsg);
             throw new ServiceDetailsNotAccessible(errorMsg);
         }
-        return EntityTransUtils.transToVendorHostedServiceDetails(serviceDeploymentEntity);
+        VendorHostedDeployedServiceDetails details =
+                EntityTransUtils.transToVendorHostedServiceDetails(serviceDeploymentEntity);
+        ServiceTemplateEntity serviceTemplateEntity = serviceTemplateStorage
```
baixinsui, line 196: It is better to add a method to query all templates in batches, instead of invoking the query SQL statement one by one in a loop. Or we can modify the serviceTemplateId  to serviceTemplateEntity with @oneToone in the ServiceDeploymentEntity, in this way, ServiceTemplateEntity will be as a part of ServiceDeploymentEntity.

yuanxiaojun987:
LGTM

profoundwu:
LGTM

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/ServiceConfigurationManager.java
```diff
@@ -133,11 +133,11 @@ public ServiceOrder changeServiceConfiguration(String serviceId,
         try {
             ServiceDeploymentEntity serviceDeploymentEntity = serviceDeploymentEntityHandler
                     .getServiceDeploymentEntity(UUID.fromString(serviceId));
-            ServiceTemplateEntity serviceTemplateEntity = serviceTemplateStorage
-                    .getServiceTemplateById(serviceDeploymentEntity.getServiceTemplateId());
+            ServiceTemplateEntity serviceTemplateEntity =
```
swaroopar, line 136: for all methods in this class, it is better to change the type of serviceId to UUID instead of string.

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/ServiceDetailsViewManager.java
```diff
@@ -199,7 +214,69 @@ public List<DeployedService> listDeployedServicesOfIsv(Category category, Csp cs
...
+                                configurationParameters, serviceTemplate.getLastModifiedTime());
+                details.setServiceConfigurationDetails(serviceConfigurationDetails);
+            }
+        }
+    }
+
+    private ServiceConfigurationDetails getServiceConfigurationDetails(UUID serviceId,
+            List<ServiceConfigurationParameter> parameters, OffsetDateTime updateTime) {
+        ServiceConfigurationDetails details = new ServiceConfigurationDetails();
+        details.setServiceId(serviceId);
```
swaroopar, line 246: we must remove the serviceId field in the ServiceConfigurationDetails. we do not need it.

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/ServiceDetailsViewManager.java
```diff
@@ -199,7 +214,69 @@ public List<DeployedService> listDeployedServicesOfIsv(Category category, Csp cs
...
+        Map<String, Object> configurationParameterMap = new HashMap<>();
```
swaroopar, line 247: this does not match to what we discussed. 
We take what is available in the 'SERVICE_CONFIGURATION' and then merge it with values from the service entity.

swaroopar (changes requested):

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/ServiceDetailsViewManager.java
```diff
@@ -199,7 +213,84 @@ public List<DeployedService> listDeployedServicesOfIsv(Category category, Csp cs
...
+    private ServiceConfigurationDetails getServiceConfigurationDetails(
+            List<ServiceConfigurationParameter> parameters, Map<String, Object> configuration,
+            OffsetDateTime updateTime) {
+        ServiceConfigurationDetails details = new ServiceConfigurationDetails();
+        Map<String, Object> configurationParameterMap = new HashMap<>();
+        parameters.forEach(configurationParameter -> {
+            configurationParameterMap.put(configurationParameter.getName(),
+                    configurationParameter.getInitialValue());
+        });
+        if (!CollectionUtils.isEmpty(configuration)) {
```
swaroopar, line 253: does this also handle the 'merge' use case?
where the `configuration` is not empty but the `parameters` in the template have change (some new parameters added and some old ones removed)?

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/ServiceDetailsViewManager.java
```diff
@@ -199,7 +213,84 @@ public List<DeployedService> listDeployedServicesOfIsv(Category category, Csp cs
...
+                ServiceConfigurationDetails serviceConfigurationDetails =
+                        getServiceConfigurationDetails(configurationParameters,
+                                configuration,
+                                serviceTemplate.getLastModifiedTime());
+                details.setServiceConfigurationDetails(serviceConfigurationDetails);
+            }
+        }
+    }
+
+    private ServiceConfigurationDetails getServiceConfigurationDetails(
```
swaroopar, line 244: rename this to 'mergeConfigurationParametersFromTemplate'
Also add a method doc -> Method to dynamically merge and generate the current configuration of a service in case the service configuration parameters have been updated in the template by the ISV after a service was deloyed.

swaroopar (changes requested):

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/ServiceDetailsViewManager.java
```diff
@@ -199,7 +213,94 @@ public List<DeployedService> listDeployedServicesOfIsv(Category category, Csp cs
         String namespace = userServiceHelper.getCurrentUserManageNamespace();
         query.setNamespace(namespace);
         List<ServiceDeploymentEntity> deployServices = serviceDeploymentStorage.listServices(query);
-        return deployServices.stream().map(EntityTransUtils::convertToDeployedService).toList();
+        return setServiceConfigurationForDeployedService(deployServices);
+    }
+
+    private void setServiceConfigurationDetailsForDeployedService(DeployedService details,
+            ServiceDeploymentEntity serviceDeployment) {
```
swaroopar, line 220: Do we really need the serviceDeployment argument? the serviceTemplateId is also available in the DeployedService object.

swaroopar (changes requested):

swaroopar (approved):
LGTM

PR #2133, opened 2024-11-20
Add new table SERVICE_TEMPLATE_HISTORY

baixinsui:
Fixed https://github.com/eclipse-xpanse/xpanse/issues/2099

Constructor of table service_template_history in H2 database:
chrome_D4dDXv5RsT

Constructor of table service_template_history in MySql database:
image

### DIFF
--- a/modules/api/src/test/java/org/eclipse/xpanse/api/config/ServiceTemplateRegistrationStateConverterTest.java
+++ b/modules/api/src/test/java/org/eclipse/xpanse/api/config/ServiceTemplateRegistrationStateConverterTest.java
@@ -15,7 +15,7 @@ class ServiceTemplateRegistrationStateConverterTest {
     @Test
     void testConvert() {
         assertEquals(ServiceTemplateRegistrationState.IN_PROGRESS,
-                converterTest.convert("inProgress"));
+                converterTest.convert("in-progress"));
         assertEquals(ServiceTemplateRegistrationState.APPROVED, converterTest.convert("approved"));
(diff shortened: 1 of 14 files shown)


modules/api/src/main/java/org/eclipse/xpanse/api/config/ServiceTemplateEntityConverter.java
```diff
@@ -25,7 +25,7 @@
 public class ServiceTemplateEntityConverter {
 
     /**
-     * Convert ServiceTemplateEntity to ServiceTemplateDetailVo.
+     * Convert ServiceTemplateHistoryEntity to ServiceTemplateDetailVo.
```
swaroopar, line 28: i think this comment change is not correct.

modules/api/src/main/java/org/eclipse/xpanse/api/config/ServiceTemplateEntityConverter.java
```diff
@@ -69,7 +69,7 @@ public static ServiceTemplateDetailVo convertToServiceTemplateDetailVo(
     }
 
     /**
-     * Convert ServiceTemplateEntity to UserOrderableServiceVo.
+     * Convert ServiceTemplateHistoryEntity to UserOrderableServiceVo.
```
swaroopar, line 72: please check comment change.

swaroopar (changes requested):

swaroopar (approved):
LGTM

PR #2135, opened 2024-11-21
implement changes to service template registration method

### LINKED ISSUE #2100: implement changes to service template registration method
image

- [ ] Add entry additionally to SERVICE_TEMPLATE_HISTORY table with REQUEST_TYPE as CREATE.
- [ ] Status in SERVICE_TEMPLATE must be a new value ‘IN_PROGRESS’ if review is is pending. otherwise, we can directly set it to ACCEPTED: 
- [ ] Return Serviceid and Change Id.

baixinsui:
**Fixed #2100** 

**Register service template :**
image

### DIFF
--- a/modules/servicetemplate/src/main/java/org/eclipse/xpanse/modules/servicetemplate/ServiceTemplateManage.java
+++ b/modules/servicetemplate/src/main/java/org/eclipse/xpanse/modules/servicetemplate/ServiceTemplateManage.java
@@ -22,6 +22,8 @@ import org.eclipse.xpanse.modules.database.service.ServiceQueryModel;
 import org.eclipse.xpanse.modules.database.servicetemplate.ServiceTemplateEntity;
 import org.eclipse.xpanse.modules.database.servicetemplate.ServiceTemplateQueryModel;
 import org.eclipse.xpanse.modules.database.servicetemplate.ServiceTemplateStorage;
+import org.eclipse.xpanse.modules.database.servicetemplatehistory.ServiceTemplateHistoryEntity;
+import org.eclipse.xpanse.modules.database.servicetemplatehistory.ServiceTemplateHistoryStorage;
 import org.eclipse.xpanse.modules.deployment.DeployerKindManager;
 import org.eclipse.xpanse.modules.models.common.enums.Csp;
 import org.eclipse.xpanse.modules.models.common.exceptions.OpenApiFileGenerationException;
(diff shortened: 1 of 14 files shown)


modules/models/src/main/java/org/eclipse/xpanse/modules/models/servicetemplate/view/ServiceTemplateHistoryVo.java
```diff
@@ -0,0 +1,31 @@
...
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+/**
+ * Defines view object for service template history.
+ */
+@Data
+@AllArgsConstructor
+@NoArgsConstructor
+public class ServiceTemplateHistoryVo {
```
swaroopar, line 22: We need a better name for this type. Maybe 'ServiceTemplateChangeInfo' ?
Also we can move it out of the view package.

swaroopar (changes requested):

modules/servicetemplate/src/main/java/org/eclipse/xpanse/modules/servicetemplate/ServiceTemplateManage.java
```diff
@@ -162,12 +168,26 @@ private ServiceTemplateEntity getNewServiceTemplateEntity(Ocl ocl) {
...
+    private ServiceTemplateHistoryEntity createServiceTemplateHistory(
+            ServiceTemplateRequestType requestType, ServiceTemplateEntity serviceTemplate) {
+        ServiceTemplateHistoryEntity serviceTemplateHistory = new ServiceTemplateHistoryEntity();
+        serviceTemplateHistory.setChangeId(UUID.randomUUID());
+        serviceTemplateHistory.setServiceTemplate(serviceTemplate);
+        serviceTemplateHistory.setOcl(serviceTemplate.getOcl());
+        serviceTemplateHistory.setRequestType(requestType);
+        Csp csp = serviceTemplate.getCsp();
+        if (isAutoApproveEnabledForCsp(csp)) {
+            serviceTemplateHistory.setStatus(ServiceTemplateChangeStatus.ACCEPTED);
```
swaroopar, line 180: can we also write a default ' reviewComment' -> auto-approved by CSP

swaroopar (changes requested):

runtime/src/test/java/org/eclipse/xpanse/runtime/ServiceTemplateApiTest.java
```diff
@@ -69,7 +71,6 @@
  * Test for ServiceTemplateManageApi.
  */
 @Slf4j
-@Transactional
```
swaroopar, line 72: the import for this is not removed?

swaroopar (approved):
LGTM

PR #2137, opened 2024-11-21
Add full order result information to service_order table and ServiceOrderStatusUpdate

### LINKED ISSUE #2082: Add full order result information to service_order table and ServiceOrderStatusUpdate
- [ ] Rename enum `ResultType` to `ErrorType` and remove `SUCCESS` from it. (We never use SUCCESS or return this enum in case of successful execution)
- [ ] Rename type `Response` to `ErrorResponse`
- [ ]  Remove property `success` from `Response` type. (It is always set to false in our case)
- [ ] `ServiceOrderStatusUpdate` - change property errorMsg to 'error' and type `ErrorResponse`
- [ ] In SERVICE_ORDER entity, change the `ERROR_MSG` column to type 'Response' and also rename column to `ERROR`
- [ ] Compile interface on UI and make necessary changes to display correct error response from `ServiceOrderStatusUpdate

WangLiNaruto:
Fixed #2082


baixinsui:
LGTM

modules/models/src/main/java/org/eclipse/xpanse/modules/models/response/ErrorType.java
```diff
@@ -78,11 +77,14 @@ public enum ResultType {
     SERVICE_CONFIG_UPDATE_REQUEST_NOT_FOUND("Service Configuration Update Request Not Found"),
     SERVICE_CONFIGURATION_NOT_FOUND("Service Configuration Not Found"),
     INVALID_DEPLOYER_TOOL("Invalid Deployer Tool"),
-    DEPLOYMENT_SCRIPTS_CREATION_FAILED("Deployment Scripts Creation Failed");
+    DEPLOYMENT_SCRIPTS_CREATION_FAILED("Deployment Scripts Creation Failed"),
+    ASYN_START_SERVICE_ERROR("Async Start Service Error"),
```
swaroopar, line 81: ASYN must be ASYNC for all three enums

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/ServiceOrderManager.java
```diff
@@ -117,7 +119,8 @@ public void completeOrderProgress(UUID orderId, TaskStatus taskStatus, String er
         }
         serviceOrder.setTaskStatus(taskStatus);
         serviceOrder.setCompletedTime(OffsetDateTime.now());
-        serviceOrder.setErrorMsg(errorMsg);
+        serviceOrder.setError(ErrorResponse
```
swaroopar, line 122: in line 112 - the argument type for errorMsg must be  ErrorResponse and let calling methods to define the correct exception.

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/ServiceStateManager.java
```diff
@@ -98,7 +100,9 @@ private void asyncStartService(ServiceOrderEntity serviceOrderTaskEntity,
         try {
             result = plugin.startService(request);
         } catch (Exception e) {
-            serviceOrderTaskEntity.setErrorMsg(e.getMessage());
+            serviceOrderTaskEntity.setError(
```
swaroopar, line 103: Its better we use the methods in SerivceOrderManager class to update in-progress, completed, etc.

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployResultManager.java
```diff
@@ -110,7 +112,8 @@ public void updateServiceWithDeployResult(DeployResult deployResult) {
         boolean isTaskSuccessful = deployResult.getIsTaskSuccessful();
         if (isFailedDeployTask(isTaskSuccessful, taskType)) {
             storedOrderEntity.setTaskStatus(TaskStatus.FAILED);
-            storedOrderEntity.setErrorMsg(deployResult.getMessage());
+            storedOrderEntity.setError(ErrorResponse
```
swaroopar, line 115: same here.. better we use the ServiceOrderManager here as well.

swaroopar (changes requested):

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/ServiceOrderManager.java
swaroopar, line 122: For all deployment failures, for example in https://github.com/eclipse-xpanse/xpanse/blob/f332609e15cac28b6181c1962603357b3306abaa/modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/migration/steps/ProcessDeploymentResult.java#L88

we can define a new ErrorType -> DeploymentFailedException

swaroopar (changes requested):

swaroopar (approved):
LGTM

PR #2151, opened 2024-11-26
Implement changes to update service template method

baixinsui:
**Fixed https://github.com/eclipse-xpanse/xpanse/issues/2101
Related to https://github.com/eclipse-xpanse/xpanse-ui/pull/1310**

**1. auto-approve of service template by Csp HuaweiCloud is enabled:**
- Register a service template of HuaweiCloud:
image
- Update service template of HuaweiCloud:
image
**2. auto-approve of service template by Csp FlexibleEngine is disabled:**
- Register a service template of FlexibleEngine:
image
- Approve service template of FlexibleEngine:
image
- Update service template of FlexibleEngine with param isRemoveServiceTemplateUntilApproved=false
image
- If the value of is_update_pending is true, update service template will throw not allowed exception:
image


Alice1319:
LGTM

modules/servicetemplate/src/main/java/org/eclipse/xpanse/modules/servicetemplate/ServiceTemplateManage.java
```diff
@@ -91,36 +93,78 @@ public class ServiceTemplateManage {
...
+        }
+        // update service template openApi doc only when service template is available in catalog
+        if (updatedServiceTemplate.getAvailableInCatalog()) {
+            serviceTemplateOpenApiGenerator.updateServiceApi(updatedServiceTemplate);
+        }
+        return storedUpdateHistory;
+    }
+
+
+    private ServiceTemplateHistoryEntity queryUpdatingHistoryForServiceTemplate(
```
swaroopar, line 160: queryAnyInProgressServiceTemplateUpdateRequest

modules/servicetemplate/src/main/java/org/eclipse/xpanse/modules/servicetemplate/ServiceTemplateManage.java
```diff
@@ -345,6 +384,21 @@ public ServiceTemplateEntity getServiceTemplateDetails(UUID id, boolean checkNam
...
+                                            ServiceTemplateRequestType requestType,
+                                            ServiceTemplateEntity serviceTemplate) {
+        if (isAutoApproveEnabled) {
+            serviceTemplate.setServiceTemplateRegistrationState(
+                    ServiceTemplateRegistrationState.APPROVED);
+            serviceTemplate.setAvailableInCatalog(true);
+            serviceTemplate.setIsUpdatePending(false);
+        } else {
+            serviceTemplate.setIsUpdatePending(requestType == ServiceTemplateRequestType.UPDATE);
+            serviceTemplate.setServiceTemplateRegistrationState(
```
swaroopar, line 397: the registration state must not change with update. 
The registration state is only changed and managed with first/initial registration. After that it remains untouched.

modules/servicetemplate/src/main/java/org/eclipse/xpanse/modules/servicetemplate/ServiceTemplateManage.java
```diff
@@ -345,6 +384,21 @@ public ServiceTemplateEntity getServiceTemplateDetails(UUID id, boolean checkNam
         return existingTemplate;
     }
 
+    private void updateServiceTemplateState(boolean isAutoApproveEnabled,
+                                            ServiceTemplateRequestType requestType,
+                                            ServiceTemplateEntity serviceTemplate) {
+        if (isAutoApproveEnabled) {
+            serviceTemplate.setServiceTemplateRegistrationState(
```
swaroopar, line 391: same here. registration state must not touched during 'update' process.

modules/servicetemplate/src/main/java/org/eclipse/xpanse/modules/servicetemplate/ServiceTemplateManage.java
```diff
@@ -345,6 +384,21 @@ public ServiceTemplateEntity getServiceTemplateDetails(UUID id, boolean checkNam
         return existingTemplate;
     }
 
+    private void updateServiceTemplateState(boolean isAutoApproveEnabled,
```
swaroopar, line 387: This method is a bit confusing from the caller perspective because sometimes we send old template entity and sometimes new template entity. Would be better if we can split this method or any other solution can be used.

modules/servicetemplate/src/main/java/org/eclipse/xpanse/modules/servicetemplate/ServiceTemplateManage.java
```diff
@@ -91,36 +93,78 @@ public class ServiceTemplateManage {
...
+            updateServiceTemplateState(false, ServiceTemplateRequestType.UPDATE,
+                    existingServiceTemplate);
+            if (existingServiceTemplate.getAvailableInCatalog()
+                    && isRemoveFromCatalogUntilApproved) {
+                existingServiceTemplate.setAvailableInCatalog(false);
+            }
+            updatedServiceTemplate = templateStorage.storeAndFlush(existingServiceTemplate);
+        }
+        // update service template openApi doc only when service template is available in catalog
+        if (updatedServiceTemplate.getAvailableInCatalog()) {
```
swaroopar, line 153: here the value can be true but the update can be still in-review. So better to keep this as open-api generation as part of the block in line 139

modules/api/src/main/java/org/eclipse/xpanse/api/controllers/ServiceTemplateApi.java
```diff
@@ -100,12 +100,17 @@ public ServiceTemplateChangeInfo register(@Valid @RequestBody Ocl ocl) {
...
-    public ServiceTemplateDetailVo update(
+    public ServiceTemplateChangeInfo update(
             @Parameter(name = "id", description = "id of service template") @PathVariable("id")
-            String id, @Valid @RequestBody Ocl ocl) {
-        ServiceTemplateEntity templateEntity =
-                serviceTemplateManage.updateServiceTemplate(UUID.fromString(id), ocl);
-        return convertToServiceTemplateDetailVo(templateEntity);
+            String id,
+            @Parameter(name = "isRemoveServiceTemplateUntilApproved", description =
+                    "If remove the service template from catalog until the updated one is approved")
```
swaroopar, line 107: the description is good. just one small extension to make it more explicit.

If true, the old service template is also removed from catalog until the updated one is reviewed and approved

swaroopar (changes requested):

WangLiNaruto:
LGTM

swaroopar (approved):
LGTM

PR #2159, opened 2024-11-29
update terraform-boot and tofu-maker client api

WangLiNaruto:


baixinsui (commented):
The generated code should be formatted.

swaroopar (approved):
LGTM

PR #2182, opened 2024-12-06
Add config to avoid always loading all available versions of deployer at boot

baixinsui:
Fixed https://github.com/eclipse-xpanse/xpanse/issues/2136
Related to https://github.com/eclipse-xpanse/tofu-maker/pull/90
Related to https://github.com/eclipse-xpanse/terraform-boot/pull/172

Start xpanse with config `support.default.deployment.tool.versions.only=true`:
Get default versions from config to cache:
image

Start xpanse with config `support.default.deployment.tool.versions.only=false`:
Get all available versions from website to cache:
image


Alice1319:
LGTM

yuanxiaojun987:
LGTM

runtime/src/main/resources/application.properties
```diff
@@ -71,6 +71,7 @@ huaweicloud.european.price.calculator.url=https://portal.eu.huaweicloud.com/api/
 huaweicloud.chinese.price.calculator.url=https://portal.huaweicloud.com/api/cbc/global/rest/BSS/billing/ratingservice/v2/inquiry/resource
 enable.agent.api.only=false
 agent.api.end.point=http://localhost:8080
+support.default.deployment.tool.versions.only=true
 deployer.terraform.install.dir=/opt/terraform
 deployer.terraform.versions=1.6.0,1.7.0,1.8.0,1.9.0
```
swaroopar, line 76: can we rename this to deployer.terraform.default.supported.versions pls.
and for opentofu - deployer.opentofu.default.supported.versions pls

swaroopar (changes requested):

runtime/src/main/resources/application.properties
baixinsui: renamed

swaroopar (approved):

Issue #2188, opened 2024-12-10
rename namespace field to 'serviceVendor'


swaroopar:
Currently the field 'namespace' is not very self-explanatory in the context of xpanse. We will need to rename it to something that exactly represents the underlying meaning of the word.

- [x] in the OCL - change this value to serviceVendor
- [x] in the oauth metadata - the name of the field must be 'isv'
- [x] Update samples. 
- [x] Update UI
- [x] Update xpanse-iam
- [x] Update documentation in xpanse-website

baixinsui:
I will take it

PR #2224, opened 2024-12-18
rename namespace field to serviceVendor.

Alice1319:
fixes: #2188


modules/security/src/main/java/org/eclipse/xpanse/modules/security/common/CurrentUserInfo.java
```diff
@@ -22,7 +22,7 @@ public class CurrentUserInfo {
 
     private Map<String, String> metadata;
 
-    private String namespace;
+    private String serviceVendor;
```
baixinsui, line 25: This should be 'isv'

baixinsui (commented):
All keys 'namespace' should be updated to 'isv' in all jwt_xxx.json as below:
image

modules/security/src/main/java/org/eclipse/xpanse/modules/security/common/CurrentUserInfo.java
Alice1319: updated

modules/api/src/main/java/org/eclipse/xpanse/api/config/ServiceTemplateEntityConverter.java
```diff
@@ -37,10 +37,11 @@ public static ServiceTemplateDetailVo convertToServiceTemplateDetailVo(
             serviceTemplateDetailVo.setServiceTemplateId(serviceTemplateEntity.getId());
             serviceTemplateDetailVo.setIcon(serviceTemplateEntity.getOcl().getIcon());
             serviceTemplateDetailVo.setDescription(serviceTemplateEntity.getOcl().getDescription());
-            if (StringUtils.isNotEmpty(serviceTemplateEntity.getNamespace())) {
-                serviceTemplateDetailVo.setNamespace(serviceTemplateEntity.getNamespace());
+            if (StringUtils.isNotEmpty(serviceTemplateEntity.getServiceVendor())) {
+                serviceTemplateDetailVo.setServiceVendor(serviceTemplateEntity.getServiceVendor());
             } else {
```
baixinsui, line 42: The if... else... code is unnecessary, the value already has been set by called BeanUtils.copyProperties(serviceTemplateEntity, serviceTemplateDetailVo) .

baixinsui:
LGTM

swaroopar (approved):
LGTM

luoping-12345:
LGTM

PR #2237, opened 2024-12-24
change all ID parameters in URI to UUID.

### LINKED ISSUE #2204: change all ID parameters in URI to UUID
Currently we use string for the ID field in some places and UUID field in someplaces. When we are using string, we are again internally creating UUID out of it.  This can avoided. 

- [ ] Check all controller classes and set ID to type UUID
- [ ] fix the audit annotation to pass the correct type.

Alice1319:
fixes #2204

### DIFF
--- a/modules/api/src/main/java/org/eclipse/xpanse/api/config/GetCspInfoFromRequest.java
+++ b/modules/api/src/main/java/org/eclipse/xpanse/api/config/GetCspInfoFromRequest.java
@@ -87,18 +87,18 @@ public class GetCspInfoFromRequest {
     /**
      * Get Csp with id of service.
      *
-     * @param id id of service.
+     * @param serviceId id of service.
      * @return csp.
      */
-    public Csp getCspFromServiceId(String id) {
+    public Csp getCspFromServiceId(UUID serviceId) {
         try {
             ServiceDeploymentEntity deployService =
-                    deployServiceStorage.findServiceDeploymentById(UUID.fromString(id));
+                    deployServiceStorage.findServiceDeploymentById(serviceId);
             if (Objects.nonNull(deployService)) {
                 return deployService.getCsp();
             }
         } catch (Exception e) {
-            log.error("Get csp with service id:{} failed.", id, e);
+            log.error("Get csp with service id:{} failed.", serviceId, e);
         }
         return null;
     }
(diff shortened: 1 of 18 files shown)


WangLiNaruto:
LGTM

profoundwu:
LGTM

modules/api/src/main/java/org/eclipse/xpanse/api/config/GetCspInfoFromRequest.java
```diff
@@ -146,9 +145,9 @@ public Csp getCspFromServicePolicyId(String id) {
      * @param taskId id of migrate task.
      * @return csp.
      */
-    public Csp getCsFromWorkflowTaskId(String taskId) {
+    public Csp getCspFromWorkflowTaskId(UUID taskId) {
         try {
```
baixinsui, line 149: The taskId in workflow is not a UUID. Just keep type String

modules/api/src/main/java/org/eclipse/xpanse/api/controllers/WorkFlowApi.java
```diff
@@ -67,31 +68,33 @@ public List<WorkFlowTask> queryTasks(
...
     @Tag(name = "Workflow", description = "APIs to manage the Workflow")
     @Operation(description = "Complete tasks by task ID and set global process variables .")
-    @PutMapping(value = "/workflow/complete/task/{id}", produces = MediaType.APPLICATION_JSON_VALUE)
+    @PutMapping(
+            value = "/workflow/complete/task/{taskId}",
+            produces = MediaType.APPLICATION_JSON_VALUE)
     @ResponseStatus(HttpStatus.OK)
-    @AuditApiRequest(methodName = "getCsFromWorkflowTaskId")
+    @AuditApiRequest(methodName = "getCsFromWorkflowTaskId", paramTypes = UUID.class)
     public void completeTask(
```
baixinsui, line 76: The taskId in workflow is not a UUID. Just keep type String

modules/api/src/main/java/org/eclipse/xpanse/api/config/GetCspInfoFromRequest.java
Alice1319: updated

modules/api/src/main/java/org/eclipse/xpanse/api/controllers/WorkFlowApi.java
Alice1319: updated.

luoping-12345:
LGTM

swaroopar (approved):
LGTM

PR #2306, opened 2025-01-16
update no permission message for service template requests

baixinsui:
Fixed https://github.com/eclipse-xpanse/xpanse/issues/2299

User ISV-B register service template with service vendor ISV-A:
image
User ISV-B list requests to service template with service vendor ISV-A:
image
User ISV-B cancel a request to service template with service vendor ISV-A:
image

### DIFF
--- a/modules/api/src/main/java/org/eclipse/xpanse/api/controllers/ServiceTemplateApi.java
+++ b/modules/api/src/main/java/org/eclipse/xpanse/api/controllers/ServiceTemplateApi.java
@@ -27,6 +27,7 @@ import org.eclipse.xpanse.modules.database.servicetemplate.ServiceTemplateQueryM
 import org.eclipse.xpanse.modules.database.servicetemplaterequest.ServiceTemplateRequestHistoryEntity;
 import org.eclipse.xpanse.modules.models.common.enums.Category;
 import org.eclipse.xpanse.modules.models.common.enums.Csp;
+import org.eclipse.xpanse.modules.models.common.enums.UserOperation;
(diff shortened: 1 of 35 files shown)


ruwenqiang123:
LGTM

Alice1319:
LGTM

qigaoneng:
LGTM

modules/api/src/main/java/org/eclipse/xpanse/api/controllers/ServiceTemplateApi.java
```diff
@@ -338,8 +338,10 @@ public ServiceTemplateDetailVo getServiceTemplateDetailsById(
             @Parameter(name = "serviceTemplateId", description = "id of service template")
                     @PathVariable("serviceTemplateId")
                     UUID serviceTemplateId) {
+        String operation = "view details of service template";
```
swaroopar, line 341: i think it would be good to make the operation also an internal Enum?  and just have one word - Eg., 'view details'

because currently the sentence is not very clear - 

"No permissions to view details of service temple, the service template belonging to other service vendors."
Better would be -> no permission to view details of the service templates owned by other service vendors. 

same for CSP also.

swaroopar (changes requested):

modules/api/src/main/java/org/eclipse/xpanse/api/controllers/ServiceTemplateApi.java
baixinsui: Updated all error messages about no permission

swaroopar (approved):
LGTM

PR #2314, opened 2025-01-17
isv credential set managed service vendor as userId

baixinsui:
Fixed https://github.com/eclipse-xpanse/xpanse/issues/2308

ISV user ISV-A add credential with key which contains managed isv by user:
chrome_laWrXxC8Fo
ISV users manage other isv could not view the created credential anymore:
image

### DIFF
--- a/modules/api/src/main/java/org/eclipse/xpanse/api/controllers/IsvCloudCredentialsApi.java
+++ b/modules/api/src/main/java/org/eclipse/xpanse/api/controllers/IsvCloudCredentialsApi.java
@@ -11,6 +11,7 @@ import static org.eclipse.xpanse.modules.security.common.RoleConstants.ROLE_ISV;
 import io.swagger.v3.oas.annotations.Operation;
 import io.swagger.v3.oas.annotations.Parameter;
 import io.swagger.v3.oas.annotations.tags.Tag;
+import jakarta.annotation.Resource;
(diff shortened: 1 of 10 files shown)


ruwenqiang123:
LGTM

modules/api/src/main/java/org/eclipse/xpanse/api/controllers/IsvCloudCredentialsApi.java
```diff
@@ -148,6 +147,10 @@ public void deleteIsvCloudCredential(
             @Parameter(name = "name", description = "The name of credential.")
                     @RequestParam(name = "name")
                     String name) {
-        credentialCenter.deleteCredential(csp, siteName, type, name, null);
+        credentialCenter.deleteCredential(csp, siteName, type, name, getManagedIsv());
+    }
+
+    private String getManagedIsv() {
```
swaroopar, line 153: i think the word 'managedIsv' does not give a clear meaning here. 

I see the method names in the UserHelper needs to be corrected. 
getCurrentUserManageIsv() -> getIsvManagedByCurrentUser()
getCurrentUserManageCsp() -> getCspManagedByCurrentUser()

And then use the similar namings at other places.

swaroopar (changes requested):

swaroopar (approved):
LGTM

PR #2321, opened 2025-01-21
add service actions to UI

### LINKED ISSUE #2293: add service actions to UI
- [x] add new 'actions' option to myservices drop down. 
- [x] On click - the modal must show all actions possible - on the either a tab like option on the top or on the left border for the user to choose between different actions. 
- [x] On section of specific action tab - the corresponding variables must be displayed - similar to new order or config update. 
- [x] On submit, the action order must be submitted. 

The exact UI design can be discussed during the implementation

WangLiNaruto:
related to #2293

### DIFF
--- a/modules/api/src/main/java/org/eclipse/xpanse/api/config/ServiceTemplateEntityConverter.java
+++ b/modules/api/src/main/java/org/eclipse/xpanse/api/config/ServiceTemplateEntityConverter.java
@@ -102,6 +102,10 @@ public class ServiceTemplateEntityConverter {
                 userOrderableServiceVo.setConfigurationParameters(
                         serviceConfigurationManage.getConfigurationParameters());
             }
+            if (Objects.nonNull(serviceTemplateEntity.getOcl().getServiceActions())) {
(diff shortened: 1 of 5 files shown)


modules/models/src/main/java/org/eclipse/xpanse/modules/models/service/view/DeployedService.java
```diff
@@ -125,4 +127,7 @@ public class DeployedService {
 
     @Schema(description = "Details of the service configuration.")
     private ServiceConfigurationDetails serviceConfigurationDetails;
+
+    @Schema(description = "manage service action.")
+    private List<ServiceAction> serviceActions;
```
swaroopar, line 132: these are the service actions from the template. not the service actions ordered by the end user. 
If the end user want to display the service actions that he can order, then we must call the getOrderableServiceDetailsByServiceId which must return the service actions.

swaroopar (changes requested):

swaroopar (approved):
LGTM

PR #2345, opened 2025-01-29
fetch user id in audit only for authenticated requests

swaroopar:
fixes https://github.com/eclipse-xpanse/xpanse/issues/2329


iskey (approved):
LGTM

PR #2352, opened 2025-01-29
fix rest request logger to log in correct sequence

swaroopar:
fixes https://github.com/eclipse-xpanse/xpanse/issues/2351


iskey (approved):
LGTM

PR #2362, opened 2025-02-02
fix secrets encryption and decryption for credentials

swaroopar:
fixes https://github.com/eclipse-xpanse/xpanse/issues/2358


swaroopar:
@iskey the number of files is a bit high due to package restructuring. The actual changes are only in  few classes.

iskey (approved):
LGTM

PR #2400, opened 2025-02-13
fix refetch to update deployment state and process the result

swaroopar:
fixes https://github.com/eclipse-xpanse/xpanse/issues/2399


iskey (approved):
LGTM

Issue #2379, opened 2025-02-08
make healthcheck only for admin users and create new service for all users


swaroopar:
Currently the healthcheck service is allowed for all users. This must be changed. Since only admin must know all configuration and backend systems involved the entire stack. 

- [ ]  remove processShownFields method - https://github.com/eclipse-xpanse/xpanse/blob/1f8d0a24f8d7b4d28a8426c069c0646fcd76ea74/modules/api/src/main/java/org/eclipse/xpanse/api/controllers/AdminServicesApi.java#L193
- [ ] Allow healthCheck() method only for 'ADMIN' ROLE and rename method to stackHealthStatus() -> /xpanse/stack/health
- [ ] Rename existing SystemStatus data model to -> StackStatus
- [ ] create new method -> healthCheck() -> which returns only new SystemStatus data model which has only 'HealthStatus' as a property. 
- [ ] Allow new method to all roles - admin, user, isv and csp. 
- [ ] Update UI to use the new healthCheck in layout header. and stackHealthStatus() in admin's healthcheck component.

AKARSHITJOSHI:
@swaroopar I would love to give this a try.

swaroopar:
@AKARSHITJOSHI are you working on this?

AKARSHITJOSHI:
> [@AKARSHITJOSHI](https://github.com/AKARSHITJOSHI) are you working on this?

Yes I will raise PR be EOD.

PR #2405, opened 2025-02-13
#2379 Make healthcheck only for admin users and create new service for all users

AKARSHITJOSHI:
1. remove processShownFields method -
[xpanse/modules/api/src/main/java/org/eclipse/xpanse/api/controllers/AdminServicesApi.java](https://github.com/eclipse-xpanse/xpanse/blob/1f8d0a24f8d7b4d28a8426c069c0646fcd76ea74/modules/api/src/main/java/org/eclipse/xpanse/api/controllers/AdminServicesApi.java#L193)

Line 193 in [1f8d0a2](https://github.com/eclipse-xpanse/xpanse/commit/1f8d0a24f8d7b4d28a8426c069c0646fcd76ea74)

 private void processShownFields(BackendSystemStatus backendSystemStatus) { 

2. Allow healthCheck() method only for 'ADMIN' ROLE and rename method to stackHealthStatus() -> /xpanse/stack/health

3.Rename existing SystemStatus data model to -> StackStatus

4. create new method -> healthCheck() -> which returns only new SystemStatus data model which has only 'HealthStatus' as a property.

5. Allow new method to all roles - admin, user, isv and csp.

### DIFF
--- a/modules/api/src/main/java/org/eclipse/xpanse/api/controllers/AdminServicesApi.java
+++ b/modules/api/src/main/java/org/eclipse/xpanse/api/controllers/AdminServicesApi.java
@@ -18,6 +18,10 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
+
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.NoArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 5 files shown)


swaroopar:
@AKARSHITJOSHI we cannot do anything until you sign the ECA. Kindly do that.

swaroopar:
no update. will close it for the moment. if the contributor is wishes, can be reopened.

PR #2406, opened 2025-02-13
decrypting fixed to be done only for credentials in cache

swaroopar:
fixes https://github.com/eclipse-xpanse/xpanse/issues/2401


baixinsui:
LGTM

PR #2408, opened 2025-02-14
Fix Maven Plugin Generate API Client Failed on Windows

baixinsui:
Fixed  https://github.com/eclipse-xpanse/xpanse/issues/2407

Susscced on Windows:
image
Susscced on Linux:
image


swaroopar (approved):
LGTM

PR #2418, opened 2025-02-16
Filter modified Service Configuration

### LINKED ISSUE #2411: changeServiceConfiguration always creates requests even for unchanged configuration
When user sends the service configuration request, we must always compare the current configuration. 
And then filter out only the configuration parameter that has been changed.  This is necessary because the user can even send the complete configuration again. 

The updateRequestMap sent to createServiceChangeRequestsInDatabase must contain only the changed properties. So the comparison must happen before that. 

https://github.com/eclipse-xpanse/xpanse/blob/8272a4982237eb8f583fd597fdebcbf40d55f374/modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/ServiceConfigurationManager.java#L155

Even merging configuration from the service template is wrong. We must consider the existing configuration data for the service instead. 
Even the  updating the result is wrong.

AKARSHITJOSHI:
Fixes #2411 

Update only newly added or modified service configuration.

### DIFF
--- a/modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/ServiceConfigurationManager.java
+++ b/modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/ServiceConfigurationManager.java
@@ -9,6 +9,7 @@ package org.eclipse.xpanse.modules.deployment;
 import jakarta.annotation.Resource;
 import jakarta.transaction.Transactional;
 import java.time.OffsetDateTime;
+import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


swaroopar:
@AKARSHITJOSHI Please only work on issues which are prior discussed and assigned to you . Otherwise this will only lead to duplicate work.

swaroopar:
This does not solve the problem fully. 
1. One is, we must filter only changed parameters.
2. Other is, for the changed parameter  - based on its change handler, we must send even the unchanged parameters. 
I'm working on the issue.

PR #2421, opened 2025-02-17
fix service configuration to create change requests only for changed parameters

swaroopar:
fixes https://github.com/eclipse-xpanse/xpanse/issues/2411


iskey (approved):
LGTM

PR #2455, opened 2025-02-24
changeServiceConfiguration allows multiple config change requests

### LINKED ISSUE #2410: changeServiceConfiguration allows multiple config change requests
When there is already one CONFIG_CHANGE order which is still not in state SUCCESSFUL, new config change requests must not be allowed. 

changeServiceConfiguration  must be changed. 

Image

WangLiNaruto:
Fixed #2410


baixinsui:
LGTM

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/ServiceConfigurationManager.java
```diff
@@ -133,6 +138,38 @@ public ServiceOrder changeServiceConfiguration(
...
+            boolean isAllConfigChangeOrdersCompleted =
+                    configChangeOrders.stream()
+                            .allMatch(
+                                    order ->
+                                            order.getOrderStatus() == OrderStatus.SUCCESSFUL
+                                                    || order.getOrderStatus()
+                                                            == OrderStatus.FAILED);
+
+            if (isAllConfigChangeOrdersCompleted) {
+                throw new ServiceOrderNotFound(
```
swaroopar, line 164: this exception is wrong. We must return - 'There is already a pending config change order. Order ID - xxxx'
We must also need a correct exception type here.

swaroopar (changes requested):

swaroopar (approved):
LGTM

PR #2470, opened 2025-02-28
Feature/generate UUID

### LINKED ISSUE #2428: use @GeneratedValue to generate UUID for the below entities
- [ ] ServiceTemplateRequestHistoryEntity
- [ ] ServiceTemplateEntity
- [ ] ServiceDeploymentEntity
- [ ] ServiceOrderEntity

This is necessary because sometimes we return the ID to client in case of exceptions but the ID is never stored. 

Please note that the currently we creating the ID and using it even before it is stored in DB. So the entire sequence must be changed with this.

WangLiNaruto:
Fixed #2428

register:
image
image
image

deployment:
image
image

recreate:
image
image

port:
image
image

### DIFF
--- a/modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java
+++ b/modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java
@@ -7,7 +7,6 @@
 package org.eclipse.xpanse.modules.deployment;
 
 import static org.eclipse.xpanse.modules.async.TaskConfiguration.ASYNC_EXECUTOR_NAME;
-import static org.eclipse.xpanse.modules.logging.LoggingKeyConstant.SERVICE_ID;
 
 import jakarta.annotation.Resource;
 import java.time.OffsetDateTime;
@@ -33,7 +32,6 @@ import org.eclipse.xpanse.modules.database.servicetemplate.ServiceTemplateQueryM
 import org.eclipse.xpanse.modules.database.servicetemplate.ServiceTemplateStorage;
 import org.eclipse.xpanse.modules.database.utils.EntityTranslationUtils;
 import org.eclipse.xpanse.modules.deployment.polling.ServiceDeploymentStatusChangePolling;
-import org.eclipse.xpanse.modules.logging.CustomRequestIdGenerator;
 import org.eclipse.xpanse.modules.models.common.enums.Csp;
 import org.eclipse.xpanse.modules.models.common.enums.UserOperation;
 import org.eclipse.xpanse.modules.models.service.config.ServiceLockConfig;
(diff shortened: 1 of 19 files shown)


modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java
```diff
@@ -423,6 +418,10 @@ private void deployService(DeployTask deployTask) {
         DeployerKind kind = deployTask.getOcl().getDeployment().getDeployerTool().getKind();
         Deployer deployer = deployerKindManager.getDeployment(kind);
         ServiceDeploymentEntity serviceEntity = storeNewDeployServiceEntity(deployTask);
+        if (Objects.nonNull(serviceEntity.getId())) {
```
swaroopar, line 421: if there a null is possible, then please check and throw an exception.

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java
```diff
@@ -674,29 +673,28 @@ private void purgeService(DeployTask purgeTask, ServiceDeploymentEntity serviceD
...
                     createNewDeployTask(
-                            newServiceId,
-                            workFlowDeployTask.getUserId(),
-                            workFlowDeployTask.getDeployRequest());
+                            workFlowDeployTask.getUserId(), workFlowDeployTask.getDeployRequest());
             deployTask.setOriginalServiceId(workFlowDeployTask.getOriginalServiceId());
             deployTask.setParentOrderId(workFlowDeployTask.getParentOrderId());
             deployTask.setWorkflowId(workFlowDeployTask.getWorkflowId());
             deployService(deployTask);
+            return new ServiceOrder(workFlowDeployTask.getOrderId(), deployTask.getServiceId());
```
swaroopar, line 697: is there a difference between this and line 699?

swaroopar (approved):
LGTM

swaroopar (changes requested):
approved by mistake

Alice1319:
LGTM

swaroopar (changes requested):

ruwenqiang123:
LGTM

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/utils/MdcUtils.java
```diff
@@ -0,0 +1,27 @@
...
+/** The tools of MDC. */
+public final class MdcUtils {
+
+    /** Put serviceId to MDC. */
+    public static void putServiceId(String serviceId) {
+        MDC.put(LoggingKeyConstant.SERVICE_ID, serviceId);
+    }
+
+    /** Put orderId to MDC And Set orderId for DeployTask. */
+    public static void putOrderId(String orderId, DeployTask task) {
```
swaroopar, line 23: Not the right place to handle deployTask. 
MDCUtils must only set MDC based data.
Also, is it not possible to set the MDC directly in the storage classes when the entity is created?

And for service Id also do the same for deploy use case?

swaroopar (changes requested):

swaroopar (approved):
LGTM

PR #2471, opened 2025-03-03
add output variables to handle sensitive fields in output properties

baixinsui:
Fixed https://github.com/eclipse-xpanse/xpanse/issues/2364
Related to https://github.com/eclipse-xpanse/xpanse-ui/pull/1566 
and https://github.com/eclipse-xpanse/xpanse-samples/pull/19
Encoded value of admin_passwd in table `service_deployment_output`:
image
Decoded value of admin_passwd in `DeployedServiceDetails`:
image


Alice1319:
LGTM

dengyali0125:
LGTM

ruwenqiang123:
LGTM

qigaoneng:
LGTM

WangLiNaruto:
LGTM

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/ServiceDetailsViewManager.java
```diff
@@ -342,16 +342,23 @@ private void setServiceConfigurationDetailsForDeployedService(
...
-            deployedService.setInputProperties(inputPropertiesWithSensitiveFields);
+        // Handle sensitive data in input properties
+        List<InputVariable> inputVariables =
+                serviceTemplate.getOcl().getDeployment().getInputVariables();
+        sensitiveDataHandler.handleSensitiveDataInInputProperties(
+                inputVariables, deployedService.getInputProperties());
+        // Handle sensitive data in output properties
+        List<OutputVariable> outputVariables =
+                serviceTemplate.getOcl().getDeployment().getOutputVariables();
+        if (deployedService instanceof DeployedServiceDetails deployedServiceDetails) {
```
swaroopar, line 355: i see both if and the else have the same logic. Can we simply merge this?

modules/models/src/main/java/org/eclipse/xpanse/modules/models/servicetemplate/Deployment.java
```diff
@@ -37,10 +37,23 @@ public class Deployment implements Serializable {
...
+
+    @Valid
+    @NotNull
+    @NotEmpty
+    @UniqueElements
+    @Schema(
+            description =
+                    "The output variables for the deployment, which will be generated by the"
+                        + " deployer.The list elements must be unique. All output variables are put"
+                        + " together to build a JSON 'object' with each variable as a property of"
```
swaroopar, line 54: Does the last line here is needed? - 

All output variables are put"
                        + " together to build a JSON 'object' with each variable as a property of"
                        + " this object.

swaroopar (changes requested):

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/SensitiveDataHandler.java
```diff
@@ -112,4 +86,81 @@ public void encodeDeployVariable(
...
+     * @param outputVariables defined output variables in the service template.
+     * @param outputProperties output properties of the service deployment.
+     */
+    public void encodeOutputVariables(
+            List<OutputVariable> outputVariables, Map<String, String> outputProperties) {
+        if (CollectionUtils.isEmpty(outputVariables) || CollectionUtils.isEmpty(outputProperties)) {
+            return;
+        }
+        outputVariables.stream()
+                .filter(variable -> variable.getSensitiveScope() != SensitiveScope.NONE)
```
swaroopar, line 123: how is the handling for ONCE? in this case of ONCE, we can directly set to *****

swaroopar (changes requested):

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/SensitiveDataHandler.java
baixinsui: Thismethod is used to encode the value in the table 'SERVICE_DEPLOYMENT_OUTPUT', It cannot be set to ******* in this method, so the user cannot obtain the original value

swaroopar (approved):
LGTM

PR #2478, opened 2025-03-05
make service_template as foreign key in service_deployment

### LINKED ISSUE #2429: make service_template as foreign key in service_deployment
Current SERVICE_DEPLOYMENT stores the template ID ID without a FK link to it. 

- [ ] This must be changed to a FK. 
- [ ] No changes in data model. We still return only ID of the service template.

WangLiNaruto:
Fixed #2429 

image
image


baixinsui:
LGTM

modules/database/src/main/java/org/eclipse/xpanse/modules/database/service/ServiceDeploymentEntity.java
```diff
@@ -139,9 +141,11 @@ public class ServiceDeploymentEntity extends CreatedModifiedTime {
     @Column(name = "SERVICE_STATE")
     private ServiceState serviceState = ServiceState.NOT_RUNNING;
 
-    /** The id of the Service Template. */
-    @Column(name = "SERVICE_TEMPLATE_ID", nullable = false)
-    private UUID serviceTemplateId;
+    /** The Service Template. */
+    @ManyToOne
+    @JoinColumn(name = "SERVICE_TEMPLATE_ID", nullable = false)
+    @OnDelete(action = OnDeleteAction.CASCADE)
```
swaroopar, line 147: This must have no delete cascade. This table will only reference the other table.

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java
```diff
@@ -390,7 +390,9 @@ private ServiceDeploymentEntity storeNewDeployServiceEntity(DeployTask deployTas
         entity.setDeployResources(new ArrayList<>());
         entity.setServiceVendor(deployTask.getServiceVendor());
         entity.setServiceDeploymentState(ServiceDeploymentState.DEPLOYING);
-        entity.setServiceTemplateId(deployTask.getServiceTemplateId());
+        ServiceTemplateEntity serviceTemplateEntity =
+                serviceTemplateStorage.getServiceTemplateById(deployTask.getServiceTemplateId());
```
swaroopar, line 394: Is it really needed to fetch the full entity here to set it?it should be ideally sufficient to simply create a new ServiceTemplateEntity with only ID value set and then the hibernate will automatically link to the correct template entity

modules/database/src/main/java/org/eclipse/xpanse/modules/database/service/ServiceDeploymentEntity.java
swaroopar, line 147: This can also be a lazy dependency. Pls check other such examples in our codebase

qigaoneng:
LGTM

swaroopar (changes requested):

modules/database/src/main/java/org/eclipse/xpanse/modules/database/service/ServiceDeploymentEntity.java
WangLiNaruto: https://github.com/eclipse-xpanse/xpanse/blob/211a2dc94195de801360f6784376679b3e26ba18/modules/database/src/main/java/org/eclipse/xpanse/modules/database/servicepolicy/ServicePolicyEntity.java#L50

https://github.com/eclipse-xpanse/xpanse/blob/211a2dc94195de801360f6784376679b3e26ba18/modules/database/src/main/java/org/eclipse/xpanse/modules/database/servicetemplaterequest/ServiceTemplateRequestHistoryEntity.java#L51

ruwenqiang123:
LGTM

profoundwu:
LGTM

modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/DeployService.java
```diff
@@ -390,7 +390,10 @@ private ServiceDeploymentEntity storeNewDeployServiceEntity(DeployTask deployTas
         entity.setDeployResources(new ArrayList<>());
         entity.setServiceVendor(deployTask.getServiceVendor());
         entity.setServiceDeploymentState(ServiceDeploymentState.DEPLOYING);
-        entity.setServiceTemplateId(deployTask.getServiceTemplateId());
+        ServiceTemplateEntity serviceTemplateEntity = new ServiceTemplateEntity();
+        serviceTemplateEntity.setId(deployTask.getServiceTemplateId());
+        serviceTemplateEntity.setOcl(deployTask.getOcl());
```
swaroopar, line 395: @WangLiNaruto do we really need this?

swaroopar (changes requested):

PR #2494, opened 2025-03-19
create SERVICE_OBJECTS and SERVICE_DEPENDENT_OBJECT table

### LINKED ISSUE #2486: create SERVICE_OBJECTS table
- [ ] Similar to SERVICE_CONFIGURATION Table.
- [ ] Columns – OBJECT_ID (generated ID), SERVICE_ID(FK), OBJECT_TYPE(VARCHAR), OBJECT_IDENTIFIER (VARCHAR), Properties (JSON), 
- [ ] Entries added, removed and updated only when the update from agent comes via the ‘updateServiceChangeResult’ method.  
- [ ] Add new optional OBJECT_ID column to  SERVICE_ORDER. This must be an FK to SERVICE_OBJECT table. – This value will be available at the beginning only for update and delete when order is received and for create, if the ACTION type is OBJECT_CREATE this will be updated at the end of the creation for completeness.

### LINKED ISSUE #2487: create SERVICE_DEPENDENT_OBJECT table
- [ ] For every service object that is dependent on any other service objects must be listed here. 
- [ ] OBJECT_ID and DEPENDENT_OBJECT_ID  - both should have FK dependency on SERVICE_OBJECT table. 
- [ ] This must be a child table to SERVICE_OBJECT table. We can try to make it Similar to SERVICE_DEPLOYMENT and Output properties.

WangLiNaruto:
Fixed #2486 
Fixed #2487 

image
image
image

### DIFF
--- a/modules/database/src/main/java/org/eclipse/xpanse/modules/database/serviceobject/ServiceObjectEntity.java
+++ b/modules/database/src/main/java/org/eclipse/xpanse/modules/database/serviceobject/ServiceObjectEntity.java
@@ -1,0 +1,66 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * SPDX-FileCopyrightText: Huawei Inc.
+ *
+ */
+
+package org.eclipse.xpanse.modules.database.serviceobject;
+
+import io.hypersistence.utils.hibernate.type.json.JsonType;
+import jakarta.persistence.CollectionTable;
+import jakarta.persistence.Column;
+import jakarta.persistence.Convert;
+import jakarta.persistence.ElementCollection;
+import jakarta.persistence.Entity;
+import jakarta.persistence.FetchType;
+import jakarta.persistence.GeneratedValue;
+import jakarta.persistence.GenerationType;
+import jakarta.persistence.Id;
+import jakarta.persistence.JoinColumn;
+import jakarta.persistence.OneToOne;
+import jakarta.persistence.Table;
+import java.io.Serial;
+import java.io.Serializable;
+import java.util.Map;
+import java.util.Set;
+import java.util.UUID;
+import lombok.Data;
+import org.eclipse.xpanse.modules.database.common.ObjectJsonConverter;
+import org.eclipse.xpanse.modules.database.service.ServiceDeploymentEntity;
+import org.hibernate.annotations.Type;
+
+/** ServiceObjectEntity for persistence. */
+@Table(name = "SERVICE_OBJECT")
+@Entity
+@Data
+public class ServiceObjectEntity implements Serializable {
+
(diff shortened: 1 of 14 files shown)


modules/deployment/src/main/java/org/eclipse/xpanse/modules/deployment/ServiceOrderManager.java
```diff
@@ -121,6 +121,7 @@ public ServiceOrderEntity createAndStoreGenericServiceOrderEntity(
         serviceOrderEntity.setOrderStatus(OrderStatus.CREATED);
         serviceOrderEntity.setStartedTime(OffsetDateTime.now());
         serviceOrderEntity.setRequestBody(getRequestBody(originalRequest));
+        serviceOrderEntity.setHandler(Handler.AGENT);
```
swaroopar, line 124: Handler is not AGENT always. it depends on the order type.

modules/database/src/main/java/org/eclipse/xpanse/modules/database/serviceobject/ServiceObjectEntity.java
```diff
@@ -0,0 +1,68 @@
...
+
+    @OneToOne(fetch = FetchType.LAZY)
+    @JoinColumn(name = "SERVICE_ID", nullable = false)
+    private ServiceDeploymentEntity serviceDeploymentEntity;
+
+    @JoinColumn(name = "OBJECT_TYPE", nullable = false)
+    private String objectType;
+
+    @JoinColumn(name = "OBJECT_IDENTIFIER", nullable = false)
+    private ObjectIdentifier objectIdentifier;
```
swaroopar, line 54: here we must only save the name of the Objectidentifier. Not the entire object.

modules/database/src/main/java/org/eclipse/xpanse/modules/database/serviceobject/ServiceObjectEntity.java
```diff
@@ -0,0 +1,68 @@
...
+
+    @Column(name = "PROPERTIES", columnDefinition = "json", length = Integer.MAX_VALUE)
+    @Type(value = JsonType.class)
+    @Convert(converter = ObjectJsonConverter.class)
+    private Map<String, Object> properties;
+
+    @ElementCollection
+    @CollectionTable(
+            name = "SERVICE_DEPENDENT_OBJECT",
+            joinColumns = @JoinColumn(name = "OBJECT_ID", nullable = false) // 当前对象ID
```
swaroopar, line 64: please keep the comment in english.

swaroopar (changes requested):

Alice1319:
LGTM

PR #2556, opened 2025-05-22
use prompt templates and JSON responses from LLM

swaroopar:
fixes https://github.com/eclipse-xpanse/xpanse/issues/2555


iskey (approved):
LGTM

